PR #33, opened 2016-07-26
WIP: starting gal_part integration

nbatta:


sbird:
This very much doesn't match the modern code (although it is a good idea!) so I'm finally closing it.

PR #61, opened 2017-02-19
Checkforce

rainwoodman:
This branch adds some infrastructure for testing (RestartFlag = 99), as well as adds a new pair wise short range force module. It is hell a lot slower than the tree! (x2?), but the results seem to be consistent.

There is a vast clean up regarding TimeBegin. It is no longer read from the parameter files. Also IonizeParam is moved into set_global_time.

RestartFlag and RestartSnapNum are now local variables.


sbird:
Ok, go for it. Neutrino merge is not yet. We always reserve the ability to advertise at some later date.

sbird:
hm - then it is weird a bit that the force error is so large. Is this running on MPI?

rainwoodman:
I had some tests regarding MPI in the notebook. 1 rank and 4 ranks gave identical forces.

sbird:
I wonder if the large errors are just particles with small absolute force (and so it doesn't matter much)?

rainwoodman:
Yes. They are. The dots are the 400 most different particles.

[image: Inline image 1]

sbird:
I can't see that image but that sounds like not a problem then.

rainwoodman:
image

rainwoodman:
I'll attach the image for the record. (didn't know now I can copy and paste from a notebook to github comments.

PR #65, opened 2017-03-07
Yufeng

sbird:
Merge the code for lyman alpha forest analyses. This changes the helium photo-heating rate in
an attempt to approximately account for helium reionisation. It allows the temperature and 
density of the IGM to be marginalised over. This is four extra optional parameters, 
which are fairly forest-specific.

Also fix the tests and save a little time in the cosmology.

### DIFF
--- a/param.c
+++ b/param.c
@@ -175,6 +175,7 @@ create_gadget_parameter_set()
     param_declare_double(ps, "UVRedshiftThreshold", OPTIONAL, -1.0, "Earliest Redshift that UV background is enabled. This modulates UVFluctuation and TreeCool globally. Default -1.0 means no modulation.");
 
     param_declare_int(ps, "HydroOn", REQUIRED, 1, "Enables hydro force");
+    param_declare_int(ps, "DensityOn", OPTIONAL, 1, "Enables SPH density computation.");
     param_declare_int(ps, "TreeGravOn", OPTIONAL, 1, "Enables tree gravity");
     param_declare_int(ps, "StarformationOn", REQUIRED, 0, "Enables star formation");
     param_declare_int(ps, "RadiationOn", OPTIONAL, 0, "Include radiation density in the background evolution.");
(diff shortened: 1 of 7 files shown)


sbird:
Pushed as a branch to this repo as github doesn't show my copy being a fork anymore (?)

rainwoodman:
Renaming? See if the customer service can do anything?

sbird:
I would guess because it was forked from the old -private repo, not the public. I may ask customer service at some point.

sbird:
By the way: the reionisation model has been extensively tested with 50+ small Lyman alpha simulations by now

param.c
```diff
@@ -281,7 +282,12 @@ create_gadget_parameter_set()
     param_declare_double(ps, "WindFreeTravelLength", OPTIONAL, 20, "");
     param_declare_double(ps, "WindFreeTravelDensFac", OPTIONAL, 0., "");
 
+    /*These parameters are Lyman alpha forest specific*/
     param_declare_double(ps, "QuickLymanAlphaProbability", OPTIONAL, 0, "");
+    param_declare_int(ps, "HeliumHeatOn", OPTIONAL, 0, "");
```
rainwoodman: What about the units? HeliumHeatThresh is redshift or?

timestep.c
```diff
@@ -32,7 +32,6 @@ void set_global_time(double newtime) {
     All.cf.fac_egy = pow(All.Time, 3 * GAMMA_MINUS1);
     All.cf.hubble = hubble_function(All.Time);
     All.cf.hubble_a2 = All.Time * All.Time * hubble_function(All.Time);
-    All.cf.D1 = GrowthFactor(All.cf.a);
```
rainwoodman: I thought we write D1 in the power spectrum -- the last column is normalized to Redshift 0.

sbird: We do - I just compute it when writing the powerspectrum now, instead of in set_global_time

sbird: It is a small change, but useful, because the growth factor used is not quite right (at the level of 1%) in the presence of radiation (or massive neutrinos). So we don't want to use it for anything dynamical.

rainwoodman: OK. I recall I was foreseeing to add the fastpm time stepping factors when I left D in `set_global_time`; but it is not otherwise useful, and a bunch of things are still missing.

param.c
sbird: I will add doc strings

timestep.c
sbird: Ah, I see. I can lose the commit if you want to leave it in there for fastpm?

rainwoodman: Don't worry about it. I can move it back when I actually put in fastpm.

sbird:
Am I good to merge this?

rainwoodman:
Yes. go ahead. Sorry I've been quite occupied recently...

I think we need to add some sanity tests to travis, now that the repo is public; after that we can ease up the merging process : if tests are green, then old features are not broken.

sbird:
No problem. More sanity tests would be good. You could run the cmocka tests as well, although there aren't very many.

PR #68, opened 2017-04-12
For yufeng: another set of cleanups from me

sbird:
This set of cleanups:
- Fixes compile warnings
- Fixes a parsing problem I encountered with my parameter files
- Makes TypeOfTimestepCriterion an actual nullop; the parameter is declared for compatibility, but does nothing.
- Fixes an ifdef around the IonizeParams call; without this all hydro sims wil be wrong!
- Removes ADAPTIVE_GRAVSOFT_FORGAS_HSML
- Removes NoTreeType as it is non-functional. Another way to make particle neutrino sims work is pending.
- Avoids computing initial smoothing lengths for DM particles. This was mainly a problem with NoTreeType, which I later removed anyway, but is I think a generally good cleanup.

### DIFF
--- a/timestep.c
+++ b/timestep.c
@@ -406,15 +406,8 @@ int get_timestep(int p,		/*!< particle index */
             }
             dt = sqrt(2 * All.ErrTolIntAccuracy * All.cf.a * All.SofteningTable[P[p].Type] / ac);
 #ifdef ADAPTIVE_GRAVSOFT_FORGAS
-#ifdef ADAPTIVE_GRAVSOFT_FORGAS_HSML
             if(P[p].Type == 0)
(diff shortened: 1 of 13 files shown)


timestep.c
```diff
@@ -36,9 +37,7 @@ void set_global_time(double newtime) {
 #ifdef LIGHTCONE
     lightcone_set_time(All.cf.a);
 #endif
-#ifdef COOL
```
rainwoodman: This is it. This is it. I think this is why the star formation was all wrong. IonizeParams are always set to z=0 if this line is not enabled -- which means additional heating and less star formation..

rainwoodman:
Thanks. I've merged -- you can delete the branch.

timestep.c
sbird: Glad I noticed it!

PR #69, opened 2017-04-27
WIP : Compatibility with FastPM.

rainwoodman:
1. MimicFastPMIO = 1 for MP-Gadget. Will write/expect peculiar velocity in the snapshot and FOF.
We may want to change this to the default. 

2. MimicFastPMIO = 1 for MP-GenIC. will write peculiar velocity in the IC; This shall probably be made to default as well. 
3. InputInLog10 = 0 means the Input is not in log10.

4. Units and BoxSize are read from the IC; such that if MP-GenIC / FastPM sets the Mpc/h units in the header, those numbers will be used. Otherwise a set of default is used (depending on MimicFastPMIO)

@sbird How do you think?

This should mean we can immediately use a FastPM output as an IC, though I haven't tested it.  This means we can use an IC with constrained gaussian, Non-Gaussianity, etc.
I haven't figured a way to add Gas particles.


begrun.c
```diff
@@ -196,3 +202,105 @@ void close_outputfiles(void)
 }
 
 
+/*! Computes conversion factors between internal code units and the
+ *  cgs-system.
+ */
+static void
+set_units(void)
+{
```
sbird: Can you set TimeBegin here as well, since it is read in petaio_read_header?
It might also be worth reading OmegaBaryon from the IC file directly.

examples/fastpm-compat/fastpm.lua
```diff
@@ -0,0 +1,40 @@
+-- parameter file
```
rainwoodman: It was not merged. I kept the file here as an example to run identical set up with FastPM.

allvars.h
```diff
@@ -320,6 +320,7 @@ extern struct global_data_all_processes
         int MinNumWriters;        /* Min Number of concurrent writers, this caps number of writers */
         int EnableAggregatedIO;  /* Enable aggregated IO policy for small files.*/
         size_t AggregatedIOThreshold; /* bytes per writer above which to use non-aggregated IO (avoid OOM)*/
+        int MimicFastPMIO;
```
rainwoodman: UseFastPMSnapshots?

rainwoodman:
The input was log10(k) log10(p) in MPC/h units. Was camb in k, p? FastPM reads k, p -- I was given k, p files more frequently then log10(k) log10(p) files..

sbird:
CAMB is k,p. I have never seen a code that outputs log(k) log(p)...probably easier to open the file in numpy and exponentiate it, unless it is something common?

examples/fastpm-compat/fastpm.lua
sbird: Ok, makes sense.

allvars.h
sbird: Sounds good.

begrun.c
rainwoodman: Yes. It shall be set after init_clouds. I'll add a commit.

rainwoodman:
P-Gadget has been reading log10k log10p ....

begrun.c
sbird: I ask because I wanted to do this for the neutrinos the other day, but could not be bothered to split out the header reading :)

sbird:
Where? genic-power.c definitely reads k,p, and takes the log on line 140. But your idea of being smart is a good one.

rainwoodman:
No it didn't.
https://github.com/rainwoodman/MP-Gadget/blob/master/genic-power.c#L141

sbird:
Silly me, I was misreading the units conversion in the line above. Indeed. Actually I have only ever used my GenIC, which definitely reads k,p!

rainwoodman:
Renaming to UsePeculiarVelocity?

sbird:
That works too

rainwoodman:
FYI: I added a RSDFactor attr to the header -- multiply velocity by this number gives you velocity in RSD shifting distance.

sbird:
What does that mean? I'm guessing it default to 1 if not set.

petaio.c
```diff
@@ -223,12 +252,20 @@ static void petaio_write_header(BigFile * bf) {
                 big_file_get_error_message());
     }
 
+    /* conversion from peculiar velocity to RSD */
+    double RSD = 1.0 / (All.cf.a * All.cf.hubble);
```
rainwoodman: RSDFactor is this number. It could be 'handy' for making spectras.

rainwoodman:
Do you use 'SofteningXXX' at all? Aren't they just 1/32 of mean separation particles? Do you do something special for neutrinos?

rainwoodman:
if you do then I'll revert 3cc1523.

sbird:
I don't - I always set it to 1/30 of the mean separation (but I'm sure it comes out the same!).

Your simplification is a very good idea. 

I do think we should have a parameter for the 1/32 though - maybe make it optional and default to 1/32?
Also I have recently wanted to do what Illustris-TNG does, and make the softening constant in physical rather than comoving units for z<1. Can we have a parameter for "redshift below which softening is constant in physical units" as well? I can implement this if you like.

rainwoodman:
I still don't think I have captured the best way to parameterize these -- does it make sense to have different comoving / proper softening for different particle types? The relative unit 1/30 will produce this when the number density of different species of particles are different.

sbird:
Yes - the softening should be the same for all particle types. Illustris-TNG uses: L_{box} /N^{1/3}_{DM} /40.
FIRE does similarly. It does not make sense for the softening to depend on particle type because the gravitational interaction is the same for all collisionless particles (although for gas adaptive softening may be useful).

rainwoodman:
I will revert to the old parametrization -- the unit is in Kpc/h or Mpc/h depending on the unit of choice in the IC. Let's worry about it in the future.

PR #71, opened 2017-05-08
Cleanups and fixes to the power spectrum

sbird:
Some cleanups

- Fixes to the power spectrum normalisation
- Make sure power is output simultaneously to the snapshot

### DIFF
--- a/gravpm.c
+++ b/gravpm.c
@@ -266,8 +266,7 @@ static void potential_transfer(int64_t k2, int kpos[3], pfft_complex *value) {
         return;
     }
 
-    double asmth2 = (2 * M_PI) * All.Asmth[0] / All.BoxSize;
-    asmth2 *= asmth2;
+    const double asmth2 = pow((2 * M_PI) * ASMTH / All.Nmesh,2);
     double f = 1.0;
     const double smth = exp(-k2 * asmth2) / k2;
     /* the CIC deconvolution kernel is
--- a/gravpm.c
+++ b/gravpm.c
@@ -276,11 +276,16 @@ void powerspectrum_compute(const int64_t k2, const int kpos[3], pfft_complex * c
     /* Measure power spectrum: we don't want the zero mode.
      * Some modes with k_z = 0 or N/2 have weight 1, the rest have weight 2.
      * This is because of the symmetry of the real fft. */
-    int kint = floor(sqrt(k2));
-    if(kint >= 0 && kint < PowerSpectrum.size) {
+    if(k2 > 0) {
+        /*How many bins per unit (log) interval in k?*/
+        const double binsperunit=(PowerSpectrum.size-1)/log(sqrt(3)*All.Nmesh/2.0);
+        int kint=floor(binsperunit*log(k2)/2.);
         int w;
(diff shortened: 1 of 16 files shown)


timestep.c
```diff
@@ -560,7 +560,7 @@ void find_dt_displacement_constraint(double hfac /*!<  should be  a^2*H(a)  */ )
 #endif
             dt = All.MaxRMSDisplacementFac * hfac * dmean / sqrt(v_sum[type] / count_sum[type]);
 
-            asmth = All.Asmth[0];
+            asmth = ASMTH * All.BoxSize / All.Nmesh;
```
rainwoodman: Why not as parameters?

gravpm.c
```diff
@@ -276,11 +276,16 @@ void powerspectrum_compute(const int64_t k2, const int kpos[3], pfft_complex * c
     /* Measure power spectrum: we don't want the zero mode.
      * Some modes with k_z = 0 or N/2 have weight 1, the rest have weight 2.
      * This is because of the symmetry of the real fft. */
-    int kint = floor(sqrt(k2));
-    if(kint >= 0 && kint < PowerSpectrum.size) {
+    if(k2 > 0) {
```
rainwoodman: What is the advantage of using log bins than linear bins? It seems to me going to log removes more information and can always be done as a postprocessing. 
Also, I was thinking of converting the binning to an integer arithmetic to ensure no architecture dependent rounding error changes the binning -- it matters when trying to compare results running from different machines and when there are few modes (very large scale).

This is probably impossible /difficult with log bins.

timestep.c
sbird: I can make it a parameter, good idea.

gravpm.c
sbird: It is because the power spectrum is normally plotted in log units, and normally you care more about what happens at low k than high k. So log scaling is a cheap way of placing more bins at low k. Practically this makes, eg, the BAO peak look like a rounded bump instead of a sharp peak. Integer binning would work too - one can easily change it back later though.

timestep.c
sbird: See  	9905c3b

gravpm.c
rainwoodman: Are you concerned when there are too few modes the variance is large? I thought we never want a large scale bin to be smaller than 2pi / L (then we start to miss the orthogonal modes)

rainwoodman: Also be aware when the number of modes is low P^hat is not a gaussian it becomes a chi square or something and quite nasty..

sbird:
Can I merge this?

gravpm.c
sbird: The variance is large - but this doesn't matter if you are trying to measure relative changes due to cosmology. In circumstances where I care about the variance I actually rebin the power spectrum to variable length bins guaranteed to have enough modes. But I don't like to do that all the time as it has hidden an interesting bug at least once before.

Also I don't use modes with k< 4/BoxSize usually.

rainwoodman:
Sure. Go ahead! Good advice not using modes < 4. We have difficulties with them too. :)

sbird:
Thanks!

PR #72, opened 2017-06-19
Statictree

sbird:
The main purpose of this (large) pull is to change Gadget's force tree, which is built periodically and then perturbed by kicking and drifting the tree nodes. By the end of this series the tree is rebuilt every timestep. In addition a domain decomposition is run before the tree is rebuilt.

The main reason for doing this is increased accuracy - I found that reducing TreeDomainUpdateFrequency to zero changed the particle positions. I believe this is because once the particles are moved, the oct-tree is no longer necessarily exact, in that the particles at each subnode need not be actually in the correct subnode any more. Although the node is moved according to the center of mass evolution of the particles, this may not precisely trace the particle flow, especially for collisionless particles. I believe (although I haven't done the full test) that this is the underlying cause of this issue: https://github.com/rainwoodman/MP-Gadget/issues/51 when the tree was changed, the frequency of the tree rebuild also changed.

Particles may also drift outside of the processor domain over a few timesteps. This is particularly a problem with fast-moving particles like neutrinos, and means that particles will have neighbours on different processors, causing excessive communication, and what looks externally like a hang (it might actually be a deadlock) in the tree walk. In addition if this happens tree build will fail. Thus a DomainDecomposition is run each timestep before the tree build.

You might think this would be much slower, but I have not found it so. Domain decompositions are a bit cheaper if done frequently (because less has to move) and the extra time spent in tree build is compensated by the reduced communication from a compact particle distribution (also, of course, the occasional hang is avoided, which doesn't hurt).

Once the tree nodes were no longer moving, I was able to remove large amounts of book-keeping, including all of the ExtNodes structure except hmax (which moved into Nodes), and all drift_node calls, which substantially reduced lock contention. This makes tree walk, which dominates the time, about 10-20% faster. There may be some scalability issues with the extra domain calls on large numbers of processors - I have not tested that.

I was also able to make the tree build much faster, mostly I think by removing a re-computation of the peano keys, re-using the still valid keys from the domain, which are stored in particle_data.
Finally, I OpenMP parallelized the treebuild main loop, using omp atomics to keep it fairly lockless.
This was already the limit on thread-parallelization for me, and being called every timestep made that worse.

I removed a few compile time defines, INHOMOG_GASDISTR_HINT, and DENSITY_SPH_DEBUG, and set TOPNODEFACTOR equal to 5.0 by default. I also make energy.txt optional and disabled by default.

Testing done:
I have checked that 32ff6bdfefe1 and HEAD give identical (to fp precision) particle positions and star formation rates down to z=4 in both a pure gravity and an SPH simulation. For pure gravity I checked 2 MPI ranks, 1 thread, 1 MPI rank and 2 threads, and 2 MPI ranks and 2 threads. For SPH I just checked 2 MPI ranks.

This pull has several parts:
fa0a3a7 - 0e12e457 (The first three commits) are simple cleanups.
baebacd19 - 5ff80faf72 (two commits) enable tree build and domain decomposition before PM and FoF.
32ff6bdfefe1  is an important commit. It sets the frequency of tree rebuild to every timestep. This commit changes the final output particle distribution. After this commit there should be no change in output whatsoever - everything else is just cleanup and optimization.
1301ca7c6d - 913c43340 (35 commits) removes the functions for drifting and kicking the force tree, recursively removing all functions they call, removes now useless data members, and completely removes the ExtNodes structure, moving the one remaining member into Nodes.
21e54d78d2 - 1994e198 (11 commits) makes the main loop in force_tree_build_single faster and OpenMP parallel. This introduces a hard dependency on OmpenMP 3.1
d3d1cded7a - c0155b6a (10 commits) makes the list of active particles an array instead of a linked list. Once it is an array, several loops can be made OpenMP parallel easily. I then cleaned up timestep.c in the hope of making the NextInTimeBin lists also arrays, but in the end just did the cleanups.


rainwoodman:
Thanks I'll try this out before merging. It is a bit concerning that we are now doing a full tree and domain decomposition even if say only 1000 particles are active -- In your tests how many time step bins did the simulation go? If it is more than 10 bins the balance may start to favor less tree builds?

I also remember the size of each tree node is expanded during the drift, thus the tree force should have been accurate if we go all the way down to individual particles?

sbird:
If we go to individual particles it should be accurate - but the inaccuracy only shows up after multiple timesteps, which is why I don't think your tests before showed it up.

sbird:
I ran it down to z=4 - a few hundred steps. Definitely test it a bit more - I just noticed there is some kind of rare data race in a pure OpenMP run (seems to affect exactly one particle)

sbird:
It is worth mentioning that this is also something that Gadget upstream has been doing. The domain decomp can become expensive in large runs at z=0, because each particle has to be touched even if not many are active.

sbird:
I wonder if one can speed it up by only doing domain for the active particles?

rainwoodman:
1. Not only the active particles drift during a step -- all particles
coupling with them will be drifted. It sounds like we will need to first
probe all particles in the interaction radius, drift them, then do the
actual tree walk. Did you take care of this? I am no longer sure how this
even worked with the old code, seems like we will always be missing some
pairs because the tree look up is always done on the previous location of
the particles -- unless we increase the search radius to include those
infalling particles.

2. If we have these two functions, position_to_tree_node and
position_to_domain_rank; then it is cleaner to implement updates of the
tree and the domain structure. We can keep the spatial partition fixed
between PM steps, and only work out modifications to the tree and domain
each step.

sbird:
I think the answer to 1. is that particles are always drifted to the current time in domain_Decomposition (domain.c:143). So indeed we cannot just do it with the active particles.

2. sounds a good idea. We should then split the domain into "make the domains" , which runs every PM step, and "exchange particles", which runs every timestep.

rainwoodman:
1 : I understand your new code is drifting all particles every step. But the old code was definitely not doing the full drift -- and I wonder how it could have worked at all or it indicates we don't quite understand adaptive stepping. I'll read a bit on this, too Since this is a major rework, let's make it right.

2 : The issue in 1 becomes relevant because we need a way to identify the particles that are drifted in the step in order to dynamically update the tree.

sbird:
The particle drift would have been done later, at the timestep for the particle in question. So the issue is just that the force nodes would have been misaligned, and I think this was supposed to have been taken care of by drifting the force nodes. All force nodes encountered when walking towards an active particle were drifted, I think. I suspect this was not that accurate, which may be a strong reason why in practice it did not give quite the right answer.

sbird:
Particles are also drifted during treewalk - does the call to drift_particle(other,All.Ti_Current) handle this?

rainwoodman:
Yes -- but the drift happens only if a particle is walked as a neighbour of an active particle. However since the tree only reflects the position of particles before the drift, we do not have a full list of all particles that may have interacted with the current particle. This affects not only the gravity but also other forces.

rainwoodman:
In gadget-2 paper, equation 33, the `K_sr` operators require a recent force calculation. That's why we have to drift all neighbours to the current location.

I am more confident the way it was done we would have always missed those particles that would have fallen into the radius due to the drift; and thus wrong. the particles have moved out from the cut off radius are OK, since the new position will be out. As a result we systematically underestimated the short range forces. 

Are you convinced? Did I miss something?

sbird:
Yes! That makes sense. Do you think this applied to Gadget prior to the new treewalk code as well, or only after?

rainwoodman:
After a closer look, I see how this works. Before this PR (after/before treewalk), the tree nodes are enlarged by dt_drift according to 2 * max velocity of within the node. Then all particles in any tree nodes that intersect the search radius of the particle will be drifted. 

This way the infalling particles are not missed.

rainwoodman:
I do not remember if we do have giant loops over all particles every step any ways. If so then it doesn't hurt to demand drifting all particles every step; and the code becomes closer to eq 33.

sbird:
Just pushed some extra commits to fix the threading in the tree build - before it was a) racy and b) actually slower with the tree build threaded because it screwed up cache-locality for the Node structure which made the tree walk slower.

sbird:
We do not have other giant loops around all particles - but I do not think move_particles is a bottleneck. It touches every local particle, but it is trivially threadable and not expensive. I am more worried about the particle exchanges limiting scalability.

sbird:
Illustris had exactly this problem, and Volker solved it - read the last paragraph of Section 2.2.3 of 1703.02970 and see if it means anything to you. The paper is still in prep, unfortunately.

rainwoodman:
https://arxiv.org/pdf/1205.5668.pdf is more or less reasonable. It was
based on DKD but we can rewrite it to KDK and implement it. Though I am a
bit suspicius about splitting the hamiltonians arbitrarily. The time
stepping scheme there was a function of both r and v. I remember Quinn's
9710043 motivated a stepping function that only depends on a for KDK and
only depends on v for DKD -- but the context is different.

I'd like to play a bit with 1205.5668 later of the week. I also think a
rewrite of the domain decomposition and tree is indeed long due... Do you
think we are sufficiently close to a full replacement?

sbird:
Ok - good luck. I agree that the domain code in particular is in need of attention, but I was always concerned that bugs introduced by a full rewrite would only show up at scale. What sort of thing did you have in mind? Just the split between "make the structure" and "exchange the particles" or something else?

rainwoodman:
Yes. I have second thought about a full rewrite. I don't think I am sufficiently close to a full basket of solutions... 

- For domain, yes I think we need to decouple make the structure and exchange the particles, then run exchange the particles more requently than 'make the structure'.

- What could stop us from doing that? Maintaining the coupling between tree and domain may be a huge hurdle. Do you see a way to decouple them?

sbird:
I think it is possible. Mostly just a matter of getting the memory management right.
To avoid having to drift every particle every timestep we still need 1205.5668 I think.

sbird:
In density.c we have: 
"P[place].GravCost += All.HydroCostFactor * All.cf.a * remote->Ninteractions;"

Do you know why there is a factor of 'a' in there? Hydro is not less expensive at high redshift! I'm guessing this is some kind of fudge factor for star formation?

sbird:
I tried and removing the a actually speeds up density estimation a little at high redshift - and slows down normal tree gravity a little. Doesn't make a lot of difference, but it does seem to be a slight improvement.

sbird:
As of a252e5f the domain is shorter - it makes the code marginally faster in my tests, but not hugely so (but this is on my laptop, where there are only 4 MPI ranks and they can do zero-copy data transfers). Output is unchanged. I think larger-scale tests would be a good idea now. Maybe you could check it roughly works for your use-case?

sbird:
The diffstat is 10% of the total code now and a net reduction of 1000 lines - 5% of the total. I will stop making changes that are not bugfixes for now.

rainwoodman:
My part has been slow...

Do you remember how the PM step size in Gadget is picked? 

Extending HOLD scheme to TreePM is trivial,

```
H = V_{pm} + H_S + H_F
```

After splitting it becomes

```
K_{PM}^0.5 H_F^{0.5} H_S H_F^{0.5} K_{PM}^0.5
```

Where H_S is  KDK where force is Tree force between Slow and and all particles. H_F is then divided to H_S^1 and H_F^1 recursively.

The PM step size in gadget I sort of remember is automatically decided from the slowest time step?

sbird:
I was on a flight to Hong Kong with the entertainment system broken yesterday, which really sped things up.

The PM step is set from the displacement timestep constraint; that the average particle velocity may 
not be more than 0.2 times a particle cell in a PM timestep.

I actually just (by accident) extended the PM step somewhat and I think that it could be made much longer than it is. The current choice of timestep seems rather arbitrary. Maybe we could use some multiple of the longest occupied timebin instead? I will experiment.

run.c
```diff
@@ -267,12 +266,13 @@ void find_next_sync_point_and_drift(int with_fof)
 
         set_global_time(nexttime);
 
-        move_particles(All.Ti_nextoutput);
-
+        /* Do domain decomp and rebuild tree before doing
+         * power spectrum or FoF; if the particles have
+         * moved a long way, performance will be degraded.*/
```
rainwoodman: Does this imply `without doing this`?

timestep.c
```diff
@@ -132,218 +179,193 @@ void advance_and_find_timesteps(void)
...
-           make sure that we do not 'overcool'. */
-        dt_entr = (P[i].TimeBin ? (1 << P[i].TimeBin) : 0) / 2 * All.Timebase_interval;
+    /* In case the timestep increases in the new step, we
+       make sure that we do not 'overcool'. */
+    dt_entr = (P[i].TimeBin ? (1 << P[i].TimeBin) : 0) / 2 * All.Timebase_interval;
 
-        if(SPHP(i).Entropy + SPHP(i).DtEntropy * dt_entr < 0.5 * SPHP(i).Entropy)
-            SPHP(i).DtEntropy = -0.5 * SPHP(i).Entropy / dt_entr;
-    }
+    if(SPHP(i).Entropy + SPHP(i).DtEntropy * dt_entr < 0.5 * SPHP(i).Entropy)
```
rainwoodman: This looked quite similar to line 336... any comments?

run.c
sbird: Yes - without doing this the neutrinos will break. Do you want me to add "without doing this" to the comment?

timestep.c
sbird: It does - the duplicate is there because the timestep and entropy floor may change the condition. It seems it would be best removed. But I do not understand why the condition is there at all, so I was not sure whether to touch it. Do you understand?

run.c
rainwoodman: I am confused. neutrinos will break and performance will be degraded seem to be two different things.

A related issue is last BlueTides run crashed in PM, where it alerted the number of particles included on the nodes marked with ``DEPEND_ON_LOCAL_MASS`` differs from NumPart.

It may be related to this? I haven't got around to look at it closely.

timestep.c
rainwoodman: This is the full comment from Gadget-2, if you can make sense of it..
```
          /* In case the timestep increases in the new step, we
             make sure that we do not 'overcool' when deriving
             predicted temperatures. The maximum timespan over
             which prediction can occur is ti_step/2, i.e. from
             the middle to the end of the current step */
```

rainwoodman:
In order to implement the 'HOLD' scheme, we will get rid of the entire timebin logic all together. 
A few things:
- Do you think it is reasonable to extend the code to make it capable of building several trees for several subset of particles? At every recursion, we will need a tree for the slower particles and a tree for the faster particles.

- The KDK scheme may also need to be reworked / simplified. With HOLD we no longer drift particles arbitrarily and we can finally call an integrator / ODE solver to solve for the Kick / Drift factor in each step, getting rid of the linear interpolations.

- I was able to trace down the standard implemntation of HOLD from the AMUSE project. Look for evolve_sf.c. [huayno.zip](https://github.com/rainwoodman/MP-Gadget/files/1101527/huayno.zip)

- I agree the PM step can be put very large. But if there are too few of them the large scale error will be big. I think I also see a way to change the PM stepping to use FastPM to fix this in the HOLD scheme.
I've been trying to play around with this with a Python prototype; and should be able to get it working around the week.

run.c
sbird: What I mean is "if the particles have moved far enough that they should be exchanged, tree walk performance will be degraded, sometimes to the point that the timestep will not actually complete". Particle neutrinos are a specific example that move fast enough that this happens all the time.

The symptom you describe is indeed one of the failure modes I have encountered with this problem.

timestep.c
sbird: Could it be that this is limiting an interaction between the implicit integrator for cooling (which is only moderately accurate) and the particle timesteps?

sbird:
We will lose adaptive time stepping? Or you will just rework the logic so there are multiple hierarchical trees? 

At the moment we drift particles to the start of every timestep once, in the short domain decomposition, and every other drift is a no-op (except for the particle dump one: but that I have code in progress to remove). So really it is already pure kick drift kick but the length of the drifts differ.

How large is too large? How do you set the timestep in FastPM? I accidentally changed it so that from a=0.02 to a=0.1 ran in one PM step and got almost identical particle positions to the normal gadget.

sbird:
The commits I just pushed change the snapshot saving routine so that, instead of writing in the middle of a timestep, the snapshot always aligns with the end of a PM timestep. This means that the kick is always accurately applied, even over restarts. We adjust the timesteps to the desired output times to make sure it is still writing when we want it to.

sbird:
Note a side effect is that we never drift except when we drift all particles.

sbird:
Except the logic for timeouts is broken...let me fix it.

rainwoodman:
No. We don't lose adaptive stepping. It is a cleaner way of doing adaptive stepping.

- all particles are put on a PM time line. The PM only has kick steps. We can assign a PM step at every snapshot time to make sure the snapshot is synchonized, like you currently are doing (or plan to do).

- between the PM time line we use the HOLD scheme to sync and move particles towards the next PM step.

- From what I understand currently every step a tree for all particles are built. This is good enough for the scheme, but I think we can speed it up by only building trees of active particles. I have to play with that to see if it works.

- HOLD only drifts the active particles once during the KDK -- do you drift only the active particles or all of them?

- The other change would be when counting interactions, only do it between the active particles and those in the next time bin, or something like that. I need to think more about this.

rainwoodman:
`a = 0.1`, you mean `dloga = 0.1`?

sbird:
No, I mean a=0.1: actually the PM step was from a=0.02- a=1 but I only ran it to a=0.1 before I realised the bug.

sbird:
That sounds good. At the moment we drift all particles in domain, and it would make sense not to. 
I just pushed the latest version of the PM step synced to outputs code. At the moment it has two problems, which I will hold off solving since it sounds like you will make them obsolete:

1) the requirement that timesteps be a power of two means that snapshot outputs are not quite at the desired redshift (although they are close). This is annoying because one sometimes wants to compare two snapshots that should be almost identical and if they are at slightly different redshifts this doesn't work too well.
2) TIMEOUT events do not wait until the next PM timestep, but occur at the end of the next small timestep. Estimating how long it would take until the next PM timestep was too hard at the moment. So they truncate the PM timestep instead. This slightly screws up SPH(i).VelPred. Since this quantity is zeroed anyway when we write the snapshot it isn't a large error, and should be smaller than making in-timestep snapshots. But for your scheme you probably want TIMEOUT events to only occur at the end of a PM timestep.

rainwoodman:
1. We really should have a list of PM step that are exactly sitting on the output redshift. Then this problem is solved.

2. We have to work around this. K_PM doesn't commute with H_fast, so we can't arbitrarily insert PM steps in the middle of a HOLD -- that being said though, it appears if we don't do anything but just dump the snapshot as is, we can probably work out a way torecover from the middle of the HOLD step.

I have sufficiently worked it out in a python script (attached if you'd like to play with it too) that demonstrates the time sequence of the HOLD scheme. ``good`` is the original recursive scheme, and ``bad`` is the linearized scheme.

The example is kind of extreme, with two large bins (1, 2) then a very small bin 7. If you grep for 'Bldg', it prints when a tree is created and for which time bin. You can see that to really take advantage of the HOLD scheme we need to have separate domain decomposition and tree for each time bin. It calls for a full overhaul of both and I suppose that's the future work.

Without doing that, we need to mimic several trees with one tree, which means even the smallest step we need to rebuild the entire tree, and skipping pairs by checking P.timebin. You already do this in this static tree branch. 

[hold.zip](https://github.com/rainwoodman/MP-Gadget/files/1106844/hold.zip)

I'll think a bit more about recovering from a snapshot in the middle of a HOLD

rainwoodman:
Well it is a lot simpler than I thought. Look at the function eolve2. If we save the full particle state vector, and the integer i, then we know how to continue the loop.

sbird:
By "the full particle state vector" you just mean "the list of timebins", right? Am I right in thinking that in the attached we must have the output time be a power of two?

rainwoodman:
- I mean then entire P array (and SphP, BhP)

- Yes and no. My way of think of ``evolve`` and ``evolve2`` in ``hold.py`` is that either is to replace the PMDrift step between two PM Kicks. In a uniform step PM simulation the only synchronized state is at the end of the K^1/2 D K^1/2. 

- Within ``evolve2``, we can dump full vectors as ``checkpoints``. The variables won't be synchronized, but there is sufficient information to continue the loop. A checkpoint is probably accurate only to a lower order than doing a full step, because slower particles will systematically lag behind. I haven't got my head around this yet.

sbird:
I see. It might be easier to just sync PM steps to output, actually. I have come to the conclusion that if we want to hit the output redshift exactly and preserve the power of two ordering of the timesteps we need to construct the integer timeline around the snapshot output anyway. 

The main difficulty then is the timeout events, which would need to estimate whether the PM step will finish before the timeout happens, but that is not too difficult.

rainwoodman:
Do you know the math behind:

```
        for(j = 0; j < 3; j++)
            SPHP(i).VelPred[j] +=
                (P[i].GravAccel[j] + P[i].GravPM[j]) * dt_gravkick + SPHP(i).HydroAccel[j] * dt_hydrokick;
```
in drift.c / predict.c?

This won't really work if the drift also depends on the kick factor. How important is it to have VelPred accurate in a hydro solver?

rainwoodman:
`advance_long_range_kick` also looked a bit strange to me; I never really understood the kicking back business with kickA kickB. Adding to the confusion some variables seem to be inconsistenly used? I'll give it another go later of day.

rainwoodman:
The use VelPred was described in Gadget-1 paper, section 4 and 5; in Kicks we need to take care of them.
https://wwwmpa.mpa-garching.mpg.de/gadget/gadget1-paper.pdf

rainwoodman:
It's more complicated than that. If a fast particle is active and we are computing the interaction against a slow passive particle, we need to rerun the prediction on the slow particle. Thus instead of keeping track of VelPrec, we shall modify hydro to take in Vel and the time stamp of the particle, then make prediction in the active particle.

sbird:
What you suggest makes sense. I think much more sense than what is done now. I suspect that this is a fairly small effect

rainwoodman:
Another situation that's not covered. What about newly created particles? They introduce a cross talk between the ParticleManager (domain.c) and the TimeBinManager. (timestep.c). Currently the newly formed particles are put to the same hierarchy as the parent particle. But with the new scheme, their next activation time will be computed from a variety of time scales.

The big concern is how blackhole.c works. It adds the feedback to an attribute of the surrounding gas, but doesn't update the thermal state of that particle immediately. 

We won't update the time step size of the receiver till the particle is activated next time. During this period it is quite possible the particle has already received quite some feedback without properly dissipating them to the further neighbours. 

This will be relieved after we realize that the added tree based time step size calculator is symmetrized, up to a radius of r_cut. So in physically sound situations, any particle that receives the feedback of the BH will have the same time step as the BH. In unphysical situations this may bite us.

sbird:
For newly created particles it probably makes sense to use the timestep of the parent. For blackhole.c it probably doesn't matter that much because limited resolution probably over-cools anyway. One way out would be to implement the timestep limiter.

rainwoodman:
I realized SPH / Density / SML is a problem. SPH interaction goes beyond the fast / slow split. It is not clear to me how these can be written in Kick and Drift operators.

sbird:
Why is that a problem? Right now we only use SPH interactions with active particles, and they are all local, so that doesn't seem inaccurate? Or are you just saying it is a problem to potentially walk multiple trees at once, because active particles will have inactive neighbours?

rainwoodman:
It is more of the second part. "Active particles will have inactive neighbours." The pair-wise force in the HOLD scheme adds up smartly with the recursive slow/fast division; but to get the smoothing length reasonably we need all particles, leaving the hydro force in inconsistency -- if we do all neighbours, then it is not slow/fast divided; if we do not do all neighbours, well I am not sure if the hydro force is still correct.

Take a look at
http://www.tapir.caltech.edu/~phopkins/Site/GIZMO_files/gizmo.pdf
section 2.5

If we only run the SML calculation at PM step and use a Reimann solver in between on a particle background that moves along the PM this problem is gone. Thoughts?

rainwoodman:
This could work consistently

1. Do not use any prediction / additional drifts. Always pick the quantities at the per-particle time.
2. Always redo HSML of the active particles against all particles after any drift.
 
- Prediction without additional drifts violates momentum conservation for the pair-wise interaction breaks newton's third law. 
- Prediction with additional drifts breaks the entire time integration. 

It remains to be seen how this affects the SPH -- we may have lowered the time order of the SPH integration by fixing all quantities to the latest known value. Adding in a Riemann solver of sorts could have fixed this (finding a reasonable time average between two asynchronized particles), but I don't understand Riemann solvers sufficiently to say so.

rainwoodman:
No it doesn't work quite right. It should be:

1.   Do not use any prediction / additional drifts. Always pick the quantities at the per-particle time.
2.   Always redo HSML of the both fast and slow particles against all particles *before any Kick.*

sbird:
Are you suggesting implementing MFM (the Gizmo hydro solver) above? That's been on my todo list for a while, as I think it is more accurate than SPH. It would probably solve this problem (and I think it is just as accurate if one uses HSML from an earlier timestep, since it only uses the pairwise fluxes and thus the relative HSML values (but I am not sure about this). But let us make this work first and see about that later.

rainwoodman:
Yes. I shifted the focus to make the tree and domain incremental. Even if I can't land the HOLD scheme, the incremental tree and domain update should be able to speed up the current code.

sbird:
Yes - that I think is a much better approach. Now I read through it I see the new mainloop is a major change and I think should be done in a separate branch (the switch to a short-range PP force is not free, indeed). The incremental update code fixes the major scalability problem with the statictree - which is that it touches all particles. I think I can see my way to fixing the double-kick of velocities in snapshot writes as well, with some small-ish tweaks to the integer timeline to make sure PM steps sync with snapshot outputs.

I think we should make sure this is fully baked (ie, bugs fixed, used in action) first, then switch the hydro, make sure that is baked, THEN switch the mainloop. Probably we should also write a lot more tests along the way. And as this is a new algorithm we should make a paper out of it and run the full gamut of gravity and hydro tests. 

Can we revert 6fd6e5e, make a new branch with mainloop.c in it, and remove mainloop.c from this branch before merging?

rainwoodman:
It may be easier to first merge the liftsph branch, then get rid of the mainloop.c to a new branch. 'binmask_t' can be made relevant -- by converting TimeBinActive to binmask. This way we don't have to redo the glue layer. 

My view on the double kick issue is that we can take this chance to clean up the `predict`, and shrink the integer time line to be within PM steps. We will need the kick operators if we are to put in the HOLD scheme.

sbird:
Ok, the merging strategy makes sense to me. 
binmask_t is I agree is much cleaner than TimeBinActive.

We could deal with double kicks by outputting snapshots only at PM timesteps.

Then we could make sure that the integer timeline uses the bottom few bits of the integer for the power-of-two timebin hierarchy. The upper bins could be used for the PM steps, which would not necessarily be powers of two, but set up to sync to the snapshot outputs. This would ensure there is always a global timestep counter. We could use, say, 16 bits for the power-of-two, leaving 13 bits for snapshots, which leaves room for 8192 snapshot outputs, which I think is enough.

How about that? Or do you think we should just rebuild the power-of-two timeline entirely for each PM-step/snapshot output?

sbird:
But we do still need to revert switching the short-range gravity from tree to pairwise forces in 6fd6e5e

rainwoodman:
if incremental update ia used, the moments of nodes are no longer correct.
we must fix that before switchimg back to tree force. my gut feeling is it
is costly to maintain the moments. Unlike hmax the moments is not a bound.
Maybe we can collect  tainted nodes and update them in the end?

sbird:
Yes - I think though that building a tree every timestep may still be faster than doing the pairwise force.

rainwoodman:
I had an elegant API for the tree update
```
struct transaction{
   int * tainted_nodes;
};

transaction = force_tree_begin_update_transaction();

drift_all_particles(transaction);

domain_maintain(transaction);

force_tree_end_finalize_transaction(transaction);

```

In finalize, we do a down top propagation of mass and hmax the tainted nodes, then update the pseudo particles (there is a routine that looked half readable for this)

But I worry in the end all nodes are tainted due to the use of drift_all_particles, and in that case it is more efficient to rebuild the entire tree.

sbird:
I am not sure how many nodes would in fact be tainted. This probably needs checking.

sbird:
I mean; all of them would be tainted. But I am not sure how many would be tainted at a level where it matters.

rainwoodman:
We can definitely measure that. The closer to the leaves the bigger the difference is.

sbird:
I am having trouble with the new domain code in a long test. It ran for 6 hours on 24 MPIs, 2 threads each and then crashed at z~0.5. It looks like the new domain is not load balancing correctly. Do you have any ideas before I blow a lot of time on a bisect? I suspect an inverted conditional, as it looks like there was a small imbalance initially which gradually got more severe as the simulation progressed, eventually crashing.

Here is the last step of the buggy run: All tasks above 19 look broken. It may be assigning the domain for tasks 20-23 to task 19.

==
[ 025489.96 ] Begin Step 2086, Time: 0.541833, Redshift: 0.845586, Nf = 00000000061521, Systemstep: 0.000304364, loga: 0.000561889, status:
[ 025489.96 ] TotNumPart: 0000134217728 SPH 0000000000000 BH 0000000000 STAR 0000000000000
[ 025489.96 ] Occupied:            0            1            2            3            4            5 dt
[ 025489.96 ]    bin=15            0     74163885            0            0            0            0 0.00899022
[ 025489.96 ]    bin=14            0     40198029            0            0            0            0 0.00449511
[ 025489.96 ]    bin=13            0     16208717            0            0            0            0 0.00224756
[ 025489.96 ]  X bin=11            0        61521            0            0            0            0 0.000561889
[ 025489.96 ]                -----------------------------------
[ 025489.96 ] Total:               0        61521            0            0            0            0  Sum:
   61521
[ 025489.96 ] Start force computation...
[ 025489.96 ] Begin tree force.  (presently allocated=2919.41 MB)
[ 025490.16 ] tree is done.
[ 025490.16 ] force computation done.
[ 025490.58 ] domain decomposition... (presently allocated=2176.92 MB)
[ 025490.58 ] Allocated 170.667 MByte for top-level domain structure
[ 025490.61 ] use of 256 MB of temporary storage for domain decomposition... (presently allocated=2603.59 MB)
[ 025494.16 ] Before=209
[ 025494.16 ] Final NTopNodes = 673 per segment = 28.0417
[ 025494.16 ] NTopLeaves= 589  NTopNodes=673 (space for 5592406)
[ 025494.35 ] Expected segment cost 6.10072e+10
[ 025494.35 ] Created 25 segments for an expectation of 24
[ 025494.35 ] Largest deviations from average: work=5.01758 particle load=5.06507
[ 025494.35 ] desired memory imbalance=5.06507  (limit=11184810, needed=28325913)
[ 025494.35 ] Note: the domain decomposition is suboptimum because the ceiling for memory-imbalance is reached
[ 025494.35 ] Expected segment cost 5.59241e+06
[ 025494.35 ] Created 24 segments for an expectation of 24
[ 025494.35 ] Largest deviations from average: work=4.66919 particle load=4.61798
[ 025494.35 ] Balance breakdown:
[ 025494.35 ] Task: [  0]  work=  0.9660  particle load=  1.0121
[ 025494.35 ] Task: [  1]  work=  0.8931  particle load=  0.9556
[ 025494.35 ] Task: [  2]  work=  0.9201  particle load=  0.9627
[ 025494.35 ] Task: [  3]  work=  1.2798  particle load=  1.0768
[ 025494.35 ] Task: [  4]  work=  0.9347  particle load=  1.0197
[ 025494.35 ] Task: [  5]  work=  0.8289  particle load=  0.8778
[ 025494.35 ] Task: [  6]  work=  1.0903  particle load=  1.0894
[ 025494.35 ] Task: [  7]  work=  0.9149  particle load=  1.0027
[ 025494.35 ] Task: [  8]  work=  0.7697  particle load=  0.8700
[ 025494.35 ] Task: [  9]  work=  1.2430  particle load=  1.1067
[ 025494.35 ] Task: [ 10]  work=  0.8185  particle load=  0.9430
[ 025494.35 ] Task: [ 11]  work=  0.7261  particle load=  0.8511
[ 025494.35 ] Task: [ 12]  work=  0.8953  particle load=  0.9215
[ 025494.35 ] Task: [ 13]  work=  0.9632  particle load=  0.9073
[ 025494.35 ] Task: [ 14]  work=  1.0987  particle load=  1.0802
[ 025494.35 ] Task: [ 15]  work=  0.9875  particle load=  0.9274
[ 025494.35 ] Task: [ 16]  work=  1.3722  particle load=  1.1317
[ 025494.35 ] Task: [ 17]  work=  1.3737  particle load=  1.1914
[ 025494.35 ] Task: [ 18]  work=  0.9074  particle load=  1.0297
[ 025494.35 ] Task: [ 19]  work=  4.6692  particle load=  4.6180
[ 025494.35 ] Task: [ 20]  work=  0.1035  particle load=  0.0976
[ 025494.35 ] Task: [ 21]  work=  0.0000  particle load=  0.0000
[ 025494.35 ] Task: [ 22]  work=  0.0011  particle load=  0.0010
[ 025494.35 ] Task: [ 23]  work=  0.2432  particle load=  0.3269
[ 025494.35 ] desired memory imbalance=4.61798  (limit=11184810, needed=25825607)
[ 025494.35 ] No domain decomposition that stays within memory bounds is possible.

rainwoodman:
I think I know roughly where this is. Better to have a way of reproducing
this though. Let me see what I can do.

sbird:
The only thing I noticed on a read-through was line 1009, which I think should be deleted: shouldn't have a parallel for in a reduction step.

rainwoodman:
It seems to be a factor of 2 ~ 3 slower on Edison running the lya example with Ngrid = 128 than the non-static tree.

run.c
```diff
@@ -86,24 +86,24 @@ void run(void)
             case STOP:
                 message(0, "human controlled stop with checkpoint.\n");
                 /*Write when the PM timestep completes*/
-                Ti_nextoutput = All.Ti_Current-1;
+                Ti_nextoutput = All.Ti_Current;
```
rainwoodman: Indeed. We only need it before the first snapshot 000.

genic/params.c
```diff
@@ -51,6 +50,7 @@ create_parameters()
     param_declare_int(ps, "NumWriters", OPTIONAL, 0, "");
 
 
+    param_declare_int(ps, "Ngrid", OPTIONAL, 0, "Ignored.");
```
rainwoodman: I suggest we use Ngrid in GenIC, which controls the particle grid.

Because Nmesh is used in Gadget for the long range resolution, and it is usually different from the particle number. It is supprising to see Nmesh in Gadget and Nmesh in GenIC are different.

sbird: I usually set Nmesh in both Gadget and GenIC to 2x NParticle! If you set Nmesh = NParticle in GenIC you get odd systematics near the grid scale, where the delta function from the particle grid aliases to lower k.

rainwoodman: This only occurs with small Nmesh? And moving particles to the center of cell could have fixed it (cic window drops to zero.)

sbird: No, because after particle displacement the particles are not aligned to cells. This isn't a big effect, but it should happen with all Nmesh. You care only if you are trying to get as accurate as possible power spectra.

rainwoodman: It is only suppressed by the LCDM power spectrum then (less small scale power -> less aliasing).

sbird: I see now why the confusion - in GenIC Nmesh controls the number of particles while Ngrid controlled the size of the FFT grid. In S-GenIC I use Nmesh for the size of the FFT grid and N{CDM|Baryon|Neutrino} for the particle number.

How about I file a PR report to add NumParticle as an extra parameter to GenIC, which controls the number of particles? Then Nmesh is the same as in Gadget?

sbird:
How shall we merge this to master? 

I feel we should have a period of stabilisation where we do not add features or optimisations but only try to fix bugs that we might have introduced at some point. But I also notice that we have some pending feature work (forcetree-test and the memory allocater). Shall we finish off these features and then stabilise, or shall we just stabilise right away?

genic/params.c
rainwoodman: The particle is initially on a grid -- that's why I thought Ngrid is a good name -- NumParticle sounds like Ngrid ** 3. Another choice is Nc (number of mass cells per side?), which is used in `cola_halo and fastpm`

sbird: Good point. Let me bring Ngrid back - then we preserve backwards compatibility as well.

rainwoodman:
I picture master as that sort of stable branch -- but currently master is broken too, and a fix may diverge it from statictree, where most development is in. That's one reason I think we may want to merge statictree to master.

Over longer term, we shall try to keep feature branch small, and only merge them when they are stable. As we build up the test suite it will be easier to determine if a branch is stable enough for merging. Currently we are still largely in the blind on the correctness of the code.

There are two reasons opposing merging statictree to master yet:

1. the incremental / dynamic update feature is missing, and it makes the run slower (2x slower than master on andreu's 128**3 lya box to the point before master crashes).

2. the mysterious hang up in forcetree.

I think 1 is not fatal, since the plan is to redo the dynamic update once we have proper utility to support it.

run.c
```diff
@@ -57,8 +57,8 @@ void run(void)
 
     write_cpu_log(NumCurrentTiStep); /* produce some CPU usage info */
 
-    /* find the first output time, -1 is to allow immediate output at first step. */
-    inttime_t Ti_nextoutput = find_next_outputtime(All.Ti_Current - 1);
+    /* find the first output time. */
+    inttime_t Ti_nextoutput = find_next_outputtime(All.Ti_Current);
```
rainwoodman: I have doubts about this change. Without the -1 even if the first redshift is in outputlist no output on first step is made.

sbird: But with it output is always made on the first step, which is probably not what we want. Why do we want output on the first step? Would an extra RestartFlag serve the purpose?

rainwoodman: If a snapshot is requested on the first step, then the code shall produce a snapshot. It is the least surprising behavior. Practically, a snapshot on the first step will run through all modules in a realistic setting; so it will crash early if there are mis-configurations.

rainwoodman: I agree it is a bug that when the first redshift is not requested a snapshot is always written, but the issue is probably in find_next_outputtime, not in this -1.

rainwoodman: I see that the initial time is added to OutputTimeList in init.c to avoid special conditioning between the starting point to the end point. It is not an OutputTimeList item though. To fix this I think we need to introduce a new concept, 'SyncPoints', which contains all OutputTimes. The code will do PM steps on SyncPoints for sure, and we will use the last syncpoint prior to outputtime (up to a small tolerance) to dump the snapshot.

rainwoodman: I'll give this a go. I think these sync points can serve a way to avoid running out of timebins when snapshots are too sparse, too.

rainwoodman: Take a look at this https://github.com/rainwoodman/MP-Gadget/pull/109

sbird: Looks a good idea.

sbird:
Note: the mysterious hang in tree was actually just a mistake I made in the parameter file. This branch works fine for me and completes the neutrino simulations I had trouble with in master. 

Happy to merge this to master when you are ready. I suggest though that we keep an oldmaster branch around that tracks what master used to be, as an easy point of comparison.

rainwoodman:
The very vanilla P-Gadget3 is the vanilla branch. There is also a version used to run BlueTides-i (bluetides-i) .. We can tag the current master old-dynamic-update?

rainwoodman:
I am starting to land this now.

sbird:
Hooray! Congratulations - an epic journey has completed. Or - commenced?

rainwoodman:
Both..

PR #73, opened 2017-07-05
Incremental update of the tree

rainwoodman:
this implements incremental updates of the tree.

- SPHP is lifted as BHP
- Key is persistent and updated in drift
- Added morton key to Node for easier detection of Father from new particles
- bunch of functions to manipulate the tree, not optimal
- use pairwise gravity when the moments are bad. There is away to update the moments, worth it?
- particles sent to another domain are set to zero mass, marked for GC.

It runs with the 'small' example. I haven't checked the convergence. Could you take a look @sbird ?

It is a bit strange I never seen toGo bigger than zero on the zero rank. Maybe I didn't run it long enough?

### DIFF
--- a/exchange.c
+++ b/exchange.c
@@ -6,6 +6,8 @@
 #include "endrun.h"
 #include "system.h"
 #include "exchange.h"
+#include "forcetree.h"
+#include "timestep.h"
 
 static MPI_Datatype MPI_TYPE_PARTICLE = 0;
 static MPI_Datatype MPI_TYPE_SPHPARTICLE = 0;
@@ -16,13 +18,13 @@ static MPI_Datatype MPI_TYPE_BHPARTICLE = 0;
  * exchange particles according to layoutfunc.
  * layoutfunc gives the target task of particle p.
 */
-static int domain_exchange_once(int (*layoutfunc)(int p), int* toGo, int * toGoSph, int * toGoBh, int *toGet, int *toGetSph, int *toGetBh);
+static int domain_exchange_once(int (*layoutfunc)(int p), int* toGo, int * toGoSph, int * toGoBh, int *toGet, int *toGetSph, int *toGetBh, enum ExchangeType exchange_type);
 static int domain_countToGo(ptrdiff_t nlimit, int (*layoutfunc)(int p), int* toGo, int * toGoSph, int * toGoBh, int *toGet, int *toGetSph, int *toGetBh);
 
 static void domain_count_particles();
 static void domain_refresh_totals();
 
-int domain_exchange(int (*layoutfunc)(int p)) {
+int domain_exchange(int (*layoutfunc)(int p), enum ExchangeType exchange_type) {
     int i;
     int64_t sumtogo;
     int failure = 0;
@@ -82,7 +84,7 @@ int domain_exchange(int (*layoutfunc)(int p)) {
 
         message(0, "iter=%d exchange of %013ld particles\n", iter, sumtogo);
 
-        failure = domain_exchange_once(layoutfunc, toGo, toGoSph, toGoBh,toGet, toGetSph, toGetBh);
+        failure = domain_exchange_once(layoutfunc, toGo, toGoSph, toGoBh,toGet, toGetSph, toGetBh, exchange_type);
         if(failure)
             break;
         iter++;
@@ -103,8 +105,15 @@ int domain_exchange(int (*layoutfunc)(int p)) {
     return failure;
 }
 
-static int domain_exchange_once(int (*layoutfunc)(int p), int* toGo, int * toGoSph, int * toGoBh, int *toGet, int *toGetSph, int *toGetBh)
+static int
+domain_exchange_once(int (*layoutfunc)(int p), int* toGo, int * toGoSph, int * toGoBh, int *toGet, int *toGetSph, int *toGetBh, enum ExchangeType exchange_type)
 {
+    if (exchange_type == EXCHANGE_INCREMENTAL) {
+        if(!force_tree_allocated()) {
+            endrun(0, "Force tree must be allocated to run an incremental domain exchange\n");
+        }
+    }
+
     int count_togo = 0, count_togo_sph = 0, count_togo_bh = 0, 
         count_get = 0, count_get_sph = 0, count_get_bh = 0;
 
@@ -202,10 +211,18 @@ static int domain_exchange_once(int (*layoutfunc)(int p), int* toGo, int * toGoS
         partBuf[offset[target] + count[target]] = P[i];
         count[target]++;
 
-        /* remove this particle from local storage */
-        P[i] = P[NumPart - 1];
-        NumPart--;
-        i--;
+        if (exchange_type == EXCHANGE_INCREMENTAL) {
+            /* mark the particle for removal in GC; TODO : remove it from the tree */
+            P[i].Mass = 0;
+            force_remove_node(i);
+            TimeBinCountType[P[i].Type][P[i].TimeBin] --;
+            TimeBinCount[P[i].TimeBin] --;
+        } else {
+            /* remove this particle from local storage */
+            P[i] = P[NumPart - 1];
+            NumPart --;
+            i--;
+        }
     }
     walltime_measure("/Domain/exchange/makebuf");
 
@@ -295,6 +312,12 @@ static int domain_exchange_once(int (*layoutfunc)(int p), int* toGo, int * toGoS
         }
     }
 
+    if(exchange_type == EXCHANGE_INCREMENTAL) {
+        for(i = NumPart - count_get; i < NumPart; i++) {
+            force_insert_particle(i);
+        }
+        message(0, "Added %d particles to the force tree, Numpart = %d\n", count_get, NumPart);
+    }
     myfree(bhBuf);
     myfree(sphBuf);
     myfree(partBuf);
--- a/exchange.c
+++ b/exchange.c
@@ -108,6 +108,13 @@ int domain_exchange(int (*layoutfunc)(int p), enum ExchangeType exchange_type) {
 static int
 domain_exchange_once(int (*layoutfunc)(int p), int* toGo, int * toGoSph, int * toGoBh, int *toGet, int *toGetSph, int *toGetBh, enum ExchangeType exchange_type)
 {
+
+    /* FIXME: change the last argument to a ForceTreeUpdateTransaction pointer. 
+     * if the exchange is transactional we will leave the tree in a inconsistent state but the
+     * transaction will remember sufficient to recover all invariants of the tree;
+     * which then can be done when the transaction is closed.
+     * */
+
     if (exchange_type == EXCHANGE_INCREMENTAL) {
         if(!force_tree_allocated()) {
             endrun(0, "Force tree must be allocated to run an incremental domain exchange\n");
@@ -184,6 +191,20 @@ domain_exchange_once(int (*layoutfunc)(int p), int* toGo, int * toGoSph, int * t
     for(i = 0; i < NTask; i++)
         count[i] = count_sph[i] = count_bh[i] = 0;
 
+    /* Prepare the communication buffer. 
+     *
+     * P[i].PI points to the extended attributes of the particle.
+     * When packing the particle to the buffer, we first pack the
+     * extended attributes, then pack the particle.
+     *
+     * (*1) For a packed particle, we modify PI to the offset of the
+     * extended attribute in the receive buffer relative to all
+     * particleds received to this rank. 
+     *
+     * (*2) After particles are received (appended) to the local storage
+     * we then reattach PI to the correct offset on the local storage.
+     * */
+
     /*FIXME: make this omp ! */
     for(i = 0; i < NumPart; i++)
     {
@@ -196,14 +217,14 @@ domain_exchange_once(int (*layoutfunc)(int p), int* toGo, int * toGoSph, int * t
         if(P[i].Type == 0)
         {
             sphBuf[offset_sph[target] + count_sph[target]] = SPHP(i);
-            /* Set PI to the comm buffer of this rank rather than the slot*/
+            /* See (*1) */
             P[i].PI = count_sph[target];
             count_sph[target]++;
         } else
         if(P[i].Type == 5)
         {
             bhBuf[offset_bh[target] + count_bh[target]] = BHP(i);
-            /* Set PI to the comm buffer of this rank rather than the slot*/
+            /* See (*1) */
             P[i].PI = count_bh[target];
             count_bh[target]++;
         }
@@ -212,18 +233,26 @@ domain_exchange_once(int (*layoutfunc)(int p), int* toGo, int * toGoSph, int * t
         count[target]++;
 
         if (exchange_type == EXCHANGE_INCREMENTAL) {
-            /* mark the particle for removal in GC; TODO : remove it from the tree */
-            P[i].Mass = 0;
+            /* mark the particle for removal in GC. This creates an empty slot
+             * in the main local storage (and its extended storage),
+             * both of which will be GCed eventually.
+             * */
+
+            /* Watch out: mark for removal after remove_node,
+             * in case we update the moments there. (we currently don't) */
             force_remove_node(i);
-            TimeBinCountType[P[i].Type][P[i].TimeBin] --;
-            TimeBinCount[P[i].TimeBin] --;
+            P[i].Mass = 0;
+
         } else {
-            /* remove this particle from local storage */
+            /* directly remove the particle from the local storage; this creates
+             * an empty slot in the extended attribute storage,
+             * which will be GCed eventually */
             P[i] = P[NumPart - 1];
             NumPart --;
             i--;
         }
     }
+
     walltime_measure("/Domain/exchange/makebuf");
 
     for(i = 0; i < NTask; i ++) {
@@ -282,6 +311,9 @@ domain_exchange_once(int (*layoutfunc)(int p), int* toGo, int * toGoSph, int * t
                 MPI_COMM_WORLD);
     walltime_measure("/Domain/exchange/alltoall");
 
+    /* The two identical looking code here are unpacking and setting PI.
+     * See (*2) */
+
     if(count_get_bh > 0) {
         for(target = 0; target < NTask; target++) {
             int i, j;
@@ -307,12 +339,13 @@ domain_exchange_once(int (*layoutfunc)(int p), int* toGo, int * toGoSph, int * t
                 j++;
             }
             if(j != count_recv_sph[target] + offset_recv_sph[target]) {
-                endrun(1, "communication bh inconsistency\n");
+                endrun(1, "communication sfr inconsistency\n");
             }
         }
     }
 
     if(exchange_type == EXCHANGE_INCREMENTAL) {
+        /* Insert newly received particles to the tree */
         for(i = NumPart - count_get; i < NumPart; i++) {
             force_insert_particle(i);
         }
--- a/run.c
+++ b/run.c
@@ -376,7 +376,6 @@ void compute_accelerations(void)
                              * criterion for consistent accuracy.
                              */
 
-
     if(NTotal[0] > 0)
     {
         /***** density *****/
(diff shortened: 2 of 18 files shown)


rainwoodman:
Remaining issues:

- I am trying to split advance_and_find_time_steps to an explicit Kick and an explicit FindTimeStep function; because I do not see how the kicks give correct Velocity in snapshots. The only way to do it right is to do half kicks.

The current logic of Ti_begstep looked quite twisted. It could have been much cleaner if one track Ti_drift and Ti_kick instead of Ti_begstep and Ti_current (in leapfrog current is never clearly defined). But I worry if I am missing some time stepping voodoos.

- We need to get rid of the on-demand drift in treewalk. To the first order, I think they can be safely replaced with a 'predict' that only updates the SPH properties (XXXPred) without updating the position or Ti_current. We need to add a Ti_predict to keep track of XXXPred (avoid frequently recomputing them)

Any comments on them?

run.c
```diff
@@ -58,14 +59,50 @@ void run(void)
...
+            /* PM step requires synced positions of all particles */
+
+            drift_all_particles(All.Ti_Current);
+
+            /* full decomposition rebuilds the tree */
+            domain_Decomposition();
+
+        } else {
+
+            drift_active_particles(All.Ti_Current);
```
sbird: Is it ok to only drift active particles here? I am thinking of SPH, but I guess that is handled by VelPred?

sbird:
It might also be worth adding a version of the SPHP and BHP macros, guarded behind "ifdef DEBUG", which checks consistency of the ID fields.

sbird:
What about the calls to drift_particle in gravshort_tree and treewalk.c? I left them in before because they should be no-ops under the new scheme and even if not no-ops they did no harm. But now I think they should be made asserts, as in PM. If they do anything then the tree is inconsistent.

peano.h
```diff
@@ -22,9 +22,9 @@ struct peano_hilbert_data
         (int) ((P[i].Pos[2] + All.BoxSize/2000) * DomainFac(All.BoxSize*1.001)), \
         BITS_PER_DIMENSION)
 
-#define MORTON(i) morton_key((int) ((P[i].Pos[0] + All.BoxSize/2000) * DomainFac(All.BoxSize*1.001)), \
-        (int) ((P[i].Pos[1] + All.BoxSize/2000) * DomainFac(All.BoxSize*1.001)), \
-        (int) ((P[i].Pos[2] + All.BoxSize/2000) * DomainFac(All.BoxSize*1.001)), \
+#define MORTON(Pos) morton_key((int) ((Pos[0] + All.BoxSize/2000) * DomainFac(All.BoxSize*1.001)), \
+        (int) ((Pos[1] + All.BoxSize/2000) * DomainFac(All.BoxSize*1.001)), \
+        (int) ((Pos[2] + All.BoxSize/2000) * DomainFac(All.BoxSize*1.001)), \
```
sbird: I did this in 9843f98f

rainwoodman:
Two ways of dealing with drift in Tree:

- (bad) Add PosPred, dup drift_particle, making it predict_particle, which updates all prediction variables. In tree, use PosPred for position.

- dup `drift_particle` => `predict_particle` which only updates XXPred var. Call predict_particle instead. Add timestamp `Ti_predict`. 

The second approach makes more sense to me. In Hernquist and Katz's TreeSPH paper they don't seem to advocate predicting the position of particles neither. By definition in force calculation, the position of particles are not supposed to change, since it doesn't commute with drift..

sbird:
If I understand what you say, I think that the drift_particle in gravshort_tree should just be removed, right? For pure gravity (gravshort-tree) we should never need prediction, because of the definition of the force calculation. But now I am worried because that this is not what Gadget is doing originally, and I don't understand why. What do other gravity solvers (eg, Hanno Rein's N-body code for planetary dynamics) do?

Pred variables are right now only used for SPH particles. So Ti_predict should be an SPH particle member. Then the question is really about how gravity commutes with hydro, and whether it is ok to use a (slightly) outdated Position for inactive particles when walking the tree in density.c and hydro.c?

My suspicion would be that it is fine, because SPH density estimates are noisy anyway - we already allow some slack in the number of neighbours found, and I would think that is a larger error. But we should probably check how large the effect is.

sbird:
This is Hanno's paper: https://arxiv.org/abs/1110.4876

rainwoodman:
The way to check the difference is to force all particles to sit with the fastest time bin. Let's keep that in mind.

I think we will tend to under-estimate the density because structure formation is a collapsing motion. I think the current form of the code allows adding a PosPred as well. If the effect turns of to be big. 

TreeSPH's discussion: 
http://adsabs.harvard.edu/abs/1989ApJS...70..419H
Page 427. I only read it briefly, so I may have missed some points. The argument for prediction is because of the viscosity, and they didn't do anything about position. The notation does ring  me that the prediction may as well served as a poorman's Riemann solver of sort.

sbird:
On reflection I don't think this can work. We cannot use predicted positions for the same reason that we could not have a dynamic tree. If the particles are not fully drifted and the tree rebuilt around them, the tree nodes are not in the right place. A particle currently far away with a predicted position nearby will not be found when doing a treewalk: because the particles may have moved, you cannot cull based on r anymore. This error may not be small, because we could have particles with small accelerations but large velocities.

For this same reason, using grav_short_pair in run.c does not work, and you cannot use the tree to estimate density or hydro without a full rebuild.

We need to drift the particles to the new timestep and rebuild the tree fully, or have an fast/slow operator splitting like HOLD. There is no way we can have partial tree rebuild otherwise.

sbird:
It seems to me that this is getting complicated. I actually think we should split this into a branch which is close to what I had before, plus the cleanups and bug fixes from your work (and SPH lifting). The operator splitting mainloop stuff we should keep separate. I pushed a cleanup-only rebased branch to statictree2.

The worst that happens with a full tree rebuild every timestep is that it is somewhat inefficient - it should not have poor scalability because the force tree rebuild is purely local. All we are trying to do here is to make the code faster and for that we should be profiling it more anyway - is not clear to me that a short-range pairwise force is faster than just rebuilding the tree, especially once we add in the reduced culling.

sbird:
Indeed, the predicted velocities are mostly used in the artificial viscosity computation - and artificial viscosity is very dodgy in SPH, which is another thing that makes me worry about using the predicted variables more heavily.

rainwoodman:
Using predicted position in density estimation is actually wrong.  I'll push a commit to cleanly disable the entire incremental update (reverting to your old behavior).

rainwoodman:
I think the current head of this branch is semantically identical to the head of static tree.

The file mainloop.c is not even compiled in `Makefile.rules`. Any other concerns?

Predicted variable is used to solve Hydro part of the Hamiltonian, which depends on position via the density estimation. To maintain the order or the solver one doesn't need to predict the position -- especially in virialized systems this doesn't matter at all. In non-virialized system we have bigger issues to worry at the O(time step size)**n -- the linear infall will have this error already, so there is no need to drift. This is the scheme implemented in original Gadget-1 as described by the paper the prediction stage. http://wwwmpa.mpa-garching.mpg.de/gadget/gadget1-paper.pdf, above eq 23.

I suspect the drifts in the tree were added by me at very early days. Let me check.

rainwoodman:
Indeed. There were no such drifts in Gadget-2.

rainwoodman:
In Gadget-2, all particles are drifted to the global time at the rate of the minimal time step; this differs from Gadget-1 as described in the paper. The discussion on this change in Gadget-2 paper was in Equation 33 of gadget2 paper. Notice there was no subscript in D. So the drift_all_particle operator is there for symplectic integrators.

Even with drift_all_particle one can still save substantial time by checking of the particle still remains in the same tree node and migrate particle between nodes.

The function drift_particle in tree walk was (at least) introduced in P-Gadget3. I now have two comments for adding this:
- they must have believed drifting all particles was too slow. 
- they didn't completely convert it to the HOLD scheme, which was the sympletic alternative of eq 33 that doesn't require drifting all particles; the code needs a lot of massage before we can get there.

So let's get rid of the drift particles in the treewalk. 

I would keep the supporting routines for the incremental decompose, such that we can compare the performance of incremental tree update against full tree rebuild and decide in another place. 

What I also want to keep is the changes that maintain a time stamp for each state of the particle. These can later be merged to TimeBin_TiXXX variables. The exercises made the logic in prediction easier to follow. I plan to remove the tracking of beg_step.

sbird:
Ok - let's check that this indeed gives the same result as before. 

Although the memory usage is somewhat worse because the domain key is no longer in a union and there is an extra timestep variable...

The drift_particles in treewalk and in gravshort_tree should be no-ops at the moment, so suggest making them asserts instead.

exchange.c
```diff
@@ -175,52 +185,42 @@ static int domain_exchange_once(int (*layoutfunc)(int p), int* toGo, int * toGoS
...
+        P[i].OnAnotherDomain = 0;
+        P[i].WillExport = 0;
+        target = layoutfunc(i);
 
-        if(P[n].Type == 0)
+        if(P[i].Type == 0)
         {
-            partBuf[offset_sph[target] + count_sph[target]] = P[n];
-            sphBuf[offset_sph[target] + count_sph[target]] = SPHP(n);
+            sphBuf[offset_sph[target] + count_sph[target]] = SphP[P[i].PI];
```
sbird: (Fixed by me)

sbird:
Did you see my comments above? I think you are double-counting TimeBin decrements in some places.

rainwoodman:
Oops I forgot to clean that up. It was double counted indeed. I'll clean it
up.

The memory increase for time stamps is temporary -- in principle if we know
the timebin of the particle we can look up these numbers from a table for
each time bin.

Once the hard coded parts in domain.c is cleaned up we can lift the star
variables to type 4. That will give us the memory for Key back.

In production run we can also remove ID in SphP and BhP.

rainwoodman:
OK. I'll stop pushing to this branch and let you go through -- rebase or merge to static2. Let me know when I can delete this branch .

A reminder to cherry-pick 3c9b457851143fe2dc04e149aa9cdc0b17103953 -- we'll test it tomorrow on BW.

sbird:
Ok, I have tested the pure gravity part of this (not the SPH part), and it looks good. There were some minor bugs, somewhat difficult to find because they were overlapping. For this reason I created the statictree2 and liftsph2 branches, which are semantically identical, but with the commits reordered so they can be bisected more easily (so I could I bisect them!). 

I think like to propose merging statictree2 to master when we have done more testing (especially of the SPH  code), and keeping liftsph2 for code which is not used right now, but which one can build the new mainloop upon.

rainwoodman:
Thanks. You are a git master!

- How long does your test spend in drifting and rebuilding the full tree, per particle? My hand waving numbers: the average load is about 1 million particles per core, so each full loop (like drift) is about 10 ~ 100 milli-seconds on Ghz cores assuming 10 ~ 100 cycles per particle. This is probably on the same order as the communication / synchronization of ranks. But the tree rebuild probably takes a lot more cycles (looked like on the order of 1000s or more?) 

- I will split liftsph2 to incremental_tree and mainloop, with the intention of merging the incremental tree after benchmarking sooner than mainloop. mainloop can wait. Even if the incremental tree takes more cycles per particle (say 10x), on steps where only 0.001% of particles are active we will still be able to eliminate the time in tree build.

- If the drift takes than 10 ~ 100 cycles per particle we may still want to go back to the on-demand drift scheme in P-Gadget3 when the active particles is few,  along a similar line of thoughts with tree. 

- I will continue massaging the time step variables based on statictree2, these can  probably in parallel with the incremental tree. The intention here is to clarify the current mainloop, making it easier to see the relation between the current mainloop and the future mainloop.

- A huge saving of time (factor of 2) is to compute the SPH force of pairs at the same time. AKA pair interaction is only computed when ID1 > ID2, but the force is added back to the queryoutput and then reduced to the total hydro acc. GIZMO (a version of) has this, I was told.

sbird:
I'm not sure how many cycles. In my experience the speed of the tree walk is *extremely* sensitive to the order the particles (and the tree nodes) are in. So it is very hard to tell whether any given change will be a net benefit without trying it. I have tried a few things that seemed like clear wins which were actually much slower - like a naive parallelization of the tree build. Gadget likes having the particles in P-H order, but it is not clear to me that this is optimal - when I removed the explicit P-H order after domain_maintain, treewalk got 10% faster.

On my (extremely simple) gravity-only test which has ~4 million particles per rank, TreeBuild dominates over Drift by a factor of about 10. The slowest part of the drift is likely to be the re-computation of the Peano keys. I'm not sure how fast that is compared to the tree build - computing the morton key for the tree is much faster.

Note that the current state of liftsph2 makes no semantic changes compared to statictree2 - it just adds routines for later enablement.

sbird:
I have checked SPH - it is the same result, but there is some slowdown (10%) in tree walk due (I think) to the SPH slots no longer being sorted.

rainwoodman:
The SPH slots are sorted by the gc. So the slow down is probably because now in P SPH and DM are interlaced. Maybe sort the SPH before DM in Peano ordering?

sbird:
Ah, indeed. I missed that. Thanks

sbird:
Yu, I am confused by the logic in density_postprocess . It seems to go something like:

1. In drift_all_particles, multiply density and smoothing length by a factor of exp(-Div(V)*ddrift).
2. In density_postprocess, compute DhsmlDensityFactor from the Hsml of the last timestep, multiplied by this drift factor.
3. Update Hsml with the new neighbours from this timestep.
4. Use DhsmlDensityFactor in hydro_ngbiter, which was set from predicted HSML of the last timestep.

But this looks like DhsmlDensityFactor is using the smoothing length from the last timestep, multiplied by this slightly strange factor (which I guess is meant to compensate for thermal velocities in the SPH particle?). Would not a better logical flow go:

1. Update Hsml and Density.
2. Compute DhsmlDensityFactor (probably in hydro_ngbiter) from the *updated HSML* at the current timestep.

If it were using the old timestep explicitly, that would make some sense, but it is explicitly trying to predict this factor at the new timestep (badly) rather than just using the values it is about to compute.

Or am I missing something about the SPH integration?

sbird:
Ah, I see it is used in density_check_neighbours as (I think) an ansatz to reduce the number of iterations needed for neighbour finding.

The other place it is used is in computing hydro accelerations: presumably the factor is there because the acceleration has to be reduced by some factor to compensate for the change in the cell size. But that does not explain why it uses the drifted earlier timestep, when it could difference the current values, nor why the drift factor has that form.

sbird:
Especially since Density and DhsmlDensityFactor themselves are by this point modified to be the new values; it is only Hsml on line 387 that is the old (slightly bogus) value.

rainwoodman:
This is how I make sense of it:

- Force comes from position. Hsml is a position derived quantity, thus all particles whose force is computed, the Hsml must be accurately recomputed from the position.

- Density is an iterative process. (You may have missed this) The input Hsml and DhsmlDensityFactor and Density are irrelevant (after the convergence). The new value overwrites the inaccurate value that came from drift. (I am now confused if it shall be called predict or drift).

- The dubious part is Hsml and gang of the inactive particles that are contributing as the background of the force on the active particles (force is symmetrized, using both density). The current choice is to predict/drift these values to the time of the particle, and not to use the exactly computed value at the time step, for we do not know the set of background particles in advance.

- Another choice is to mark all contributing background in the first density iteration and run a second density iteration on these background particles. This would guarantee the hydro force is computed from the current position. So we need to add the ability of running density on a selected subset of particles. I am not sure the best way to add this. Shall probably start by decoupling TreeWalk and ActiveParticle?

rainwoodman:
PS: shall we overwrite the old liftsph and statictree branch with liftsph2 and statictree2 -- github still refers to the old wrong branch on these 2 pull requests. Or we can close these two PR and move the discussion to the new PRs.

sbird:
I see, thanks for the explanation! Indeed I had missed that it was an iterative process. I think it is not worth the effort to decouple them (I think the hydro solver has larger inaccuracies anyway). If the change in position is large enough to matter for an SPH particle, it will likely become active anyway, because of the Courant condition.

I was really wondering if it would it be plausible to just get rid of the Drift step for Density and Hsml entirely. I think from your explanation the answer is no, because the inactive particles do still contribute to density, and the added terms probably do accurately reflect a diffusion timescale.

Cheers - I am happy now.

sbird:
I will overwrite the branches

sbird:
hm, I think something went wrong with the rename: it is showing conflicts which do not exist on the command line.

sbird:
That's better - fixed now.

rainwoodman:
I thought about the Hsml more and I disagree.

The KDK written in Gadget-2 paper really is K(f)D(v)F(x, v)K(f) where F(x, v) computes the acceleration from x and v.

v must be predicted because x and v are always out of sync at F. But there is no reason to use a predicted Hsml(x) -- Hsml(x) is purely a function of x. Regardless the timestep size, eventually we always end up with a contour of particles at the boundary of a fast region that are using the inaccurate predicted Hsml.

This is likely undesirable, unless I have missed some very important black magic in hydro integration -- the prediction formula is the continuity equation, and it may be more conserving to use a prediction on the background particles than using the true value?

sbird:
I'm not sure - the problem is not that Hsml is predicted - the problem is how it is modified during a drift. So it is x != x1.

I do suspect that the differences are very small (because the density estimate is noisy anyway). But I would be interested in knowing why that specific form is chosen.

rainwoodman:
Hmax is messy. Hmax is predicted every step -- and we need to use the predicted hmax in hydro due to the symmetrized kernel.

rainwoodman:
OK. This is a new plan:

1. Check NUmActiveParticles. If it is more than 1/B of total, go the full drift and rebuild approach
2. Otherwise do incremental; Drift all active particles;
3. Probe all particles within max(A x r_cut, A x predicted(Hsml)), where A is a number like 2 ~ 4. This produces a list of background particles, whose density and Hsml will be updated, but they will not necessarily be totally accurate (symmetric hydro). This can threaded and made race free if we only change a flag per particle. 
4. Pack the tree to make each node carry summations rather than averages of moments.
4. Remove background particles from the tree, update the nodes meanwhile;
5. Drift background particles;
6. Add background particles back to the tree;
7. Unpack the tree to renormalize the moments;
8. Proceed with regular force calculation.

The accuracy of the incremental tree is controlled by `A`;  The effiency is controlled by `B`. When A-> inf we drift all particles. When B->inf we also drift all particles.
This should be useful in cases where up to 10 active time bins are active.

sbird:
So the basic idea here is to do drift+rebuild only for that part of the tree which is within the short-range gravity radius of an active particle? So we would have two tree walks with O(N_active) (but a large multiplier)? That should indeed fix the main pathology of the current scheme, where the tree build can be > 60 % of a small timestep.

rainwoodman:
I realized we need to add a domain exchange between 6 and 7. Do you think it will give us trouble? My sense on this has been quite random.

sbird:
That should be ok - we do a domain exchange already there, but for the full set of particles, and it didn't seem to come close to dominating the cost in my (admittedly small) 48 processor test case.

sbird:
(but of course my sense for bluetides scale is very bad)

sbird:
We could also modify domain_exchange so it operates only on particles which were drifted. Then it should be very small.

rainwoodman:
OK. Here is why it won't work. domain exchange will change the order of particles. The tree depends on that -- particle index is the leaf node index!

rainwoodman:
We shall make GC / batch particle removal preserving the tree invariant.

On possibility is to remember the new index at old index. Then loop over all tree nodes and update them.

Or we just make sure each time we manipulate the particle the tree is updated. 

Special cases that are worrying:

- removing a particle from the first child of an internal tree node.
- inserting a particle to an empty internal tree node.

sbird:
Doh! Indeed that the exchange removes particles is a problem.

sbird:
Take a look at the forcetree-test branch - it does some of what you want. I was also able to make the tree build about 4x faster and scale well with threads. I am using it to chase a tree build crash, so not yet ready for merging.

rainwoodman:
We can have an exchange that only removes particles from the tree, and mark the particle on the storage as garbage.
We can start this by massaging the existing exchange routines -- instead of removing particles belonging to other ranks, copy and punch garbage holes. Then merge received particles to the storage by first, filling in the garbage spaces, then attach the rest to the tail.

rainwoodman:
We shall first add a 'IsGarbage' flag and modify treewalk to skip garbage from the query source and target both. This decouples Mass from garbage flag.

sbird:
That is a good idea anyway - it would allow the implementation of true tracers

sbird:
(incidentally, the better-local-refine branch fixed my tree building problem).

rainwoodman:
Is it not merged yet?

rainwoodman:
The branch was merged... I am curious how did it fix your tree build problem?

sbird:
I only just rebased my neutrino branch on top of it - I think the neutrino particles were occasionally clustering too much on one processor. The improved spatial locality from the global sort seems to have solved the problem, or at least made it rarer. That tree wait times are down also supports this.

PR #74, opened 2017-07-11
peano sort respects particle type.

rainwoodman:
This may help the 10% slow down in hydro?

### DIFF
--- a/domain.c
+++ b/domain.c
@@ -1134,13 +1134,18 @@ domain_test_id_uniqueness(void)
     message(0, "success.  took=%g sec\n", timediff(t0, t1));
 }
 
-int peano_compare_key(const void *a, const void *b)
+static int
+peano_compare_key(const void *a, const void *b)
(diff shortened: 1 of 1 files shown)


sbird:
It is possible: certainly it should restore the previous ordering. 

I will try it (but I will make sure that the domain sort doesn't respect particle type). I am a bit uncertain about the PH ordering, since fbdcf6b (do not do peano order in domain_maintain) sped things up for pure gravity (but had little effect on the SPH code). I wonder if there might be better orderings: maybe grouping similar timebins together.

sbird:
Need to experiment

sbird:
Yes, that works for fixing the regression. I will commit a similar thing to statictree2. It is not quite ideal because (full) domain now sorts it without respecting type, and then again respecting type, which is a little slow. But let's go with this for now and I'll experiment a bit with better orderings for a separate pull.

PR #75, opened 2017-07-13
Treewalk any list

rainwoodman:
This PR allows minimizes the treewalk interface, allows treewalk on any active set, updates a few names in treewalk, and eliminates almost all force related module global temp-variables.

- The code did ran to z=0 without dying on my test. I I tested density hydra and fof are giving identical results, but I do not know if they are correct for BH and SFR/COOL. f your test is stronger better run it too..

- After merging I think we shall merge ActiveParticle and NumActiveParticle to a single object, `ActiveParticleSet` which is an object of `ParticleSet`.

- The previous point highlights the lack of automated test cases. I had some idea, I'll write it down here too:
  - Upload a snapshot to amazon or somewhere with all particle types
  - finish `runtests()` to compute each force component of all particles or a subset, write the results
  - Write a python script / notebook that compares the result with known ones -- there are tools to run a notebook and diff it.
  - As the code stablizes we can add this to `.travis.yml`

### DIFF
--- a/blackhole.c
+++ b/blackhole.c
@@ -84,7 +84,7 @@ static void
 blackhole_accretion_postprocess(int n, TreeWalk * tw);
 /* feedback routines. currently also performs the drifting(move it to gravtree / force tree?) */
 static int
-blackhole_accretion_isinteracting(int n, TreeWalk * tw);
+blackhole_accretion_haswork(int n, TreeWalk * tw);
 
 static void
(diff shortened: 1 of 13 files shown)


sbird:
Looks good to me. My test is exactly what you have done.  Go ahead and merge. I agree that we should write the proposed tests!

rainwoodman:
I am next looking into shrinking the number of the data structure in domain.c and giving them more meaningful names.

PR #76, opened 2017-07-14
Domain cleanup

rainwoodman:
This PR removes the reference to a variety of 'Domain' objects in the source code. They are replaced with more precisely named variables, e.g. 

 - Domain decomposition Tree:  TopLeaves, TopNodes, Tasks
 - ForceTree : Nodes, TopLevelNodes 

algorithm for balancing is simplified. 
- for precise cumulation of cost they are all changed to int64. I hope we never overflow int64. (TimeBase is kind of big, but we still have about 40 bits)
- I hope the new code is more readable than the old one. It acts more like a state machine rather than recursion.

algorithm for update_hmax is clarified and simplified. 
- MPI_IN_PLACE looked nicer.
- Adding a struct for possibility of updating the moments with the same algorithm.
- extending this to any particle sets -- say if we want to recompute Hsml of the neighorbours.

the code currently runs, but a more through check would be nice..

### DIFF
--- a/allvars.h
+++ b/allvars.h
@@ -520,7 +520,7 @@ extern struct particle_data
     pthread_spinlock_t SpinLock;
 #endif
 
-    int GravCost;     /*!< weight factor used for balancing the work-load */
+    float GravCost;     /*!< weight factor used for balancing the work-load */
 
     int Ti_begstep;     /*!< marks start of current timestep of particle on integer timeline */
     int Ti_drift;       /*!< current time of the particle position */
--- a/allvars.h
+++ b/allvars.h
@@ -520,7 +520,7 @@ extern struct particle_data
     pthread_spinlock_t SpinLock;
 #endif
 
-    float GravCost;     /*!< weight factor used for balancing the work-load */
+    int GravCost;     /*!< weight factor used for balancing the work-load */
 
     int Ti_begstep;     /*!< marks start of current timestep of particle on integer timeline */
     int Ti_drift;       /*!< current time of the particle position */
--- a/domain.c
+++ b/domain.c
@@ -38,16 +38,14 @@
  *  communication.
  */
 
-/*Only used in forcetree.c*/
-int *DomainStartList, *DomainEndList;
-
-int *DomainTask;
-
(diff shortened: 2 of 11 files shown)


allvars.h
```diff
@@ -520,7 +520,7 @@ extern struct particle_data
     pthread_spinlock_t SpinLock;
 #endif
 
-    float GravCost;     /*!< weight factor used for balancing the work-load */
+    int GravCost;     /*!< weight factor used for balancing the work-load */
```
sbird: I thought of doing this, but the hydro cost factor is explicitly a double at the moment:

density.c:238:    P[place].GravCost += All.HydroCostFactor * All.cf.a * remote->Ninteractions;

I don't know why All.cf.a is there - I would be open to just removing it (and thus make GravCost an int), but I haven't had a chance to test what happens if I do.

sbird: See the last commit...what do you think?

rainwoodman: It must have been there to down weight the high redshift hydro. I have overlooked this. I think it is OK to leave this field float, but do the rounding in domain.c to avoid the problem -- because there we scale the variable by TimeBase, a huge number.

sbird: True. I wonder though if we should down-weight the high redshift hydro. It seems wrong...hydro is expensive at high redshift.

rainwoodman: Second thought when this rounds to zero -- it means GravCost is dominated by the gravity isn't that actually fair? This way All.HydroCostFactor roughly sets the redshift where we start to favour balancing by Hydro.

sbird: That's true. If that's what we want. 

I think it a slightly weird flag, actually - hydro is roughly 2/3 and gravity is 1/3 at early times, for one thing. But also, everything is really a tree walk, so balancing following gravity should be the same as balancing following gas. Unless it is really trying to compensate for the cost of the star formation or something at low redshift?

sbird:
As far as I can see this is fine - go ahead and merge it if you like.

rainwoodman:
I don't recall any agenda on the list. If any comes back i'll do a new one.

PR #77, opened 2017-07-18
Extratests

sbird:
This pull request adds some unit tests for the cosmology module and the timefac module. It then makes sure they pass.

This means:
- Porting the better growth function solutions from N-GenIC (only affects GenIC with radiation on)
- Slightly increasing the size of the lookup table in timefac.c

There are also some (largely unrelated) cleanups to the prediction functions, and I remove TypeOfOpeningCriterion because we never used it.

### DIFF
--- a/Makefile.rules
+++ b/Makefile.rules
@@ -48,7 +48,7 @@ INCL   = allvars.h proto.h forcetree.h cooling.h domain.h treewalk.h \
 	 fof.h cosmology.h powerspectrum.h timestep.h sfr_eff.h drift.h timefac.h \
 	 petaio.h blackhole.h
 
-TESTED = interp powerspectrum
+TESTED = interp powerspectrum cosmology
 
 EXEC = $(TARGETS:%=$(DESTDIR)/%)
 OBJS = $(OBJECTS:%=$(DESTDIR)/%)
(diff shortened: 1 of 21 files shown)


rainwoodman:
Looked good to my eyes. 

What's the purpose for adding Ti_current to Prec?

I am hoping in the near future we can remove these threadprivate caches. As we drift to Ti_current the we can compute the factors for prediction of each time bin (since we know the current time of the timebin). Then the Pred functions can directly use the time bin do make the prediction. Which means they always pred to Ti_current, thus it shall not be a parameter to Pred functions.

sbird:
Exactly! This is why I am removing the Ti_Current argument (I think you are viewing the diff backwards).

sbird:
See some partial work in that direction in the recently pushed out_timestep3

rainwoodman:
Oops. I did.

sbird:
Note by the way that this branch will change the output of N-GenIC (for the better) and may slightly change the particle output (because the drift table is marginally more accurate).

rainwoodman:
And D is correct. Merge this and rebase out_timestep3?

sbird:
Ok!

PR #84, opened 2017-07-21
Integer timeline enforces exact PM steps at snapshot output 

sbird:
This pull request changes the integer timeline so that all snapshot outputs are aligned with it. This is done by splitting the integer times into two parts: the lower bits are the usual power-of-two timeline, while the upper bits contain the number of the last snapshot output. Fitting it into one integer means that we can always compare integer timesteps, which makes things easier. It restricts us to 512 snapshots at the moment, but we can always make Ti 64 bits wide.

The goal is:
a) Snapshot output always has the right kick and drift time, and always goes through a PM timestep
b) Resuming from a snapshot output is exactly the same as continuing. 

The commits up to 18921fb are cleanups; they are a bit meandering because I wasn't sure quite what I was doing yet.  18921fb introduces the new integer timeline. Commits after that fix bugs introduced by the change. d03ad4f fixes a bug in Gadget since the beginning; that you cannot resume from a snapshot without some loss of accuracy.

All commits should compile and at least start, but the range 18921fb..5c91ff2 will get stuck after the first output.


drift.c
```diff
@@ -11,19 +11,19 @@
 #include "endrun.h"
 
 
-static int drift_particle_full(int i, int ti1, int blocking);
+static int drift_particle_full(int i, unsigned int ti1, int blocking);
```
rainwoodman: What about a new typedef uint32_t inttime_t?

global.c
```diff
@@ -59,35 +59,14 @@ struct state_of_system compute_global_quantities_of_system(void)
...
         int j;
-        double entr = 0, egyspec, vel[3];
-        double Fgravkick, Fhydrokick;
+        double entr = 0, egyspec;
 
         sys.MassComp[P[i].Type] += P[i].Mass;
 
         sys.EnergyPotComp[P[i].Type] += 0.5 * P[i].Mass * P[i].Potential / a1;
 
-        dti = P[i].TimeBin ? (1 << P[i].TimeBin) : 0;
```
rainwoodman: neat

init.c
```diff
@@ -40,10 +40,21 @@ void init(int RestartSnapNum)
...
+            All.OutputListTimes[i] = All.OutputListTimes[All.OutputListLength-1];
+            All.OutputListLength = i+1;
+            break;
+        }
+    }
 
     petaio_read_snapshot(RestartSnapNum);
 
-    init_drift_table(All.Time, All.TimeMax, TIMEBASE);
+    init_drift_table(All.TimeInit, All.TimeMax);
```
rainwoodman: I vaguely remember there was a trap regarding All.TIme and All.TimeInit. I don't remember what it is. Did you ran into anything afoul here?

run.c
```diff
@@ -37,37 +37,93 @@ enum ActionType {
...
+    unsigned int Ti_nextoutput = find_next_outputtime(All.Ti_Current);
 
     do /* main loop */
     {
         /* find next synchronization point and the timebins active during this timestep.
          * If needed, this function will also write an output file
          * at the desired time.
          */
-        All.Ti_Current = find_next_sync_point(All.Ti_nextoutput);
+        All.Ti_Current = find_next_kick(All.Ti_Current);
```
rainwoodman: Is it possible to eleminate Ti_current from All?

run.c
```diff
@@ -82,86 +138,50 @@ void run(void)
...
-        advance_and_find_timesteps();	/* 'kick' active particles in
-                                         * momentum space and compute new
-                                         * timesteps for them
-                                         */
+        /* 'kick' active particles in
+         * momentum space and compute new
+         * timesteps for them. If we are about to
+         * do output, only do a half kick.
+         */
+        advance_and_find_timesteps(WillOutput);
```
rainwoodman: What about  always to a half kick here?
Then always do another half kick? Avoid passing WillOutput to another function.

run.c
```diff
@@ -82,86 +138,50 @@ void run(void)
...
 
-            case IOCTL:
-                break;
-
-            case NO_ACTION:
-                break;
+            /*Do the extra half-kick we avoided for a snapshot.*/
+            apply_half_kick();
+            /*Find next output*/
+            Ti_nextoutput = find_next_outputtime(Ti_nextoutput);
```
rainwoodman: Does this have to be after apply_half_kick? It is related to the previous comment. If it has to be then we can probably run it every step.

run.c
```diff
@@ -193,30 +213,19 @@ update_IO_params(const char * ioctlfname)
 }
 
 static enum ActionType
-human_interaction()
+human_interaction(double lastPMlength)
```
rainwoodman: Is lastPMlength the WallClock duration between two PM steps?

test/timefac_test.c
```diff
@@ -50,6 +50,17 @@ double fac_integ(double a, void *param)
   return 1 / (h * pow(a,ex));
 }
 
+void init_integer_timeline(double a, double b)
```
rainwoodman: empty? maybe a comment explaining why it is here?

timebinmgr.c
```diff
@@ -0,0 +1,89 @@
+#include "timebinmgr.h"
+#include <math.h>
+#include "allvars.h"
+
+/*Each integer time stores in the first 10 bits the snapshot number.
+ * Then the rest of the bits are the standard integer timeline,
+ * which should be a power-of-two hierarchy.*/
+
+unsigned int out_from_ti(unsigned int ti)
```
rainwoodman: I have been in general trying to put the type of return value to a separate line in other codes as I touch the functions. (I think most GNU code does that)

Could you also do this in the new code?

timebinmgr.c
```diff
@@ -0,0 +1,89 @@
...
+    return tip - ti;
+}
+
+double get_dloga_for_bin(int timebin)
+{
+    double logDTime = logDTime_from_ti(All.Ti_Current);
+    return (timebin ? (1u << timebin) : 0 ) * logDTime;
+}
+
+unsigned int enforce_power_of_two(unsigned int dti)
```
rainwoodman: `round to`? Is it round up or round down?

timebinmgr.h
```diff
@@ -0,0 +1,36 @@
+#ifndef TIMEBINMGR_H
+#define TIMEBINMGR_H
+/* This file manages the integer timeline, 
+ * and converts from integers ti to double loga.*/
+
+/*!< The simulated timespan is mapped onto the integer interval [0,TIMEBASE],
```
rainwoodman: I am a bit confused if the outputs are not uniformly distributed, does it mean the dloga per integer step is not uniform between snapshot files?

timestep.c
```diff
@@ -19,6 +19,18 @@
  */
 
 
+/* variables for organizing PM steps of discrete timeline */
+static struct time_vars {
```
rainwoodman: typedef struct {...} TimeSpan?

rainwoodman: Is step really Maximum or the duration of the span?

timebinmgr.c
```diff
@@ -0,0 +1,89 @@
...
+/*Each integer time stores in the first 10 bits the snapshot number.
+ * Then the rest of the bits are the standard integer timeline,
+ * which should be a power-of-two hierarchy.*/
+
+unsigned int out_from_ti(unsigned int ti)
+{
+    return ti >> TIMEBINS;
+}
+
+static double logDTime_from_ti(unsigned int ti)
```
rainwoodman: is it logDTime of DlogA ? Time is such an ambiguous name here.

init.c
sbird: I think you fixed that one once TimeInit became read from the IC header.

run.c
sbird: It is possible - I actually did it at one point while I was working on this. But Ti_Current is used in so many odd places in the code that the result was rather messy and not really worth the effort.

run.c
sbird: It would be a bit cleaner, but it means we have two separate loops over all particles. Maybe that is not a problem though? Up to you.

run.c
sbird: No, it can be anywhere. It is a cheap function that does nothing to global state now.

run.c
sbird: It is - I will add a comment.

test/timefac_test.c
sbird: Whoops, that was a rebase error.

timebinmgr.h
sbird: It does - dloga/ti changes after every snapshot. I have added a comment.

timebinmgr.c
sbird: Good point, I changed the name.

timestep.c
sbird: It is the duration of the span. The typedef is a bit superfluous since it is only used once.

timebinmgr.c
sbird: rounding down. I changed the name and added a comment.

timebinmgr.c
sbird: Done - although it makes it harder to see the function return type with git grep.

drift.c
sbird: Could do that indeed. I did not just because it gets a bit messy with header includes (and is a bit painful in terms of typing).

rainwoodman: I think the fact the work is dirgy suggests there is benefit of doing so. I'll do this.

init.c
rainwoodman: I see.

run.c
rainwoodman: So it is not the right time to do this yet, later then.

run.c
rainwoodman: Force calculation is always more expensive and we are only kicking the active particles, so it shouldn't matter?

timebinmgr.c
rainwoodman: could grep -B 1 help?

run.c
sbird: The main problem actually is the calls to drift_particle() - because we don't know yet what form the long/short split will take I have been avoiding removing them. I think once we eventually settle on that algorithm we can remove All.Ti_Current. I could also hide it behind a DEBUG flag fairly easily if you wanted that.

timestep.c
rainwoodman: I suggest changing the name to reflect that.. a struct timespan shall have start and duration, or start and end (exclusive). I think the name `step`  and `PM_Ti` are a bit confusing.

run.c
sbird: That is very true. Let's make it a half-kick always then.

timestep.c
```diff
@@ -224,46 +255,63 @@ void advance_and_find_timesteps(void)
...
+        int bin = P[i].TimeBin;
+        unsigned int dti = bin ? (1 << bin) : 0;
+        /* Start of step*/
+        unsigned int tistart = P[i].Ti_begstep;
+        /* Midpoint of step*/
+        unsigned int tiend = get_kick_ti(P[i].Ti_begstep, dti);
+        /*This only changes particle i, so is thread-safe.*/
+        do_the_short_range_kick(i, tistart, tiend);
     }
+    /*Always do a PM half-kick, because this should be called just after a PM step*/
```
rainwoodman: We probably want to have this in a separate function for PM kick. It could make the outer loop closer to the formula in the paper. I'll take a look at this.

sbird: A separate function for PM kick (in the outer loop) means that the kick time is not necessarily consistent for the particle though. A separate function called here and in advance_and_find_timesteps sounds good though.

timestep.c
```diff
@@ -721,8 +768,50 @@ void rebuild_activelist(void)
...
+        ti_next_kick = Ti_Current;
+
+    /*Extract large (snapshot number) part of Ti_Current*/
+    int snap = Ti_Current & ~(TIMEBASE-1);
+    /*Zero all upper bits of Ti_Current*/
+    Ti_Current &= TIMEBASE-1;
+
+    /* find the next kick time:
+     * this finds the smallest active bin whose bit is not already set in Ti_Current. */
+    for(n = 1; n < TIMEBINS; n++)
```
rainwoodman: I am confused by this loop. It doesn't seem to do the comment above. 
To implement the loop above, I'd just do
```
for (n = 1; n < TIMEBINS; n ++) {
     if(!TimeBinCount[n])  continue;
    const inttime_t dt_bin = (1 << n);
    if (TiCurrent & dt_bin == 0) break;
}
MPI_Allreduce(MPI_INPLACE, &n, 1, MPI_INT, MPI_MIN, MPI_COMM_WORLD);
return TI_Current + (1 << n);
```

rainwoodman: This way you don't need to clear the upper bit of Ti_Current neither.

timestep.c
rainwoodman: But that's the idea of split Hamiltonian -- keeping the long range velocity fixed for the entire duration of the short range Halmiltonian.

rainwoodman: It is also benefitial to insert the Force operator to the formula. Then the function find_next_kick is actually find_next_force_computing(). A force operator updates the accel variables to the inttime based on position variables at the same time and predicted velocity variables.  A Kick operator then updates the vel variables according to the acc variables. There are kick_short(start, end), kick_long(start, end), force_short(now), force_long(now), but only one version of drift(start, end) .This wraps up the PASS scheme I think.

Especially if my observation of Ti_Current and bit was true. The detection of the active bit is similar to the 'is_edge' and 'is_center' function in my HOLD example, I think.

rainwoodman: most the literature do not explicitly write the Force operator to the formula (merged with Kick), and that's why it felt confusing how Drift and Kick are aligned, etc.

timestep.c
sbird: Ok, see  520551455

run.c
sbird: See 3fb596f

timestep.c
sbird: Ah, indeed, good point! Take a look at what I did in 3fb596f : I have split the kick functions.

timestep.c
sbird: That is not quite right - it is finding the largest timebin where this is true, not the smallest, so the loop order needs to be inverted. Let me think about it a bit more.

sbird: I thought about it and indeed all the bit work is unnecessary. What is committed in 3fb596f should be equivalent to what was there before.

timestep.c
rainwoodman: Let's try to minimize the number of functions that omit part of the behavior depending on a parameter.

I think the PM kick and short range kick can also be formally split, thus all is_PM is contained in the main loop, rather than propagated with parameters.

rainwoodman: Hold on a minute. Let me see what I can do.

sbird: Yes! This is what is done in the current code. Pull the latest.

sbird: is_PM is contained in the main loop.

sbird:
Indeed - no reason but laziness. See the last push. It is amazing to me what that rather complicated function (which had nested conditionals originally) boiled down to.

sbird:
This passes my tests. There are still some small differences from reloading from a snapshot as opposed to running from scratch, likely because the first tree grav step is opened all the way. However they are > 0.1 kpc in around 0.5% of cases in my tests, so not very large. I think I am ready to merge this, what do you think?

petaio.c
```diff
@@ -438,25 +438,25 @@ petaio_build_buffer(BigArray * array, IOTableEntry * ent, int * selection, int N
         int tid = omp_get_thread_num();
         int NT = omp_get_num_threads();
         if(NT > All.NumThreads) abort();
-        int start = (selection?((size_t) NumSelection):((size_t)NumPart)) * tid / NT;
-        int end = (selection?((size_t) NumSelection):((size_t)NumPart)) * (tid + 1) / NT;
-        int localsize = 0;
+        int start = (selection ? NumSelection : NumPart) * tid / NT;
```
rainwoodman: size_t is important. the multiplication may overflow. Did you make similar changes?

rainwoodman: if you hated two size_t, you can cast tid instead.

petaio.c
```diff
@@ -438,25 +438,25 @@ petaio_build_buffer(BigArray * array, IOTableEntry * ent, int * selection, int N
...
+        int end = (selection ? NumSelection : NumPart) * (tid + 1) / NT;
         npartThread[tid] = 0;
         for(i = start; i < end; i ++) {
             int j = selection?selection[i]:i;
             if(P[j].Type != ent->ptype) continue;
             npartThread[tid] ++;
         }
 #pragma omp barrier
-        offsetThread[0] = 0;
```
rainwoodman: I don't think there is a race -- the crash might be due to something else. 
At any step we can be assure the previous offset is computed.
Your patch makes localsize invalid except the master branch.

rainwoodman: I guess we don't use it on other threads anyways, so it is OK.

petaio.c
sbird: This was accidental, sorry.

petaio.c
sbird: The crash is I think caused by something else, actually. Defining all the ints as size_t fixes it, but this may be accidental as I do not think there is an overflow.

petaio.c
sbird: There is a pending other change to make start and end size_t as well, that I cancelled.

petaio.c
sbird: But localsize is only used inside that block - it is now only defined there, which makes it easier.

rainwoodman:
BTW you are not allowed to save a snapshot with bigfile to a NFS file system when running from several nodes. Writing to the same physical file from several clients is undefined under NFS, neither is creating directories atomic. Could it be related to yhour crash rather than the last commit?

petaio.c
rainwoodman: start and end can be int because they never go beyond NumSelection / NumPart which are ints. It is the multiplication that can overflow.

petaio.c
rainwoodman: indeed.

sbird:
I don't think so - it seems to be crashing during the building of the buffer, not the actual write. There is a segfault on line 444-448 of this branch. It only occurs with NumThreads > 1 and not with a pure DM simulation.

sbird:
I have checked that there is no corruption of the Particle array, and it works fine if offset and npartThread are both size_t - I half suspect some miscompilation by icc, since the number of particles in this simulation is nowhere close to 2**32

rainwoodman:
I think this code needs a refactor anyways -- if it is triggering an icc bug we may just do that and see if it can avoid the issue.

The list of particles to be written per type can be first computed, then used upon request for each block during the IO. If we only do this once then we no longer worry about the speed, and it doesn't have to be in threads.

sbird:
That sounds best - let me double-check that my guess is right. Do you have cycles to refactor or shall I do it?

rainwoodman:
Definitely not today... It would require adding a few functions I think.

sbird:
This is definitely a compiler error. It is crashing with:
Rewriting the loop from:
for(i = start; i < end; i ++) {
       int j = selection?selection[i]:i;
            if(P[j].Type != ent->ptype) continue;
                npartThread[tid]++;
        }
to
for(i = start; i < end; i ++) {
            if(P[i].Type != ent->ptype) continue;
                npartThread[tid]++;
        }
fixes it. Which should make no difference if selection == NULL.
I think I can also fix it by making j size_t. I will refactor because this is silly.

sbird:
My guess is this is some combination of "optimizations allowed because signed overflow is undefined" and "optimizations allowed because null pointer dereference is undefined"

sbird:
This works without the segfault for me. How does it look to you?

fofpetaio.c
```diff
@@ -77,13 +73,27 @@ void fof_save_particles(int num) {
...
+
     for(i = 0; i < IOTable.used; i ++) {
         /* only process the particle blocks */
         char blockname[128];
         int ptype = IOTable.ent[i].ptype;
         BigArray array = {0};
         if(ptype < 6 && ptype >= 0) {
+            walltime_measure("/FOF/IO/Write");
             sprintf(blockname, "%d/%s", ptype, IOTable.ent[i].name);
+            if(ptype != stype) {
```
rainwoodman: What about creating 6 arrays before the loop.  The total length is NumPart, so they will be shifted aliases of a single buffer, with the shifting calculated by return value of fof_build_selection? It may look straighter than the a single level LRU cache..

sbird: Note that the entries in IOTable are sorted by particle type, so we would only see each ptype once - the LRU cache would be rebuilt strictly once.

But I had forgotten that the particles are sorted by type at this point for efficiency anyway, so indeed we can use a shifted loop. I will add a comment too that we are relying on this.

rainwoodman: Yes it is sorted to avoid rebuilts -- but it's just slightly more twists than necessary. Perhaps add a `FIXME` such that I'll replace it when I have the time to wander the source code -- for the current version already works.

sbird: It's ok, I'll fix it on the train tomorrow. Won't take long.

sbird:
Done. Ready to merge?

rainwoodman:
Looks cool. Need a manual merge though. Are you commuting by train?

sbird:
Yup. Commuting by train from DC to Baltimore, for three more weeks. The conflict is trivial actually - just in the paramfile declaration. I will merge!

PR #85, opened 2017-07-24
Cleaner time step bookkeeping.

rainwoodman:
Explicitly remember the current state of short range Kick and Drift.
of particles, and the current state of long range Kick and Drift.

This makes the logic in prediction cleaner (compensating difference
between drift and kick).

Move the first half kick before find time step, because it is finishing
the previous time step. The second half kick uses the new time step.

The code compiles, but I suspect there are bugs.

### DIFF
--- a/timestep.c
+++ b/timestep.c
@@ -266,9 +266,9 @@ apply_half_kick(void)
         const int i = ActiveParticle[pa];
         int bin = P[i].TimeBin;
         inttime_t dti = dti_from_timebin(bin);
-        /* Start of step*/
+        /* current Kick time */
         inttime_t tistart = P[i].Ti_kick;
-        /* Midpoint of step*/
+        /* half of a step */
         inttime_t tiend = get_kick_ti(P[i].Ti_kick, dti);
         /*This only changes particle i, so is thread-safe.*/
         do_the_short_range_kick(i, tistart, tiend);
--- a/timestep.c
+++ b/timestep.c
@@ -23,9 +23,10 @@
 typedef struct {
     inttime_t length; /*!< Duration of the current PM integer timestep*/
     inttime_t start;           /* current start point of the PM step*/
+    inttime_t Ti_kick;  /* current inttime of PM Kick (velocity) */
 } TimeSpan;
 
-static TimeSpan PM_Ti;
+static TimeSpan PM;
 
 /*Get the kick time for a timestep, given a start point and a step size.*/
 inline int get_kick_ti(int start, int step)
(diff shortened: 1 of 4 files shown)


timestep.c
```diff
@@ -149,54 +151,57 @@ set_global_time(double newtime) {
...
 
-    int new_PM_Ti_length = PM_Ti.length;
-
-    /*Update the displacement timestep*/
-    if(is_PM_timestep(All.Ti_Current))
-        new_PM_Ti_length = get_long_range_timestep_ti();
+    /*Update the PM timestep size */
+    if(is_PM_timestep(All.Ti_Current)) {
+        PM.length = get_long_range_timestep_ti();
+        PM.start = PM.Ti_kick;
```
rainwoodman: I didn't see PM.start reset in the old code.

timestep.c
```diff
@@ -403,32 +380,21 @@ do_the_short_range_kick(int i, inttime_t tistart, inttime_t tiend)
...
+    const double Fgravkick2 = get_gravkick_factor(P[i].Ti_kick, ti);
+    const double Fhydrokick2 = get_hydrokick_factor(P[i].Ti_kick, ti);
+    const double FgravkickB = get_gravkick_factor(PM.Ti_kick, ti);
     int j;
     for(j = 0; j < 3; j++) {
-        VelPred[j] = P[i].Vel[j] - Fgravkick2 * P[i].GravAccel[j]
-            - P[i].GravPM[j] * FgravkickB - Fhydrokick2 * SPHP(i).HydroAccel[j];
+        VelPred[j] = P[i].Vel[j] + Fgravkick2 * P[i].GravAccel[j]
+            + P[i].GravPM[j] * FgravkickB + Fhydrokick2 * SPHP(i).HydroAccel[j];
     }
```
sbird: Hah, well noticed.

timestep.c
```diff
@@ -305,8 +280,8 @@ void
 apply_PM_half_kick(void)
 {
     /*Always do a PM half-kick, because this should be called just after a PM step*/
-    const inttime_t tistart = PM_Ti.start;
-    const inttime_t tiend =  get_kick_ti(PM_Ti.start, PM_Ti.length);
+    const inttime_t tistart = PM.Ti_kick;
+    const inttime_t tiend =  get_kick_ti(PM.Ti_kick, PM.length);
```
sbird: PM.Ti_kick + PM.length/2 might be clearer here now.

timestep.c
```diff
@@ -153,54 +155,57 @@ set_global_time(double newtime) {
...
 
-/*! This function assigns new timesteps to particles.
- * It then advances the system half a timestep in momentum space, with a half-kick operation.
- * The second half-kick is done in apply_half_kick, below.
- * The kick is split in two so that the velocities are correctly synchronized at snapshot outputs.
- */
+/* This function assigns new timesteps to particles and PM */
 void
-find_timesteps_and_half_kick(void)
+find_timesteps(void)
```
sbird: You forgot to change timestep.h

sbird:
This works for me - go ahead and merge when ready.

rainwoodman:
OK. I'll merge.

PR #87, opened 2017-07-24
Issue 86

rainwoodman:


sbird:
This should fix issue 86: it is mostly extra debugging output, with a few tweaks to the assignment algorithm.

Major changes:
1) Instead of assigning work in a round-robin, fill up each task before moving on. This means that with DomainOverDecompositionFactor > 1, we assign neighbouring segments. It also means that we 'lose' information about how much work is on the current task much more rarely, making accidental imbalances harder.
2) use the total assigned work so far rather than the mean per segment to decide when to advance. This means that an overly large segment gets balanced by a smaller one.

sbird:
Yu, one reason why 06ea1d6 was not fixing the problem is that the TopNodes were refined again at the end of determine_TopTree (line 972-1005) in this branch. But that refinement is totally bogus: it does not correctly preserve the costs or counts of the daughter nodes, making them just 1/8 of the parent nodes!
So the domain assignment would assign nodes to processors using wrong counts and an imbalance was easy to trigger.

Also the condition at the top is wrong: it should check topNodes[i].Size > 8. I do not know why that doesn't trigger much worse problems.

Shall we just remove the code that does the second refinement? It doesn't make sense to do it twice.

sbird:
If one wants a cheaper domain algorithm with worse balancing, one can just reduce TopNodeCostFactor

rainwoodman:
The refinement is not totally bogus. The old Gadget Domain decomposition is a horde of patch work to catch special cases, but it is impossible to prove that it behaves correctly.

1. The local refinement is a slow algorithm that recursively divide expensive nodes. At every level of recursion, every particle is revisited. The condition for split has a strange redundant condition, namely if a node contains most of its parent it shall be divided again. Look below

This is relevant when the merged global refinement is built. 

2. The merged global refinement is refined with the 'bogus' second refinement. The merge may combine several leaves that exist on different ranks. The result is a leaf that is more expensive than the splitting threshold. However, at this point, it becomes very complicated to split and apply the particle counting to recompute the cost -- particles are distributed on different ranks. Therefore the code resorts to a volume based divide that hopefully also divides the cost.

3. Now the strange split condition in 1. It helps to reduce the cases that occurs in 2 by splitting large volume filling nodes that are only partially occupied. The result is less overlaps of the initial refinement from different ranks.

If you get rid of the second refinement, the issue of expensive merged nodes is still there. 

I thought the domain cost computation really needs a full overhaul to solve the issue. (I may be wrong -- what did you do in the end?) I think a cleaner way is to globally sort and chop off the peano keys (possibly weighted by cost) - producing non-overlapping initial refinements which is then merged and possibly simplified globally. This will be a lot cleaner and provably even (esp if with a priority queue based balancer). I vaguely remember doing something like this in early days of PSphray.

rainwoodman:
I think this is the code I am talking about. https://github.com/rainwoodman/psphray2/blob/master/src/domain.c

sbird:
Wow. Thanks for the explanation - I will restore the second refinement and add a comment above the second refinement loop. I was able to solve the issue with a couple of tweaks - mostly by assigning segments to tasks by 'fill each task bucket at a time' instead of round robin, and by using a rolling average instead of a simple threshold to judge bucket fullness. That reduces the amount of state lost when we switched segments and made a balancing failure much less likely. It seems to work ok now.

sbird:
How does the algorithm you propose (morton keys sorted by priority) ensure locality of segments? The code before issue-86 often assigned multiple Segments to Tasks, round-robin, but did not ensure that a Task contained contiguous Segments, meaning that some the locality of the decomposition was poor and a lot of time was spent in tree-wait. Assigning topNodes purely using a cost based priority queue would suffer from the same problem, wouldn't it?

sbird:
Oh, one more thing - what about the condition in the second refine here: 
https://github.com/rainwoodman/MP-Gadget/blob/statictree/domain.c#L942

It uses topNodes[i].Size > 1, where the earlier refinement loop uses topNodes[i].Size > 8. It also does (topNodes[i].Size >> 3) within the loop body, which could lead to zero-sized topNodes if Size <= 8. I suspect this condition is never hit in practice, and should be changed just in case, but is there something I'm missing?

rainwoodman:
It is OK to have zero size nodes -- that indicates the PEANO key resolution is too low to even resolve the entire particle load. I think you can trigger it by reducing BITS. The condition simply stops refining after that since refinement becomes degenerate (8 nodes of the same size and key). But zero size nodes wouldn't otherwise cause any harm?

sbird:
In that case I will leave it alone. This branch seems to fix my problem, am I ok to merge to statictree after the run reaches z=0?

rainwoodman:
Sure!

rainwoodman:
I thought through and my solution doesn't guarentee balanced segments neither. However I think we can largely eliminate the redundant cost counting which adds about 30 mins to the initial domain decomposition of BlueTides -- by forming the TopNodes bottom up instead of top down. I'll try to work it out around this weekend from the head of static-tree branch; it will involve a mpsort on a struct of key and cost,  building a temporary oct tree down to single particles, computing the cost and trimming the branches to produce the initial local refinement. Better -- it is guarented to have minimal overlapping and I think very likely the second refinement actually needs to be supplemented with a combination. (due to extra fine leaves edges).

sbird:
Nice! Can you please use a temporary branch for it?

PR #88, opened 2017-08-14
Better local refine

rainwoodman:
This implements a drastically faster local refinement algorithm.

The main idea is to first build a full tree for a subsample of particles (current 1/16th). Then accurately count the cost and counts for each node in this tree quickly, and once. This accurate tree is then chopped off to produce the local refinement, such that each leave is smaller than the required granularity.

The secondary improvement is to avoid permuting the P data structure by the Key. We only need Key and Cost for each particle in most of the toptree meddling. Because the new data is much smaller, I required a global sorting with mpsort, this should improve the locality of the local trees, and improve the accuracy of the final merged tree.

The third improvement is to remove the poorly split logic between add_costs and merge_tree. The function is also better documented now.

The dm-only test case runs to redshift zero.

### DIFF
--- a/domain.c
+++ b/domain.c
@@ -762,6 +762,53 @@ domain_toptree_update_cost(struct local_topnode_data * topTree, int start)
     }
 }
 
+static void
+domain_toptree_truncate(struct local_topnode_data * topTree, int start, int64_t countlimit, int64_t costlimit)
+{
+    if(topTree[start].Daughter == -1) return;
+
+    if(topTree[start].Count < countlimit &&
+       topTree[start].Cost < costlimit) {
+        /* truncate here */
+        topTree[start].Daughter = -1;
+        return;
+    }
+
+    int j;
+    for(j = 0; j < 8; j ++) {
+        int sub = topTree[start].Daughter + j;
+        domain_toptree_truncate(topTree, sub, countlimit, costlimit);
+    }
+}
+
+/* remove the nodes that are no longer useful after the truncation.
+ * 
+ * We walk the topTree top-down to collect useful nodes, and move them to
+ * the head of the topTree list.
+ *
+ * This works because any child node is stored after the parent in the list 
+ * -- we are destroying the old tree just slow enough
+ * */
+
+static void
+domain_toptree_garbage_collection(struct local_topnode_data * topTree, int start, int * last_free)
+{
+
+    if(topTree[start].Daughter == -1) return;
+    int j;
+
+    int oldd = topTree[start].Daughter;
+    int newd = *last_free;
+
+    topTree[start].Daughter = newd;
+
+    (*last_free) += 8;
+    for(j = 0; j < 8; j ++) {
+        topTree[newd + j] = topTree[oldd + j];
+        domain_toptree_garbage_collection(topTree, newd + j, last_free);
+    }
+}
+
 /* check for local refinements with subsamples*/
 static int
 domain_check_for_local_refine_subsample(
@@ -823,7 +870,6 @@ domain_check_for_local_refine_subsample(
         domain_toptree_insert(topTree, LP[i].Key, LP[i].Cost);
     }
 
-    domain_toptree_update_cost(topTree, 0);
     return 0;
 }
 
@@ -1065,6 +1111,14 @@ int domain_determineTopTree(struct local_topnode_data * topTree)
 
 //    errflag = domain_check_for_local_refine(0, topTree, countlimit, costlimit, LP);
     errflag = domain_check_for_local_refine_subsample(topTree, LP, 16);
+
+    domain_toptree_update_cost(topTree, 0);
+
+    domain_toptree_truncate(topTree, 0, countlimit, costlimit);
+
+    NTopNodes = 1; /* put in the root node -- it's never a garbage . */
+    domain_toptree_garbage_collection(topTree, 0, &NTopNodes);
+
     walltime_measure("/Domain/DetermineTopTree/LocalRefine");
 
     if(NTopNodes > 4 * All.DomainOverDecompositionFactor * NTask * All.TopNodeIncreaseFactor) {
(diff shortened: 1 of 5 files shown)


domain.c
```diff
@@ -955,21 +1067,37 @@ int domain_determineTopTree(struct local_topnode_data * topNodes)
     /* FIXME: the previous function is already collective, so this step is not needed.
```
sbird: This comment is obsolete now I think

domain.c
```diff
@@ -681,103 +706,189 @@ domain_create_topleaves(int no, int * next)
...
+            last_key = LP[i].Key;
+            last_leaf = domain_toptree_insert(topTree, last_key, 0);
+            i += sample_step;
+            continue;
+        }
+    }
+    /* Now we are ready to insert all particles to the tree */
+    for(i = 0; i < NTopNodes; i ++ ) {
+        topTree[i].Count = 0;
+    }
```
sbird: Why do you need this?

domain.c
```diff
@@ -681,103 +706,189 @@ domain_create_topleaves(int no, int * next)
...
+
+    for(i = 0; i < NumPart; i ++ ) {
+        domain_toptree_insert(topTree, LP[i].Key, LP[i].Cost);
```
sbird: Please add a comment here

domain.c
```diff
@@ -874,78 +985,79 @@ int domain_nonrecursively_combine_topTree(struct local_topnode_data * topNodes)
...
-    sumup_large_ints(1, &NumPart, &TotNumPart);
-    MPI_Allreduce(&gravcost, &totgravcost, 1, MPI_INT64, MPI_SUM, MPI_COMM_WORLD);
+    walltime_measure("/Domain/DetermineTopTree/Misc");
 
-    int64_t costlimit, countlimit;
+    /* Watchout : must disgard proximity of particle type; this ordering is only required by LocalRefine */
 
-    costlimit = totgravcost / (All.TopNodeIncreaseFactor * All.DomainOverDecompositionFactor * NTask);
-    countlimit = TotNumPart / (All.TopNodeIncreaseFactor * All.DomainOverDecompositionFactor * NTask);
+    mpsort_mpi(LP, NumPart, sizeof(struct local_particle_data), mp_order_by_key, 8, NULL, MPI_COMM_WORLD);
```
sbird: Please add a comment here: I think this means that refinement is no longer using entirely local particles (which is fine) but that should be mentioned.

domain.c
```diff
@@ -874,78 +985,79 @@ int domain_nonrecursively_combine_topTree(struct local_topnode_data * topNodes)
  *  in pieces of eight segments until each segment holds at most a certain
  *  number of particles.
  */
-int domain_determineTopTree(struct local_topnode_data * topNodes)
+int domain_determineTopTree(struct local_topnode_data * topTree)
 {
     int i, j, sub;
     int errflag, errsum;
```
sbird: All but i is an unused variable now

domain.c
```diff
@@ -1238,6 +1347,13 @@ order_by_key(const void *a, const void *b)
     return 0;
```
sbird: This function is now never used - can be removed

domain.c
```diff
@@ -1065,109 +1186,97 @@ void domain_compute_costs(int64_t *TopLeafWork, int64_t *TopLeafCount)
...
-
-
 /* FIXME: this function needs some comments. I used to know what it does --. YF*/
-void domain_insertnode(struct local_topnode_data *treeA, struct local_topnode_data *treeB, int noA, int noB, struct local_topnode_data * topNodes)
+void
+domain_toptree_merge(struct local_topnode_data *treeA, struct local_topnode_data *treeB, int noA, int noB, struct local_topnode_data * topTree)
 {
     int j, sub;
     int64_t count, countA, countB;
     int64_t cost, costA, costB;
```
sbird: cost[AB] and count[AB] are unused.

sbird:
I have also run a larger test to z=16 (DM only), and confirm it speeds up domain by a factor of two (!) and does not crash, and produces pretty balanced domains: the largest load is consistently about 1.03.

As far as I'm concerned, merge it!

domain.c
```diff
@@ -681,103 +706,189 @@ domain_create_topleaves(int no, int * next)
...
+            /* split last leaf to hopefully make place for the new particle
+             * we will hold at most one sample particle per leaf */
+            if(0 != domain_toptree_split(topTree, leaf)) return 1;
+            /* reinsert the last particle */
+            topTree[leaf].Count = 0;
+            last_leaf = domain_toptree_insert(topTree, last_key, 0);
+            continue;
+        } else {
+            if(topTree[leaf].Count != 0) {
+                /* this shall not happen because key is already sorted. (cite paper)*/
```
rainwoodman: .Count is used as a flag to inidicate if the node has been finalized. There is a theorem says we never run into a finished node except the last one.

domain.c
rainwoodman: The old particles must be removed to make it a skeleton. Otherwise the counts won't be correct when we insert all particles.

domain.c
rainwoodman: The sorting has nothing to do with subsampling. It is required by the faster tree build algorithm. Global sorting enhances the accuracy of the final global topTree. Local sorting also produces correct results (but less accurate). I added comments to clarify this.

domain.c
rainwoodman: I'll add back the local sorting routine but disable it with a flag. I worry one may prefer local sorting in some cases (I can't yet think of any).

rainwoodman:
I did some clean up and rebased to the latest static tree.

domain.c
sbird: Sure, if you want to.

domain.c
```diff
@@ -99,6 +121,8 @@ static int domain_allocated_flag = 0;
  */
 void domain_decompose_full(void)
 {
+    sumup_large_ints(1, &NumPart, &TotNumPart);
+
     int retsum;
```
sbird: Unused. Sometimes I think you should add -Wall to your compile config :P

sbird:
It looks good. Go ahead and merge it when you are ready!

PR #89, opened 2017-08-15
Restart fixes

tvsananth:
Fixes to code before restarting BT2.

### DIFF
--- a/domain.c
+++ b/domain.c
@@ -1446,9 +1446,9 @@ int domain_check_for_local_refine(const int i, const double countlimit, const do
     /*If there are only 8 particles within this node, we are done refining.*/
     if(topNodes[i].Size < 8)
         return 0;
-
-    /* already have enough leaves */
-    if(NTopleaves > All.DomainOverDecompositionFactor * NTask * TOPNODEFACTOR) {
+    
+    /* already have enough nodes */
+    if(NTopnodes > All.DomainOverDecompositionFactor * NTask * TOPNODEFACTOR) {
         return 0;
     }
 
@@ -1472,7 +1472,6 @@ int domain_check_for_local_refine(const int i, const double countlimit, const do
     /*Make a new topnode section attached to this node*/
     topNodes[i].Daughter = NTopnodes;
     NTopnodes += 8;
-    NTopleaves += 7; /* because the current node is removed from leaves */
 
     /* Initialise this topnode with new sub nodes*/
     for(j = 0; j < 8; j++)
@@ -1649,7 +1648,6 @@ int domain_determineTopTree(void)
     walltime_measure("/Domain/DetermineTopTree/Sort");
 
     NTopnodes = 1;
-    NTopleaves = 1;
     topNodes[0].Daughter = -1;
     topNodes[0].Parent = -1;
     topNodes[0].Size = PEANOCELLS;
--- a/domain.c
+++ b/domain.c
@@ -1443,6 +1443,7 @@ int domain_check_for_local_refine(const int i, const double countlimit, const do
 {
     int j, p;
 
+    /* Skipping local refinement completely as it slows down BT-II */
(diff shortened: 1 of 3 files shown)


domain.c
```diff
@@ -1446,9 +1446,9 @@ int domain_check_for_local_refine(const int i, const double countlimit, const do
     /*If there are only 8 particles within this node, we are done refining.*/
     if(topNodes[i].Size < 8)
         return 0;
-
-    /* already have enough leaves */
-    if(NTopleaves > All.DomainOverDecompositionFactor * NTask * TOPNODEFACTOR) {
+    
+    /* already have enough nodes */
+    if(NTopnodes > All.DomainOverDecompositionFactor * NTask * TOPNODEFACTOR) {
```
rainwoodman: I think you shall return 0 before 1447 to totally by pass this function.

proto.h
```diff
@@ -92,6 +92,10 @@ int MPI_Alltoallv_sparse(void *sendbuf, int *sendcnts, int *sdispls,
         MPI_Datatype sendtype, void *recvbuf, int *recvcnts,
         int *rdispls, MPI_Datatype recvtype, MPI_Comm comm);
 
+int MPI_Alltoallv_fix_buserror(void *sendbuf, int *sendcnts, int *sdispls,
```
rainwoodman: Let's call it `MPI_Alltoallv_throttled()` -- we don't know if it fixes the buserror or not, and 2 years from now we won't even remember any details about the bus error on BlueWaters..

system.c
```diff
@@ -291,78 +291,40 @@ int MPI_Alltoallv_sparse(void *sendbuf, int *sendcnts, int *sdispls,
...
-        if(sendcnts[target] == 0 && recvcnts[target] == 0) continue;
-        MPI_Sendrecv(((char*)sendbuf) + send_elsize * sdispls[target], 
-                sendcnts[target], sendtype, 
-                target, 101934,
-                ((char*)recvbuf) + recv_elsize * rdispls[target],
-                recvcnts[target], recvtype, 
-                target, 101934, 
-                comm, MPI_STATUS_IGNORE);
-
+        return MPI_Alltoallv(sendbuf, sendcnts, sdispls,
```
rainwoodman: Be aware -- this is not using MPI_Alltoallv_throttled yet...

system.c
```diff
@@ -291,78 +291,40 @@ int MPI_Alltoallv_sparse(void *sendbuf, int *sendcnts, int *sdispls,
...
-#endif
-    /* ensure the collective-ness */
-    MPI_Barrier(comm);
 
-    return 0;
+int MPI_Alltoallv_fix_buserror(void *sendbuf, int *sendcnts, int *sdispls,
+        MPI_Datatype sendtype, void *recvbuf, int *recvcnts,
+        int *rdispls, MPI_Datatype recvtype, MPI_Comm comm)
+{
+   int i, j, ncut;
```
rainwoodman: Please define j as int64_t. The multiplication around line 320 may overflow an integer.

rainwoodman: We can make this function smarter -- ncut can be estimated from the total number of items exchanged and an upper bound. This way the smaller exchanges are not divided and can probably be done more efficiently. I suspect it won't make a huge difference; so I'll leave it up to you for now.

domain.c
```diff
@@ -1443,12 +1443,14 @@ int domain_check_for_local_refine(const int i, const double countlimit, const do
 {
     int j, p;
 
-    /*If there are only 8 particles within this node, we are done refining.*/
+    return 0;
```
rainwoodman: Please add a comment here for future reference, e.g. "Skipping local refinement because it gets horribly slow on BT-II".

rainwoodman:
Looks good to me. Other than a missing comment. Please go ahead a merge!

PR #90, opened 2017-08-16
Genic cosmology

sbird:
This pull request makes changes to GenIC.

The main functional change is that the growth function, F_Omega, from MP-Gadget's cosmology.c is used for the Zel'dovich velocities. This is a clear win in terms of IC accuracy, as it includes radiation corrections. 

To make this possible I first mess around with cosmology.c to make it self-contained (up to 1f9b471).

I then do a whole bunch of cleanups on GenIC, including moving it to its own subdirectory and removing various vestigial features. The only one that makes any difference is f07b078, which removes SphereMode (the default). The effect is small, but my tests suggest that using the full cubic set of modes improves agreement with the input power spectrum (not unexpectedly).


genic/params.c
```diff
@@ -82,40 +81,44 @@ void read_parameterfile(char *fname)
...
     
-    Omega = param_get_double(ps, "Omega0");
+    CP.Omega0 = param_get_double(ps, "Omega0");
+    CP.OmegaLambda = param_get_double(ps, "OmegaLambda");
+    CP.OmegaBaryon = param_get_double(ps, "OmegaBaryon");
+    CP.HubbleParam = param_get_double(ps, "HubbleParam");
+    CP.CMBTemperature = param_get_double(ps, "CMBTemperature");
+    CP.RadiationOn = param_get_double(ps, "RadiationOn");
+    /*This should be removed if massive neutrinos are implemented*/
+    CP.MasslessNeutrinosOn = 1;
```
rainwoodman: I am confused by the comment.

genic/save.c
```diff
@@ -125,13 +125,13 @@ void saveheader(BigFile * bf) {
             (big_block_set_attr(&bheader, "Time", &InitTime, "f8", 1)) ||
             (big_block_set_attr(&bheader, "Redshift", &redshift, "f8", 1)) ||
             (big_block_set_attr(&bheader, "BoxSize", &Box, "f8", 1)) ||
-            (big_block_set_attr(&bheader, "OmegaM", &Omega, "f8", 1)) ||
-            (big_block_set_attr(&bheader, "OmegaB", &OmegaBaryon, "f8", 1)) ||
-            (big_block_set_attr(&bheader, "OmegaL", &OmegaLambda, "f8", 1)) ||
+            (big_block_set_attr(&bheader, "OmegaM", &CP.Omega0, "f8", 1)) ||
```
rainwoodman: I recently started to like the CLASS set of names, LAMBDA, CDM, B, UR, NCDM, and NCDM_TOT. The word neutrino doesn't even appear.

genic/params.c
```diff
@@ -30,36 +22,43 @@ create_parameters()
...
-    param_declare_int(ps, "WDM_Vtherm_On", REQUIRED, 0, "");
-    param_declare_double(ps, "WDM_PartMass_in_kev", REQUIRED, 0, "");
+    param_declare_int(ps, "WhichSpectrum", REQUIRED, 2, "Type of spectrum, 2 for file ");
+
+    param_declare_double(ps, "MaxMemoryPerCore", OPTIONAL, 1300., "");
+    param_declare_double(ps, "CMBTemperature", OPTIONAL, 2.7255, "CMB temperature in K");
+    param_declare_double(ps, "RadiationOn", OPTIONAL, 1, "Include radiation in the background.");
+    param_declare_int(ps, "UsePeculiarVelocity", OPTIONAL, 0, "Write a IC similiar to a FastPM output");
+    param_declare_double(ps, "Sigma8", OPTIONAL, -1, "Renormalise Sigma8 to this number if positive");
+    param_declare_int(ps, "PowerIsRedshiftZero", OPTIONAL, 1, "If true, rescale the power spectrum by the linear growth factor.");
```
rainwoodman: What about InputPowerRedshift? Then backscaling is always applied from the inputpower redshift?

Have you thought about taking a transfer function + primordial power input for multiple species IC?

sbird: That's a good idea (to the first). To the second - see below.

genic/params.c
sbird: You only add massless neutrinos if there are no massive neutrinos - if MP-GenIC supported adding massive neutrinos to the ICs you would want to set CP.MasslessNeutrinosOn = 0.

genic/save.c
sbird: Ye-es...but neutrinos I think interpolate between UR and CDM (or is it NCDM?) for CLASS, depending on their temperature, which isn't what we want here. In S-GenIC I have Omega, OmegaLambda and OmegaBaryon, which I like. Then I specify the total neutrino mass and the hierarchy and compute OmegaNeutrino internally.

sbird:
While writing this I pondered the relation of MP-GenIC to S-GenIC - my personal IC code. At the moment the two codes have disjoint feature sets. MP-GenIC has very scalable MPI parallelization, but it lacks S-GenIC cosmology features:
- 2LPT
- Multiple transfer functions
- Massive neutrinos
S-GenIC is also a (very) C++ code, which uses Boost::Test for a test suite (this was probably a mistake, but there were far fewer C++ test suites easily available then).

Adding S-GenIC's features to MP-GenIC would constitute essentially a complete rewrite. Multiple transfer functions in particular are rather invasive, because the grid generation in save.cpp has to be decoupled from the actual file saving. One could either rewrite the code in C (which is a lot of work for no real gain) or merge the C++ code as-is (at which point one essentially replaces MP-GenIC with S-GenIC). In addition, the parameter files are completely different, so there is no backwards compatibility.
I partially ported the transfer function code into MP-GenIC, but the result was ugly.

So my proposed solution is:
- keep both codes around
- keep MP-GenIC as a simple C-only IC generator, as it is now.
- keep S-GenIC as a more complex C++ IC generator with various advanced or specialised cosmological features, and port MP-GenIC's MPI parallelization to it.

Then we have one code that is easy to follow and does the basic thing that a lot of people still want very well, and one code that does more cosmology, at the cost of complexity. What do you think?

genic/save.c
sbird: I don't really want to change the param names because it would break compatibility with all the existing config files...

rainwoodman:
I see -- I must have been naive about transfer function ICs; so I'll believe your words on it.

I thought it is only a matter of replacing power spectrum with transfer function, and add a loop for each species. Was the decoupling for saving the number of FFTs? It doesn't really matter to reduce the number of FFTs in IC gen because the simulation will run hell a lot more of them.

sbird:
It is just that - and it's not about saving FFT numbers. You have to:

a) Rewrite power.c (This is a bigger rewrite than it looks because the transfer function and power spectrum may be binned differently. So you need gsl interpolation)
b) Add a main loop in main.c
c) Rewrite much of save.c to move ProduceGas out of it, because the unshifted Pos for gas is no longer the same as the unshifted pos for DM.
and at this point you have rewritten half of GenIC. If we then wanted to add neutrinos or relative velocities or 2LPT to MP-GenIC then that is more work.

None of this is difficult exactly - it is just that I already did it, and so it is work for no specific reason. 
Since S-GenIC is then likely to persist in some form, we should decide which features MP-GenIC really needs. My laptop runs gnome now so I vote for only the most basic of them :)

That said, multiple transfer functions is more generally useful than the others - I'm not sure that 2LPT makes sense when there is a separate baryon fluid, and particle neutrinos are rather niche, if not obsolete. So if you think you would use multiple transfer functions for something, I could re-add it to the todo list.

rainwoodman:
Sure. Let's keep MP-GenIC as simple and hackable as possible then.

sbird:
Ok to merge?

rainwoodman:
Yes!

PR #93, opened 2017-09-07
Remove requirement particles sorted by type before IO

rainwoodman:
This remove the requirement that particles must be sorted by type before IO.

### DIFF
--- a/examples/lya/paramfile.gadget
+++ b/examples/lya/paramfile.gadget
@@ -49,7 +49,7 @@ MaxRMSDisplacementFac = 0.2
 CourantFac = 0.15     
 
 MaxSizeTimestep = 0.1
-MinSizeTimestep = 0.05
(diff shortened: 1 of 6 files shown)


sbird:
Ok, I'm curious - this seems fine, but why did you want to do it? It was only implemented a few weeks ago, at your suggestion.

rainwoodman:
i realized we cannot assume the particles are sorted by type due to GC.

rainwoodman:
I noticed this because when i ran lya on static tree it always crashed
without this fix.

On Sep 9, 2017 3:38 PM, "Feng Yu" <rainwoodman@gmail.com> wrote:

i realized we cannot assume the particles are sorted by type due to GC.

sbird:
Makes sense. I guess I hadn't tested far enough to get star formation just before a snapshot.

PR #96, opened 2017-09-08
Auto alltoall.

rainwoodman:
This PR uses smart all to all in all places. In principle it shall make the code run a bit faster. But maybe not. The heuristics is to strike a balance between memory footprint of the MPI library and speed. Really should have been implemented inside MPI but different vendors have different philosophy in `MPI_Alltoallv`.


rainwoodman:
According to Ananth, this wouldn't helped much because ``MPI_alltoallv`` from Cray crashes with hugepage error too. We need to find a way to reduce memory fragmentation.

PR #97, opened 2017-09-11
WIP: Forcetree test

rainwoodman:
Filing a PR, such that it is easier to compare the difference.

### DIFF
--- a/forcetree.c
+++ b/forcetree.c
@@ -192,6 +192,7 @@ static void init_internal_node(struct NODE *nfreep, struct NODE *parent, int sub
             const MyFloat lenhalf = 0.25 * parent->len;
 
             nfreep->len = 0.5 * parent->len;
+            nfreep->f.TopLevel = 0;
 
             for(j = 0; j < 3; j++) {
                 /* Detect which quadrant we are in by testing the bits of subnode:
@@ -301,6 +302,7 @@ int force_tree_create_nodes(const int firstnode, const int lastnode, const int n
         for(i = 0; i < 8; i++)
             nfreep->u.suns[i] = -1;
         nfreep->father = -1;
+        nfreep->f.TopLevel = 1;
         nnext++;
         /* create a set of empty nodes corresponding to the top-level domain
          * grid. We need to generate these nodes first to make sure that we have a
@@ -486,6 +488,8 @@ void force_create_node_for_topnode(int no, int topnode, int bits, int x, int y, 
                     Nodes[*nextfree].center[1] = Nodes[no].center[1] + (2 * j - 1) * lenhalf;
                     Nodes[*nextfree].center[2] = Nodes[no].center[2] + (2 * k - 1) * lenhalf;
                     Nodes[*nextfree].father = no;
+                    /*All nodes here are top level nodes*/
+                    Nodes[*nextfree].f.TopLevel = 1;
 
                     for(n = 0; n < 8; n++)
                         Nodes[*nextfree].u.suns[n] = -1;
@@ -654,7 +658,6 @@ force_update_node_recursive(int no, int sib, int tail, const int firstnode, cons
     memset(&Nodes[no].u.d.s,0,3*sizeof(MyFloat));
     Nodes[no].u.d.mass=0;
     Nodes[no].u.d.hmax=0;
-    Nodes[no].f.TopLevel=0;
     Nodes[no].f.DependsOnLocalMass=0;
     Nodes[no].f.MixedSofteningsInNode=0;
     Nodes[no].f.InternalTopLevel=0;
@@ -919,18 +922,6 @@ force_flag_localnodes(void)
 
     for(i = 0; i < NTopLeaves; i++)
     {
-        no = TopLeaves[i].treenode;
-
-        while(no >= 0)
-        {
-            if(Nodes[no].f.TopLevel)
-                break;
-
-            Nodes[no].f.TopLevel =1;
-
-            no = Nodes[no].father;
-        }
-
         /* mark also internal top level nodes */
 
         no = TopLeaves[i].treenode;
--- a/forcetree.c
+++ b/forcetree.c
@@ -194,6 +194,26 @@ int get_subnode(const struct NODE * node, const int nodepos, const int p_i, cons
     return subnode;
 }
 
+/*Initialise an internal node at nfreep. The parent is assumed to be locked, and
+ * we have assured that nothing else will change nfreep while we are here.*/
+static void init_internal_node(struct NODE *nfreep, struct NODE *parent, int subnode)
+{
+            int j;
+            const MyFloat lenhalf = 0.25 * parent->len;
+
+            nfreep->len = 0.5 * parent->len;
+
+            for(j = 0; j < 3; j++) {
+                /* Detect which quadrant we are in by testing the bits of subnode:
+                 * if (subnode & [1,2,4]) is true we add lenhalf, otherwise subtract lenhalf*/
+                const int sign = (subnode & (1 << j)) ? 1 : -1;
+                nfreep->center[j] = parent->center[j] + sign*lenhalf;
+            }
+            for(j = 0; j < 8; j++)
+                nfreep->u.suns[j] = -1;
+            nfreep->hmax = 0;
+}
+
 /* Size of the free Node thread cache.
  * 100 was found to be optimal for an Intel skylake with 4 threads.*/
 #define NODECACHE_SIZE 100
@@ -326,22 +346,10 @@ int force_tree_create_nodes(const int firstnode, const int lastnode, const int n
                 break;
             }
             struct NODE *nfreep = &Nodes[ninsert];	/* select desired node */
-            int j;
-            const MyFloat lenhalf = 0.25 * Nodes[this].len;
+            init_internal_node(nfreep, &Nodes[this], subnode);
 
-            nfreep->len = 0.5 * Nodes[this].len;
-
-            for(j = 0; j < 3; j++) {
-                /* Detect which quadrant we are in by testing the bits of subnode:
-                 * if (subnode & [1,2,4]) is true we add lenhalf, otherwise subtract lenhalf*/
-                const int sign = (subnode & (1 << j)) ? 1 : -1;
-                nfreep->center[j] = Nodes[this].center[j] + sign*lenhalf;
-            }
-            for(j = 0; j < 8; j++)
-                nfreep->u.suns[j] = -1;
-            nfreep->hmax = 0;
-
-            /*Re-add the particle to the new internal node*/
+            /* The new internal node replaced a particle in the parent.
+             * Re-add that particle to the child.*/
             const int child_subnode = get_subnode(nfreep, this, child, shift -3, MortonKey);
 
             nfreep->u.suns[child_subnode] = child;
--- a/forcetree.c
+++ b/forcetree.c
@@ -230,24 +230,6 @@ int insert_internal_node(int parent, int subnode, int p_child, int p_toplace, co
 
     struct NODE *nfreep = &Nodes[ninsert];
     struct NODE *nprnt = &Nodes[parent];
-    /* If the node is very small, just add the particle to the first empty subnode.
-     * If there are no empty subnodes, we will have to create a new node: if this
-     * happens repeatedly we probably have bigger problems
-     * (like: some corruption in domain exchange which is zeroing all the positions).*/
-    if (nprnt->len < minlen) {
-        int j;
-        for(j=0; j<8; j++)
-            if(nprnt->u.suns[j] < 0) {
-                nprnt->u.suns[j] = p_toplace;
-                Father[p_toplace] = parent;
-                return 0;
-            }
-        /* Warn about this if we have more than 4 particles
-         * attached to this very small node.*/
-        if(j > 4)
-            message(1,"Particle %d wants to attach to small node at %g %g %g (len %g), where we already have %d particles.\n",
-                    p_toplace, nprnt->center[0], nprnt->center[1], nprnt->center[2], nprnt->len, j);
-    }
     /* We create a new internal node with empty subnodes at the end of the array, and
      * use it to replace the particle in the parent's subnode.*/
     init_internal_node(nfreep, nprnt, subnode);
@@ -258,7 +240,25 @@ int insert_internal_node(int parent, int subnode, int p_child, int p_toplace, co
      * Re-add that particle to the child.*/
     const int child_subnode = get_subnode(nfreep, p_child);
 
-    const int new_subnode = get_subnode(nfreep, p_toplace);
+    int new_subnode = get_subnode(nfreep, p_toplace);
+
+    /* If the node is very small, just add the particle to the next subnode.
+     * This ensures that we can always construct the tree, and only happens far
+     * below the softening length. This happens occasionally, but if it is
+     * frequent increase the force softening.
+     */
+    if(nfreep->len < minlen && new_subnode == child_subnode) {
+        if(child_subnode < 7)
+            new_subnode = child_subnode + 1;
+        else
+            new_subnode = child_subnode - 1;
+        /* Warn about this happening.*/
+        message(1,"Close particles: %d @ [%g, %g, %g] and %d @ [%g, %g, %g]. "
+                "Attached to node %d, subnode %d, at [%g, %g, %g] (len %g).\n",
+                p_toplace, P[p_toplace].Pos[0], P[p_toplace].Pos[1], P[p_toplace].Pos[2],
+                p_child, P[p_child].Pos[0], P[p_child].Pos[1], P[p_child].Pos[2],
+                nfreep, child_subnode, nprnt->center[0], nprnt->center[1], nprnt->center[2], nprnt->len);
+    }
 
     int ret = 0;
     /*If these two are different, great! Attach both particles to this new node*/
--- a/forcetree.c
+++ b/forcetree.c
@@ -46,9 +46,12 @@ int *Father;			/*!< gives parent node in tree (nodes array) */
 static int tree_allocated_flag = 0;
 
 static int force_tree_build(int npart);
-static int force_tree_build_single(int firstnode, int lastnode, int npart);
 
-static void
+/*Next two are not static as tested.*/
+int
+force_tree_build_single(const int firstnode, const int lastnode, const int npart);
+
+size_t
 force_treeallocate(int maxnodes, int maxpart, int first_node_offset);
 
 static void
@@ -107,7 +110,11 @@ int force_tree_build(int npart)
         maxnodes = All.TreeAllocFactor * All.MaxPart + NTopNodes;
         /* construct tree if needed */
         /* the tree is used in grav dens, hydro, bh and sfr */
-        force_treeallocate(maxnodes, All.MaxPart, All.MaxPart);
+        size_t allbytes = force_treeallocate(maxnodes, All.MaxPart, All.MaxPart);
+
+        message(0, "Allocated %g MByte for BH-tree, (presently allocated %g MB)\n",
+             allbytes / (1024.0 * 1024.0),
+             AllocatedBytes / (1024.0 * 1024.0));
 
         Numnodestree = force_tree_build_single(All.MaxPart, All.MaxPart + maxnodes, npart);
         if(Numnodestree < 0)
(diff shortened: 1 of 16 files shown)


test/forcetree_test.c
```diff
@@ -0,0 +1,346 @@
...
+    }
+    qsort(P, numpart, sizeof(struct particle_data), order_by_type_and_key);
+    int maxnode = numpart;
+    assert_true(Nodes);
+    /*So we know which nodes we have initialised*/
+    for(int i=0; i< MaxNodes+1; i++)
+        Nodes_base[i].father = -2;
+    /*Time creating the nodes*/
+    struct timespec start, end;
+    clock_gettime(CLOCK_MONOTONIC,&start);
```
rainwoodman: MPI_Wtime is more portable than clock_gettime.

forcetree.c
```diff
@@ -144,57 +162,134 @@ int force_tree_build(int npart)
...
- *  points to the first tree node, while `nodes' is shifted such that
- *  nodes[All.MaxPart] gives the first tree node. Finally, node indices
- *  with values 'All.MaxPart + MaxNodes' and larger indicate "pseudo
- *  particles", i.e. multipole moments of top-level nodes that lie on
- *  different CPUs. If such a node needs to be opened, the corresponding
- *  particle must be exported to that CPU. */
-int force_tree_build_single(int npart)
+/*Get a pointer to memory for a free node, from our node cache.
+ * If there is no memory left, return NULL.*/
+int get_freenode(int * nfree, int *nfree_thread, int *numfree_thread)
```
rainwoodman: nfree_thread and numfree_thread looked confusing... Is there a better name?

forcetree.c
```diff
@@ -498,11 +581,57 @@ force_get_prev_node(int no)
...
+    else
+    {
+        if(All.ForceSoftening[new_type] > pnode->u.d.maxsoft)
+            pnode->u.d.maxsoft = All.ForceSoftening[new_type];
+    }
+#endif
+}
+
+static void
+force_set_particle_moment_on_node(struct NODE * pnode, const struct particle_data * pa)
```
rainwoodman: add_particle_moment_to_node?

forcetree.c
```diff
@@ -792,9 +826,12 @@ void force_exchange_pseudodata(void)
             Nodes[no].u.d.s[1] = TopLeafMoments[i].s[1];
             Nodes[no].u.d.s[2] = TopLeafMoments[i].s[2];
             Nodes[no].u.d.mass = TopLeafMoments[i].mass;
-            Nodes[no].hmax = TopLeafMoments[i].hmax;
-            Nodes[no].u.d.bitflags =
-                (Nodes[no].u.d.bitflags & (~BITFLAG_MASK)) | (TopLeafMoments[i].bitflags & BITFLAG_MASK);
+            Nodes[no].u.d.hmax = TopLeafMoments[i].hmax;
+            Nodes[no].u.d.DependsOnLocalMass = TopLeafMoments[i].DependsOnLocalMass;
```
rainwoodman: Watch out!

forcetree.c
```diff
@@ -1113,26 +1134,22 @@ void force_update_hmax(int * activeset, int size)
  *  maxnodes approximately equal to 0.7*maxpart is sufficient to store the
  *  tree for up to maxpart particles.
  */
-void force_treeallocate(int maxnodes, int maxpart)
+size_t force_treeallocate(int maxnodes, int maxpart, int first_node_offset)
```
rainwoodman: I was thinking of adding a new node type --  which can either point to a particle or a pseudo particle, or a internal node, or nothing.

sbird: So these things would be actual pointers instead of these slightly ridiculous semi-pointers? I thought of that too, but the churn would have been very great at this stage, so I opted to defer it. But it sounds a good long-term goal: I found at least one place where it would allow less memory allocation and I wondered if it would let the compiler do more clever things.

forcetree.c
```diff
@@ -27,13 +27,17 @@
  *  reconstruct the tree every timestep.
  */
 
+/*The node index is an integer with unusual properties:
+ * no = 0..All.MaxPart (firstnode in internal functions) corresponds to a particle.
+ * no = All.MaxPart..All.MaxPart + MaxNodes (firstnode..lastnode) corresponds to actual tree nodes,
```
rainwoodman: So we have internal tree nodes, node-like leaf tree nodes, and real leaf nodes that are particles / pseudo partilces.
To distinguish them, we shall probably call the particles / pseudo particles ParticleNodes.

Currently, the address of the node is used to encode the type of the node. I think we can introduce a struct to clean this up. It is 4 bytes extra but likely worth it (also allows easier removal of particles while preserving the tree structure).
```
typedef union {
   unsigned int IsParticle :1;
   int index :31; /* 2 Billion particle + Pseudo particles at most. */
} NodeRef;

const NodeRef REMOVED_NODE = {.IsParticle  = 1, index = -1}; /* for example */

```

forcetree.c
rainwoodman: We shall still use the indices, but adding a level of indirection. See my comment below. I agree we shall defer till this is merged and correct, but good to keep this in mind while working other things out.

test/forcetree_test.c
sbird: Fixed in 3186c2b

forcetree.c
sbird: nnext and nrem?

forcetree.c
sbird: If we are going to do this, why not make them real pointers?

typedef union {
   unsigned int IsParticle :1;
   NODE * index :31; /* 2 Billion particle + Pseudo particles at most. */
} NodeRef;

const NodeRef REMOVED_NODE = {.IsParticle  = 1, index = -1}; /* for example */

and then do:
if(IsParticle) (struct particle_data) NodeRef
to access it

Although this is almost the classic case for C++ type inheritance really.

forcetree.c
sbird: Good catch!

forcetree.c
rainwoodman: Pointer requires special care for serialization. Indices are machine independent. 

I think this is different from classical c++ type inheritance -- the type-specific data here is floating; while in C++ the subclass data is embedded. We manage the type-specific data together. More or less like an placement new in C++ ...

sbird:
The current status of this branch is broken. I added a test which is wrong and does not pass (commit 30b6184).

In addition, there is a yet unfound bug which is causing it to occasionally fail to add a particle (probably an off-by-one somewhere).

However, I have found two actual bugs in the current tree: commit 4f2bf201 will cause the tree build to hang sometimes in OpenMP code and commit 1483c37 causes the tree to be built wrongly under some circumstances. I can cherry-pick them if you are about to do some actual simulations, or we can wait until this is merged.

sbird:
Can you suggest to me what is broken in 30b6184? I want to walk all the nodes, and I do it by starting at All.MaxPart and then repeatedly doing force_get_next_node, but this seems to work so badly I must be missing something.

rainwoodman:
We are running a simulation this week -- So please cherry-pick those two fixes to statictree.

rainwoodman:
I am looking into 30b6184.

sbird:
Cherry-picked.

sbird:
Fixed it!

rainwoodman:
Yes. I was getting there. But the counting is still not matching.

sbird:
Apologies, I overwrote your commit...

rainwoodman:
It's ok -- I practised some merging.

sbird:
Line 127 needs to be fnode >=0: otherwise we stop when we reach particle 0, which is too soon. It fixes the counting.

rainwoodman:
Is there a reason why the tests are not linked against endrun.c?

sbird:
Not a good one - I think I was just trying to avoid calling MPI_Init.

rainwoodman:
Are you sure aa40fa1d25d070f8efdd94e3db83c7978db100dc is correct? I am seeing a tree build error (creating too many nodes) on pressure-entropy-init. Scanning through the code there is a 'shift -=3' immediately after the first get_subnode.

sbird:
Don't worry, I'm working on something and can do it.

sbird:
(rebased onto statictree)

sbird:
This push removes ADAPTIVE_GRAVSOFT_FORGAS and replaces it with a runtime option.

sbird:
And oops, I see I broke things.

forcetree.c
```diff
@@ -221,152 +313,143 @@ int force_tree_build_single(int npart)
...
+        if(nnext_thread >= lastnode-1)
             continue;
 
         /*First find the Node for the TopLeaf */
 
-        int shift;
-        int topleaf = domain_get_topleaf_with_shift(P[i].Key, &shift);
-        int this = TopLeaves[topleaf].treenode;
+        int this;
+        if(inside_node(&Nodes[this_acc], i)) {
```
rainwoodman: Did you finally get rid of 'shift'. I know it is implied during the tree descent but dared not touch it.. How did it work out for you?

forcetree.c
```diff
@@ -501,11 +584,45 @@ force_get_prev_node(int no)
         return 0;
     }
 }
+
+/*Sets the node softening on a node.*/
+static void
+force_set_node_softening(struct NODE * pnode, const int new_type, const double hsml)
+{
+    if(pnode->f.MaxSofteningType == 7)
```
rainwoodman: I don't quite get why we need MaxSofteningType. If you recall anything, could you add a comment in the struct definition before you forget?

rainwoodman:
Shall we land statictree to master before landing this to statictree?

forcetree.c
sbird: I got rid of it - I got rid of the Morton key stuff entirely and just compared the particle position to the node center directly. It turned out that computing the Morton keys was much slower than the speedup from using them. So the tree build got faster by a factor two without them.

forcetree.c
sbird: Maybe we don't - let me think about it a bit.

sbird:
At the moment this doesn't work - but I also have a mysterious tree walk hang with statictree master. Let me see which one I figure out first.

forcetree.c
rainwoodman: This is very surprising. Were you recomputing the morton key before the comparison? the morton key based approach is usually believed to be much faster, since it is only a few bit operations; the position is a bunch of floats and touches more memory bytes.

rainwoodman:
I can look at the tree walk hang too if you point to me a reproducer.

forcetree.c
sbird: The Morton Key was computed in a separate loop over each particle, once per tree walk, and stored in temporary memory. But I found that this was fairly slow, and computation of the Morton Key was fairly slow, and the overhead from it dominated the comparison time. 

I suspect this is a consequence of processor prefetching - at the time when the subnode is being computed the particle structure is already in cache because we read P[i].Key and this brings in P[i].Pos as well. But Morton[i] is stored elsewhere and was not prefetched, so the speedup from it being fewer bits was small.

I also thought the absolute fastest way to do this would be to use the peano-hilbert-key, which is pre-computed already and already in cache. But I could not figure out how to do this - it requires knowing the rotation at all levels.

sbird:
I don't have one yet - it is very intermittent and occurring about once an hour.

rainwoodman:
hmm. What stops this one from working?

sbird:
Very occasionally I get a message from gravpm that a particle was not assigned to a region:

For example with the dm-only run, with 4 MPIs and 1 thread, I get this:

[ 003047.34 ] Begin Step 396, Time: 0.480059, Redshift: 1.08308, Nf = 00000002097152, Systemstep: 0.00102898, Dloga: 0.00214575, status: PM-Step
[ 003047.34 ] TotNumPart: 0000002097152 SPH 0000000000000 BH 0000000000 STAR 0000000000000 
[ 003047.34 ] Occupied:            0            1            2            3            4            5 dt
[ 003047.34 ]  X bin=12            0      1769410            0            0            0            0 0.00429149
[ 003047.34 ]  X bin=11            0       327742            0            0            0            0 0.00214575
[ 003047.34 ]                -----------------------------------
[ 003047.34 ] Total:               0      2097152            0            0            0            0  Sum:       2097152
[ 003047.34 ] Start force computation...
[ 003047.37 ] max number of regions is 35
[ 003047.39 ] Task 1: i = 123733 not assigned to a region
[ 003047.39 ] Task 1: Processed only 130614 particles out of 130615

I have applied the following patch to the paramfile:
diff --git a/examples/dm-only/paramfile.gadget b/examples/dm-only/paramfile.gadget
index 09236f0..186d3c1 100644
--- a/examples/dm-only/paramfile.gadget
+++ b/examples/dm-only/paramfile.gadget
@@ -33,10 +33,6 @@ RadiationOn = 0
 HydroOn = 0
 BlackHoleOn = 0
 
-# Accuracy of time integration
-MaxSizeTimestep = 0.1
-MinSizeTimestep = 0.01
-
 SnapshotWithFOF = 1
 FOFHaloLinkingLength = 0.2
 FOFHaloMinLength = 32

rainwoodman:
How many ranks and threads?
On the forcetree-test branch?

Regions are tree nodes that covers all particles--so this suggests a particle is missing from the tree ( A strange Father node, might be a wrong internaltoplevel flag?)

The 3000 second run is a bummer... maybe try to restart from a low redshift snapshot?
Wasn't there a tree rebuild before the PM? If so, dump all particles and rebuild the tree may reproduce this quickly?

sbird:
This is with 4 ranks and 1 thread, on forcetree-test branch.

Restarting from a low redshift snapshot is in progress. Then I will try the particle dump.

sbird:
Restarting from a lower snapshot works, but the lowest one I have is 0.3333 (and it stops crashing if I insert another one), so it still take 20 minutes of walltime to crash. Trying to get a particle dump.

rainwoodman:
I worry it may have to do with rounding off error telling if a particle is inside a node.. Adding a print on the particle's position and the attributes of the Father nodes might be useful. (We probably want to have a function for pretty printing a node and a function for pretty printing a particle).

sbird:
This is what I have so far:
[ 001225.80 ] Task 1: i = 123733 not assigned to a region Father=1108407 Nextnode=107008 ID = 597390
[ 001225.80 ] Task 1: Properties of parent node: 1108407 next: 1108408 sib 123735 mass 0 TL 0

sbird:
(gdb) p P[123733]
$1 = {
  SpinLock = 1, 
  GravCost = 384, 
  Ti_drift = 348160, 
  Ti_kick = 347136, 
  Pos = {19782.565989498609, 16828.527480302801, 11535.292632799374}, 
  Mass = 0.122035258, 
  {
    Evaluated = 1, 
    DensityIterationDone = 0, 
    OnAnotherDomain = 0, 
    WillExport = 0, 
    Type = 1, 
    TimeBin = 11 '\v', 
    Generation = 0 '\000', 
    IsNewParticle = 0, 
    Swallowed = 0
  }, 
  PI = 53108, 
  ID = 597390, 
  Vel = {-420.48097171447461, -266.4431856135318, -169.32867961900027}, 
  GravAccel = {-95.362102768113346, -19.742962917980762, -26.161324540865987}, 
  GravPM = {0, 0, 0}, 
  OldAcc = 0.01509617443443154, 
  Potential = -1308623.3947574615, 
  PM_Potential = 0, 
  StarFormationTime = 0, 
  Metallicity = 0, 
  Hsml = 0, 
  SwallowID = 0, 
  Key = 2780683700866727019, 
  {
    NumNgb = -nan(0xfffffffffffff), 
    RegionInd = -1, 
    {
      GrNr = -1, 
      origintask = 1, 
      targettask = 1
    }
  }

sbird:
Here is the Node:

(gdb) p Nodes[1108407]
$4 = {
  len = 31.281249999999996, 
  center = {19769.390625, 16828.953125, 11542.421875}, 
  father = 1108406, 
  f = {
    InternalTopLevel = 0, 
    TopLevel = 0, 
    DependsOnLocalMass = 0, 
    MaxSofteningType = 1, 
    MixedSofteningsInNode = 0
  }, 
  u = {
    suns = {1372603711, 1087590502, 919662733, 1087401534, 1718123108, 1086753389, -536870912, 1070546355}, 
    d = {
      s = {19777.598743503211, 16824.972095713532, 11540.854687747727}, 
      mass = 0.24407051503658295, 
      sibling = 123735, 
      nextnode = 1108408, 
      hmax = 0
    }
  }
}

sbird:
It is well within the Node.

sbird:
The Father node is part of a chain that ultimately points to 123734, with 123732 being the sibling. So this particle is indeed being skipped. And the final Node of the chain has a len < 1e-3 *All.ForceSoftening[1].
So this is a problem with the logic that used to be All.TreeRnd.

sbird:
I see. I do not always re-add the original child to the new node. That is easily fixed.

sbird:
These two particles are very close:
(gdb) p P[123734].Pos
$23 = {19782.566493304457, 16828.544911071953, 11535.29647550442}
(gdb) p P[123733].Pos
$24 = {19782.565989498609, 16828.527480302801, 11535.292632799374}

sbird:
What do you think about increasing the length scale for the particle disambiguation? It seems that if they get within a softening length (which happens sometimes), then they experience the same large-scale force and so tend to get closer. This triggers when they are very close, but we may as well trigger it at 1/100 ForceSoftening as 1/1000 ForceSoftening. That way we just get a less lengthy tree walk.

forcetree.c
sbird: I think the Softening flags usage is wrong in an important respect, and it is causing the problems I have been having with my neutrino simulations. I will add a comment to this branch and file a PR for the other.

forcetree.c
```diff
@@ -144,75 +156,155 @@ int force_tree_build(int npart)
...
+    int new_subnode = get_subnode(nfreep, p_toplace);
+
+    /* If the node is very small, just add the particle to the next subnode.
+     * This ensures that we can always construct the tree, and only happens far
+     * below the softening length. This happens occasionally, but if it is
+     * frequent increase the force softening.
+     */
+    if(nfreep->len < minlen && new_subnode == child_subnode) {
+        if(child_subnode < 7)
+            new_subnode = child_subnode + 1;
```
rainwoodman: I feel the bucketing shall be added before a node is split. Currently we will create a lot of empty nodes when many particles are within the ambiguity radius.

forcetree.c
```diff
@@ -221,152 +313,143 @@ int force_tree_build_single(int npart)
...
+        /*If it did, walk again*/
+        while(child >= firstnode)
+        {
 #ifdef OPENMP_USE_SPINLOCK
-            /*Initialise the SpinLock off*/
-            pthread_spin_init(&SpinLocks[ninsert-All.MaxPart], 0);
-            /*Unlock the parent*/
-            pthread_spin_unlock(&SpinLocks[this-All.MaxPart]);
+            /*Move the lock to the child*/
+            pthread_spin_lock(&SpinLocks[child-firstnode]);
```
rainwoodman: I think you may have to put this and the check above into a loop.

Is there a way to prove the child node is not split again before the lock is grabbed in the second try?

rainwoodman: If this is made into a loop it is probably not very difficult to revert to hashing of child when the node is sufficiently small.  If we do it here we avoid additional splitting for each colliding particle.

forcetree.c
sbird: If we do it before the node is split, we cannot guarantee that there is an empty subnode to receive the new particle. And if we create a new node, where do we do it? If there are enough particles within the ambiguity radius to really inflate the tree length, we have Troubles.

rainwoodman:
Another possibility is to allow the leaf nodes to contain multiple particles. Then we will simply stop refining but inserting particles to via `Next[]` when the child node is sufficiently small. I think it is better defined than the current approach.

forcetree.c
rainwoodman: We can. -- keep going till we find first empty slot but doesn't require it to be a correct child. e.g., adding a fucntion `get_any_empty_or_subnode()`

sbird:
We would have to be careful that the force_set_next_node in force_update_nodes_recursive did not over-write NextNode. It would probably be better then to set NextNode and u.d.nextnode in force_tree_create_nodes.

rainwoodman:
I think it all amounts to using linked_list_append(a, b) which walks a to the end then attach b. where a is the first child particle of a node, and b is the next node in tree walk.

rainwoodman:
I don't think we ever made the assumption that a leaf tree node contains exactly one particle outside forcetree.c.

sbird:
That sounds a very good idea. We would just have to be careful with the moments computation. If we do this then there is no force error for close particles and the tree can be made much flatter. Probably we can stop appending nodes when len < ForceSoftening, even. Basically this would make the force computation direct at very short ranges.

rainwoodman:
That's an argument for a shorter stopping length.

forcetree.c
sbird: I don't completely understand what you mean here: this is in a loop already. If the child is split we simply move the lock one more level down. The child cannot be split while we hold the lock on it, and that is the reason we lock the child before unlocking the parent: we always have a lock on something.

sbird:
Do you want to implement this or do you want me too? It is the end of the day here on the east coast, so I will not implement it tonight.

rainwoodman:
Sure. I will give it a go tonight.

forcetree.c
rainwoodman: I worry the same condition requirement the second check may require a third check. The child cannot be split while we hold the lock on it, but the child of the child (which becomes the new child) may.

sbird: But in that case we will just keep walking until we find something that we can take the lock on. We always take the lock on a node before checking the node (this part is actually identical to the earlier code: the change is the addition of a lockless fast path earlier).

sbird:
I will push the green button here as well.

PR #99, opened 2017-09-13
A more flexible memory allocator.

rainwoodman:
The intention is to allow multiple allocators, and allocation
from both directions.

This would help cleaning up code sequences that are currently
limited due to we only have a single allocator and a single direction.

The code needs to be tested, but unfortunately cmocka doesn't work
on my box (many ref not found error during linking).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,7 +27,7 @@ install:
 
 script:
     - set -e
-    - make run-tests
+    - make test
     - pushd examples/travis
     - sh run.sh
     - popd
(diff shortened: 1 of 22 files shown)


sbird:
Why do you touch every page on first initialisation, incidentally? Is it just to avoid accidental memory errors? If so, maybe hide that behind DEBUG?

You could perhaps also make the output of the peak memory printer a bit more user-friendly: probably don't need all the raw pointer values.

Apart from that, this looks great to me! A much-needed improvement, in fact.

rainwoodman:
The touch is to prefault the pages. It is to avoid memory fragmentation which stops MPI from getting memory (even there is free memory) for the network device. We keep seeing that error on BlueWaters! 

Fragmentation is less likely to happen if we back all of our virtual address with phys mem before MPI and our routines starting to punch holes in each other's address space. The mouth of words is this helps. It also ensures we don't over commit the memory and die later under stress. 

This PR is included in GarbageFlag, so I'll close this and push new commits to GarbageFlag instead.
I'll get rid of the raw pointers.

sbird:
Ah, that makes a lot of sense, thanks! Feel free to merge this to statictree right away if you prefer: I successfully ran a lya simulation to z=2 with it.

PR #101, opened 2017-09-13
Alter testing

rainwoodman:
Take a look.

I think this is easier to use than the current makefile. 
This works better with tab-completion.


rainwoodman:
I added cmocka as a built-in dependency, which is only build if tests are requested.

The version on travis is too old and it conflicts with our usage. 

I also significantly shrinked the duration of the example travis run. We shall gradually move to rely on the unit tests instead of this final integration test.

rainwoodman:
This will likely make statictree green, and forcetree-test will need a rebase (for some filename changes and c99 syntax fixes.)

Makefile.tests
```diff
@@ -0,0 +1,43 @@
...
+
+.PHONY: run-tests build-tests
+
+
+depends/lib/libcmocka.a:
+	cd depends; make lib/libcmocka.a
+
+build-tests: $(TESTBIN)
+
+run-tests: build-tests
```
sbird: Can you make this just "make test" as before? That is a standard make target name.

Makefile.tests
```diff
@@ -0,0 +1,43 @@
...
+	for tt in $(SUITE) ; do $$tt ; done
+
+clean-tests:
+	rm -f $(TESTBIN)
+
+$(TESTBIN):
+
+tests/test_%: tests/test_%.c $(DESTDIR)/%.o $(DESTDIR)/endrun.o depends/lib/libcmocka.a
```
sbird: If you link to endrun.o you need to call MPI_Init in main and MPI_Finalize in teardown for all the test modules.

rainwoodman: Yes. we shall certainly do that -- they have to deal with MPI at some point.

rainwoodman:
What are the gains using the installed version of cmocka?

Makefile.tests
rainwoodman: We use `run-tests.py` in many python packages -- that's where the name is from. I guess make is already a verb then it doesn't make sense to have two verbs.

sbird:
You forgot to git add tests/stub.c

sbird:
I guess the main practical advantage is that we don't have to compile cmocka ourselves - and thus we don't need cmake.

rainwoodman:
In nbodykit, dealing with testing on a variety of number of ranks we use MPI_Comm_split and pass in the communicator to the testee, calling it with each different requested MPI size.

In MP_Gadget, the issue is our testee assumes they work on MPI_COMM_WORLD. It is a pretty poor design decision inherited from Gadget. but it will be too much work to fix that.

rainwoodman:
If you have the luxury to install a binary of cmocka, then probably have the same luxury to install cmake?

The advantage of bundling a cmocka is that we require a somewhat recent cmocka (>1.0?). On travis the default cmocka is 0.3.x and we fail compiling the tests.

I made it such that if the tests are not requested cmocka is not built. Therefore it doesn't get into the way of a quick-and-dirty user, assuming tests are only meaningful for developers.

sbird:
Sure sure - but if the issue is travis, why not just build it in the travis startup script? But it's not a big deal.

rainwoodman:
Similar issue may happen on other boxes. I assume the OS on travis is 'typical'. 
Building cmocka doesn't really take much time so it is not a big deal I agree.

sbird:
Ok, sure. It is not too bad. I am happy with this, merge when you're ready.

PR #102, opened 2017-09-13
WIP: Garbageflag

rainwoodman:
This PR adds an IsGarbage flag to particles and use  gc in exchange.

A new memory allocator is introduced.

### DIFF
--- a/blackhole.c
+++ b/blackhole.c
@@ -547,8 +547,12 @@ blackhole_feedback_ngbiter(TreeWalkQueryBHFeedback * I,
         if(P[other].SwallowID != I->ID) return;
 
         lock_particle(other);
+
+        /* Mass conservation */
         O->Mass += (P[other].Mass);
         O->BH_Mass += (BHP(other).Mass);
+        P[other].Mass = 0;
+        BHP(other).Mass = 0;
 
         int d;
         for(d = 0; d < 3; d++)
@@ -556,8 +560,7 @@ blackhole_feedback_ngbiter(TreeWalkQueryBHFeedback * I,
 
         O->BH_CountProgs += BHP(other).CountProgs;
 
-        P[other].Mass = 0;
-        BHP(other).Mass = 0;
+        P[other].IsGarbage = 1;
         BHP(other).Mdot = 0;
 
 #pragma omp atomic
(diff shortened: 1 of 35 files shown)


blackhole.c
```diff
@@ -547,17 +549,20 @@ blackhole_feedback_ngbiter(TreeWalkQueryBHFeedback * I,
         if(P[other].SwallowID != I->ID) return;
 
         lock_particle(other);
+
+        /* Mass conservation */
         O->Mass += (P[other].Mass);
         O->BH_Mass += (BHP(other).Mass);
+        P[other].Mass = 0;
+        BHP(other).Mass = 0;
```
sbird: This changes the logic, because the Mass is zerod *before* it is accreted.

blackhole.c
```diff
@@ -601,13 +606,15 @@ blackhole_feedback_ngbiter(TreeWalkQueryBHFeedback * I,
 
         lock_particle(other);
 
+        /* Mass Conservation */
         O->Mass += (P[other].Mass);
+        P[other].Mass = 0;
```
sbird: Ditto. Why zero the mass at all when you have IsGarbage?

rainwoodman: See the comment above. It signifies mass must be conserved. I keep this otherwise useless code there to remind you something potentially bad is going on -- We do not have a mechanism to notify the tree particles have turned into garbage or the mass of a particle has changed. And these operations, even emitted, shall be somehow translated into a noop, because we know the mass of the node is conserved. I haven't thought clear how to do this.

blackhole.c
rainwoodman: good catch!

blackhole.c
sbird: Ah, I see. That makes a lot of sense, but still you need this to be set later, because otherwise AccretedMomentum would be zero.

rainwoodman: Yes I fixed that and added a rebase. Actually there is no guarentee the swallowee and the swallower are in the same tree node. So it won't always become a noop. I guess we need a function to 'transfer mass'.

rainwoodman: I can add that in a new PR. It could be gradually paving the road towards incremental tree.

sbird: A very good point

PR #103, opened 2017-09-14
Use two versions of density(), and density_update()

rainwoodman:
Replaces #98 

This one doesn't slow down at all because it doesn't introduce an additional tree walk every step.

It only helps to stablize the initialization iteration in pressure-entropy SPH. No other structural benefits from split smoothing from density. 

@sbird Does this look ok to you?


sbird:
Looks good to me

PR #104, opened 2017-09-14
Fix Node softenings

sbird:
The current logic for force softenings in nodes is roughly this:

1) Check particle softening and maximum node softening
(If we have adaptive softening for gas particles, this affects what the softenings are)

2) Check whether the particle softening is less than the node softening. 
If they are the same, use the node. This in particular means that nodes 
are never opened if all softening lengths are identical.
3) Check whether MixedSofteningsInNode is also true. If it is, open the node. 
If it is not, use the node as is.

I believe the correct logic is wildly wrong, and will lead to unphysical particle 
coupling (which I have observed many times with the neutrino simulations).
If we are within the softening length of the node, we should always open the node.
The original author seems to have thought that if opening the node would not change 
the softening length, there is no point opening it. However, this is not true:
consider for example a particle close to a node containing two particles, 
equidistant between them, as so:

x  N  x

   P

If P is exactly 1 softening length from N, it is more than 1 softening length from both x's,
and the force it experiences may be quite different.

This branch implements logic where we always open a node if we find ourselves within 
the softening length. 

This should fix a real problem I was having with the neutrino simulations: 
I would observe large unphysical particle coupling, which disappeared when I made the neutrino
softening slightly larger. Even slightly larger was enough the nodes would almost 
always get opened for CDM particles, because nodes almost always contain both types.
Once the nodes are opened, the force on each particle differs more and the particle 
coupling is reduced.

This pull request removes checks 2 and 3, just opening the node if we are close to it. MixedSofteningsInNode is removed entirely.

### DIFF
--- a/gravshort-tree.c
+++ b/gravshort-tree.c
@@ -141,9 +141,9 @@ int force_treeev_shortrange(TreeWalkQueryGravShort * input,
         {
             double mass, facpot, fac, r2, r, h;
             double dx, dy, dz;
-            int otherh;
             if(no < All.MaxPart)
             {
+                int otherh;
                 /* the index of the node is the index of the particle */
                 drift_particle(no, All.Ti_Current);
 
@@ -170,6 +170,7 @@ int force_treeev_shortrange(TreeWalkQueryGravShort * input,
             }
             else			/* we have an  internal node */
             {
+                int nodeh, ph;
                 struct NODE *nop;
                 if(no >= All.MaxPart + MaxNodes)	/* pseudo particle */
                 {
@@ -248,19 +249,30 @@ int force_treeev_shortrange(TreeWalkQueryGravShort * input,
                     }
                 }
 
-                h = input->Soft;
-                otherh = nop->u.d.MaxSoftening;
-                if(h < otherh)
-                {
-                    h = otherh;
-                    if(r2 < h * h)
-                    {
-                        if(nop->f.MixedSofteningsInNode)
-                        {
+                ph = input->Soft;
+                nodeh = nop->u.d.MaxSoftening;
+                /* We want to open the node when it is possible
+                 * that some but not all of it is within
+                 * the softening length; opening is the only way
+                 * we can tell how much.*/
+                /* This only happens if at least some of the
+                 * node is within the softening length.*/
+                h = DMAX(ph, nodeh);
+                if(r2 < (h + nop->len/2)*(h + nop->len/2)) {
+                    /* a) the node softening dominates, but the node
+                     * particles do not all have the same softening value,
+                     */
+                    if(ph < nodeh) {
+                        if(nop->f.MixedSofteningsInNode) {
                             no = nop->u.d.nextnode;
-
                             continue;
                         }
+                    }
+                    /* b) No need to open if all the node is within
+                     * the softening length: all is softened.*/
+                    if(r2 > (h - nop->len/2)*(h - nop->len/2)) {
+                        no = nop->u.d.nextnode;
+                        continue;
                     }
                 }
                 no = nop->u.d.sibling;	/* ok, node can be used */
--- a/forcetree.h
+++ b/forcetree.h
@@ -18,6 +18,9 @@ extern struct NODE
         unsigned int InternalTopLevel :1; /* TopLevel and has a child which is also TopLevel*/
         unsigned int TopLevel :1; /* Node corresponding to a toplevel node */
         unsigned int DependsOnLocalMass :1;  /* Intersects with local mass */
+        /* Stores if there are types with different softenings in the node.
+         * This is used to open the node if it matters, because
+         * there is no other way to tell where the types in the node are.*/
         unsigned int MixedSofteningsInNode:1;  /* Softening is mixed, need to open the node */
     } f;
     union


rainwoodman:
It sounds like the node softening shall be defined as the minimum softening instead of the maximum. Is this an accurate statement of your concern?

sbird:
I don't think so. I am worried that using a tree node for force is inherently inaccurate when a particle is within the softening length of the tree. I think this is a rare occurrence, because if we are inside the node we will always open it anyway. So we need nodes that are smaller than the ForceSoftening. 

But I cannot see how it makes sense to check the things that are checked, and I think it could lead to particles becoming coupled unphysically in some cases.

rainwoodman:
Let's push your case a bit further. P has a huge sml, and next to it is a node with a very small sml. In this case I don't think you want to open the node -- because the force is smoothed by the max smoothing of the two. 

Perhaps there shall be a geometry factor before r2 < h2?

sbird:
Yes, you're right - maybe what we really want is just:

- if(h< otherh)
+ if(h <= otherh)

so it is opened also if the softenings are equal?

I think we do certainly always want to open the node if the node softening is larger than the particle softening, because the large softening may be coming from a particle on the far side of the node, in which case opening may change the force law.

Maybe we need a geometry factor before r2 < h2 iff h >= otherh?

rainwoodman:
Yes. I agree.

sbird:
I will try to do some maths tomorrow to calculate it.

sbird:
This code dates back to Gadget-2 and isn't discussed in the G2 paper - I verified that it has the right behaviour at r==0, but at r ~ h I am unsure. I am having trouble computing the right geometry factor as the integrals are a little complicated. Also the other opening criteria are not derived as far as I can see, just picked based on what gives good accuracy.

I wonder if, given forcetree-ll, whether the right thing to do here is to set:
```
minlen ~ ForceSoftening
```
and let the existing condition that opens when the particle is within the tree node take care of this?
This would make the tree walk much shallower, which would likely compensate for the increased number of openings. Something similar seems to be done in the HACC (section 2.4 of https://arxiv.org/abs/1410.2805). 

What do you think?

sbird:
Alternatively, we could do:
```
if(r2 < (h+len)^2 
    open;
```
which ensures that all particles in the node are within a softening length. If this is the case then I think the force is accurate.

sbird:
Push implementations of both these ideas for your perusal.

rainwoodman:
- We don't have a fast PP kernel, so I worry their benefits won't really show up. What about leaving the tree termination condition as a parameter, and default it to 1e-3 for now?

- Would it help if we kept a min smoothing as well? But if it is gets too complicated let's just open them all and see how slow it becomes..

gravshort-tree.c
```diff
@@ -257,16 +257,19 @@ int force_treeev_shortrange(TreeWalkQueryGravShort * input,
...
-                        }
+                /* We want to open the node if it is partially but not entirely
+                 * within the softening length of the particle.
+                 * Note this differs from vanilla Gadget, which
+                 * only opens nodes if they have a larger
+                 * softening than the particle.
+                 * This should reduce force coupling.*/
+                if(r2 < h * h)
+                {
+                    if(r2 > (h + nop->len/2)*(h + nop->len/2)) {
```
rainwoodman: (h + nop->len/2)**2 > h2 so this will never be hit.

sbird: Ah indeed, I screwed up my conditionals. Duh.

sbird:
That's should be a bit more sensible.

sbird:
In general I suspect the speed impact of the direct force will be pretty small: particles significantly closer than a softening length seem to always be in pairs, and when N==2, N^2 is not too much...but at the moment we can have log_2(1e-3) = 10 nodes in a row.

But I'm not sure. Let me run tests without the minimum and see if that helps.

rainwoodman:
These assumptions are no longer true when there is hydro. hydro particles can get a lot closer than the softening.

sbird:
I have thought about it more, and I think we want something more like this version. It is much more like the G2 code, and focusses on deciding whether we can prove that all or none of the node is within the softening radius: if this is so, we do not need to open. If some but not all of it is within the softening radius, we do need to open.

gravshort-tree.c
```diff
@@ -248,26 +249,37 @@ int force_treeev_shortrange(TreeWalkQueryGravShort * input,
...
+                    if(ph < nodeh) {
+                        if(nop->f.MixedSofteningsInNode) {
                             no = nop->u.d.nextnode;
-
                             continue;
                         }
                     }
+                    /* b) No need to open if all the node is within
+                     * the softening length: all is softened.*/
+                    if(r2 > (h - nop->len/2)*(h - nop->len/2)) {
```
rainwoodman: What happens if h < hop->len/2?

rainwoodman:
Is it simpler to state that there is no need to open if min(ph, node_min_smoothing) > r**2?

sbird:
Maybe. Let me do some tests with several variants of these conditions.

gravshort-tree.c
sbird: In practice this becomes a less stringent version of the "check whether we are within the node" but that is kind of accidental and needs to be cleaned up.

sbird:
I had the idea of using a realistic test to find how large the force error actually is in this case, and how one can minimise it. However, I was unable to find any actual real-simulation test where the condition was satisfied (I put a call to endrun() in the conditional and ran to z=0 with full hydro). So I conclude that this really doesn't matter and am closing the pull request.

rainwoodman:
good. So we are one more step closer to 5.0.0.

PR #110, opened 2017-09-18
Integer Timeline doesn't change over an unscheduled save/resume

sbird:
This is something I've been meaning to do for a while, but the arbitrary syncpoints make it easy.

### DIFF
--- a/run.c
+++ b/run.c
@@ -39,7 +39,7 @@ enum ActionType {
 };
 static enum ActionType human_interaction(double lastPM, double TimeLastOutput);
 static int should_we_timeout(double TimelastPM);
-static void compute_accelerations(int is_PM);
+static void compute_accelerations(int is_PM, int FirstStep);
 static void update_IO_params(const char * ioctlfname);
 static void every_timestep_stuff(int NumForces, int NumCurrentTiStep);
 static void write_cpu_log(int NumCurrentTiStep);
@@ -153,7 +153,7 @@ void run(void)
         every_timestep_stuff(NumForces, NumCurrentTiStep);	/* write some info to log-files */
 
         /* update force to Ti_Current */
-        compute_accelerations(is_PM);
+        compute_accelerations(is_PM, NumCurrentTiStep == 0);
 
         /* Update velocity to Ti_Current; this synchonizes TiKick and TiDrift for the active particles */
 
@@ -309,7 +309,7 @@ human_interaction(double TimeLastPM, double TimeLastOut)
  * be outside the allowed bounds, it will be readjusted by the function ensure_neighbours(), and for those
  * particle, the densities are recomputed accordingly. Finally, the hydrodynamical forces are added.
  */
-void compute_accelerations(int is_PM)
+void compute_accelerations(int is_PM, int FirstStep)
 {
     message(0, "Start force computation...\n");
 
@@ -324,14 +324,14 @@ void compute_accelerations(int is_PM)
             energy_statistics();
     }
 
-    grav_short_tree();		/* computes gravity accel. */
+    /* For the first timestep, we do tree force twice
+     * to allow usage of relative opening
+     * criterion for consistent accuracy.
+     */
+    if(FirstStep)
+        grav_short_tree();
 
-    if(All.Ti_Current == ti_from_loga(All.TimeInit))
-        grav_short_tree();		/* For the first timestep, we redo it
-                             * to allow usage of relative opening
-                             * criterion for consistent accuracy.
-                             */
-
+    grav_short_tree();
 
     if(NTotal[0] > 0)
     {
--- a/run.c
+++ b/run.c
@@ -326,7 +326,7 @@ void compute_accelerations(int is_PM)
 
     grav_short_tree();		/* computes gravity accel. */
 
-    if(All.Ti_Current == 0)
+    if(All.Ti_Current == ti_from_loga(All.TimeInit))
         grav_short_tree();		/* For the first timestep, we redo it
                              * to allow usage of relative opening
                              * criterion for consistent accuracy.
--- a/timebinmgr.c
+++ b/timebinmgr.c
@@ -25,7 +25,7 @@ setup_sync_points(void)
 
     /* Set up first and last entry to SyncPoints; TODO we can insert many more! */
 
-    SyncPoints[0].loga = log(All.TimeBegin);
+    SyncPoints[0].loga = log(All.TimeIC);
     SyncPoints[0].write_snapshot = 0; /* by default no output here. */
     SyncPoints[1].loga = log(All.TimeMax);
     SyncPoints[1].write_snapshot = 1;
--- a/timebinmgr.c
+++ b/timebinmgr.c
@@ -25,7 +25,7 @@ setup_sync_points(void)
 
     /* Set up first and last entry to SyncPoints; TODO we can insert many more! */
 
-    SyncPoints[0].loga = log(All.TimeInit); 
+    SyncPoints[0].loga = log(All.TimeBegin);
     SyncPoints[0].write_snapshot = 0; /* by default no output here. */
     SyncPoints[1].loga = log(All.TimeMax);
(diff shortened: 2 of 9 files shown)


run.c
```diff
@@ -324,14 +324,14 @@ void compute_accelerations(int is_PM)
...
-    if(All.Ti_Current == 0)
-        grav_short_tree();		/* For the first timestep, we redo it
-                             * to allow usage of relative opening
-                             * criterion for consistent accuracy.
-                             */
+    /* For the first timestep, we do tree force twice
+     * to allow usage of relative opening
+     * criterion for consistent accuracy.
+     */
+    if(FirstStep)
```
rainwoodman: We can probably move the first short_tree calculation to init?

timebinmgr.c
```diff
@@ -29,6 +29,16 @@ setup_sync_points(void)
     SyncPoints[0].write_snapshot = 0; /* by default no output here. */
     SyncPoints[1].loga = log(All.TimeMax);
     SyncPoints[1].write_snapshot = 1;
+    /* Set the first SyncPoint to the last scheduled output.
+     * This ensures that the timeline does not change after
+     * save/load from an unscheduled output.*/
```
rainwoodman: The insertion sort should have taken care of this already..

rainwoodman: See the loop a few lines below.

rainwoodman:
It may be easier if you simply save TimeInit and Ti_Current in the snapshot header than trying to reconstruct the scene -- less assumptions.

rainwoodman:
I mean saving TimeBegin and Ti_Current TimeInit is still the last snapshot time.

run.c
sbird: Alas, no - the OldAcc calculation uses GravPM, so that needs to be computed before this.

timebinmgr.c
sbird: No, the insertion sort puts something like:
TimeInit, Out[2], Out[3]...
whereas this makes;
Out[1], Out[2], Out[3].
The point is to make the first SyncPoint the last *scheduled* output, which may not be the same as TimeInit.

sbird:
I don't want to save Ti_Current because then the internal integer timeline numbers become contractual - we cannot change them without breaking snapshot compatibility. I could save TimeBegin to the header though.

rainwoodman:
The conversion between integer and float is governed by planned sync-points, which depends on TimeBegin, TimeMax, and OutputList.

If any of them changes Ti no longer make much sense. This makes me wonder why do you want to preserve Ti across restarts?

timebinmgr.c
rainwoodman: If TimeMin is used instead of TimeInit, then this loop is not needed. The code below will reconstruct the same set of planned sync points (which goverens ti and loga conversion) if OutputList and TimeMax are preserved as well.

run.c
rainwoodman: Perhaps setting OldAcc to the snapshot numbers, then this can be avoided?

rainwoodman: Looks like it is not worth it to go that far to avoid this flag. We have to do something special here anyways. But oldacc can be lifted out from P -- it can be computed before the tree walk in a local array from GravPM and GravAccel.

sbird:
The goal of this pull request is for an unplanned sync point - with no other change to parameter files - to not change the timesteps. This ensures that if we write a timeout (or auto-checkpoint) snapshot and resume from it later, we still get the same timesteps as we would have done had we asked for a longer allocation in the first place. 

To do this, we require that the DlogA values remain the same as they would before. DlogA is set by the gap between two planned outputs, so we aim with this pr to keep this the same across unplanned outputs.

Note Ti_Current is not in fact preserved across restarts: but the float Time values of every timestep are.

run.c
sbird: You mean saving and reloading OldAcc with a snapshot save? I did implement that, but once I had done so it seemed not really worth the effort (although it makes startup a bit faster).

I will take a look at lifting OldAcc in another pr - it doesn't really fit here.

timebinmgr.c
sbird: Yes, this is true. That is likely a simpler scheme, and it works also if the unplanned sync is before the first planned sync.

sbird:
This version saves TimeBegin to the snapshot.

rainwoodman:
I thought about this last night -- what about this?

- Parameter file : TimeMin, default = 0.001 (z=999)
- Parameter file : TimeMax, default = 1.0
- Parameter file : MaxTimeStep, default = 0.1 (or some other numbers), in dloga

- `setup_sync_points()` sets up a uniform grid of sync points arange(log10(TimeBin), log10(TimeMax), MaxTimeStep).

- Insert OutputTimeList to sync points, with `write_snapshot` and `write_fof`.

sbird:
What happens when OutputTimeList is close to but not identical to one of the entries in the uniform grid of sync points? You could end up with some very small timesteps. eg:
log(OutputTimeList) = 0.4, 0.51 0.61, 0.71
MaxTimeStep = 0.1
weird things will happen if we saturate the timestep limiter anywhere.

sbird:
I also think I would like to store TimeBegin in the snapshot anyway, because it gives us a handy way to know the redshift of any CAMB power spectrum we might need for cosmology (eg, the soon-to-be-a-pull-request neutrino model uses one). Currently it is a parameter file parameter, but storing it in the snapshot would make more sense.

rainwoodman:
I don't have objections saving the TimeBegin in the header, but perhaps a fitting name is TimeLPT, or TimeIC, as it is the Time of the LPT Initial condition. 

TimeMin, TimeMax is the starting and ending of the integer to dloga mapping. They are two different concepts.

As for the nearby points, we can create the syncpoints that's close to the step size limit around the requested outputs. (a well defined operation, e.g. https://github.com/rainwoodman/fastpm-python/blob/master/fastpm/core.py#L213). Or we can drift the output times to the nearest sync points (another well defined operation) which seems to be easier.

sbird:
What does separating the integer mapping from the outputs in this way gain us though?

Remember that, unlike in fastpm, in practice the long-range timestep is always set by the displacement constraint. So for every sync point that we insert we may end up doing extra timesteps to meet it. This can be severe because it is a full PM timestep.

We cannot drift the output times to the nearest sync points - the reason to rewrite the integer timeline was that we needed output times where the particles are fully synced and have a time which is exactly what the user asked for.

rainwoodman:
We already do insert additional PM steps to meet the MaxTimeStep and Outputlists, so planning ahead before the run won't introduce more additional PM steps.

I have second thoughts about the displacement constraint. With displacment constraint, you do get 'reproducibility' but it is not robust against perturbations of parameters. The number of PM steps becomes per-realization, introducing a different numerical artifact when the parameters are perturbed -- I suspect the total number of PM steps will differ even if the random phases are flipped.

sbird:
That is true, but even more true with the tree timesteps - I don't think we can make the timeline robust to changing the ICs. Which is why I think it makes sense to anchor it at TimeIC.

The displacement constraint is indeed a bit hokey - but as far as I know no-one else has changed it. If we do that we should do it properly and write a stand-alone paper demonstrating it is better.

rainwoodman:
OK' then. We can keep removing the displacement constraint as an option when problems emerge -- actually, the number of short range steps can also be fixed. See chapter 4 of https://arxiv.org/abs/1510.06665 .

sbird:
For that though they are explicitly losing accuracy on small (galactic) scales, which we would need to not do (also I do not really understand the HACC papers). But I do think such a paper would be interesting and would be happy to work on it. The obvious other way to set the PM step is to do something similar to the short-range timestep: compute   dt = sqrt(GridCellSize/|GravPM|). But that is so obvious there must be a reason they didn't do it for Gadget-2. Do you know why not? Probably in some Hernquist paper from the 90s.

sbird:
Shall I merge this?

rainwoodman:
Yes. Looks clean without the loop!

sbird:
Cheers!

PR #112, opened 2017-09-19
Wrong syncpoints

rainwoodman:
This PR may fix the issue  from landerson, #114  where the code crashes outputting from high redshifts. It is because of a silly typo of memmove, mixing source and dest, and as a result some syncpoints / output times are lost!


rainwoodman:
I am merging this because this is a shameful type of bug.

sbird:
Hah, I did that myself yesterday.

PR #117, opened 2017-09-20
Nbkittest

rainwoodman:


sbird:
Nice! Why not check the power of the snapshots output by MP-Gadget as well? That's a quite powerful consistency check.

rainwoodman:
I wonder if there is a way we can add some assertion of the stability of the result. Probably worth working it out using #118 as a test of the test?

sbird:
What do you mean by "stability of the result"? Do you mean that the output should remain the same? I guess we would need then to add some "known-good" data somewhere (or we could just check against CAMB with sigma_8 = 0.05, so everything is linear).

rainwoodman:
I added assertions on the growth -- that was OK from (a=0.1 to a=0.2ish), so it is pretty much the same as setting sigma8 = 0.05. (but with very few time steps)

The problem occurs asserting the power of IC/PART_000.

I find with nbodykit 0.2.6 the measured IC power disagrees with the truth IC at ~7% (systematically lower), but with nbodykit 0.2.7 the difference is within 5%. then there was some slight tweaking of handling of cosmology parameters in nbodykit between the two versions.

The 5% error is still a bit strange. The cosmology is slightly different -- nbkit uses class which contains radiation in the bg, but the travis test currently doesn't. (maybe we shall at it?).

ZA also introduces a difference from the linear spectrum. (you mentioned the mesh res is low enough for this to be an effect)

It may also is due to the sigma8 normalization, which may have shifted the entire spectrum before using it.

sbird:
I will take a look next week. It may indeed be due to the scaling

rainwoodman:
Perhaps we can merge this one first since it is already green?

sbird:
Do it!

rainwoodman:
Merged. Let's rebase the dangling PRs and hope they workout fine..

PR #118, opened 2017-09-25
Remove OldAcc and PM_Potential from struct particle_data

sbird:
Following a suggestion by Yu Feng, we do not actually need OldAcc in the particle data. 
I was originally going to create a temporary array, but I realised that because it is only used by one half of the force equation, we don't need it at all. 

To ensure this gives the same result as the previous solver,  we must perform the tree gravity first and the PM gravity second: otherwise the current PM step over-writes the last one.

I was then able to combine PM_Potential with Potential. 
I could have combined GravPM and GravAccel, but they are too useful for debugging.

### DIFF
--- a/run.c
+++ b/run.c
@@ -315,11 +315,26 @@ void compute_accelerations(int is_PM, int FirstStep)
 
     walltime_measure("/Misc");
 
+    /* The opening criterion for the gravtree
+     * uses the *total* gravitational acceleration
+     * from the last timestep, GravPM+GravAccel.
+     * So we must compute GravAccel for this timestep
+     * before gravpm_force() writes the PM acc. for
+     * this timestep to GravPM. Note initially both
+     * are zero and so the tree is opened maximally
+     * on the first timestep.*/
     grav_short_tree();
 
-    /* PM is after tree as we need GravPM
-     * from the last step to compute the opening
-     * angle in tree.*/
+    /* We use the total gravitational acc.
+     * to open the tree and total acc for the timestep.
+     * Note that any of (GravAccel, GravPM,
+     * HydroAccel) may change much faster than
+     * the total acc.
+     * We do the same as Gadget-2, but one could
+     * instead use short-range tree acc. only
+     * for opening angle or short-range timesteps,
+     * or include hydro in the opening angle.*/
+
     if(is_PM)
     {
         gravpm_force();
@@ -332,6 +347,8 @@ void compute_accelerations(int is_PM, int FirstStep)
     /* For the first timestep, we do tree force twice
      * to allow usage of relative opening
      * criterion for consistent accuracy.
+     * This happens after PM because we want to
+     * use the total acceleration for tree opening.
      */
     if(FirstStep)
         grav_short_tree();
--- a/run.c
+++ b/run.c
@@ -315,6 +315,11 @@ void compute_accelerations(int is_PM, int FirstStep)
 
     walltime_measure("/Misc");
 
+    grav_short_tree();
+
+    /* PM is after tree as we need GravPM
+     * from the last step to compute the opening
+     * angle in tree.*/
     if(is_PM)
     {
         gravpm_force();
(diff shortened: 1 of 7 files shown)


run.c
```diff
@@ -315,6 +315,11 @@ void compute_accelerations(int is_PM, int FirstStep)
 
     walltime_measure("/Misc");
 
+    grav_short_tree();
+
+    /* PM is after tree as we need GravPM
+     * from the last step to compute the opening
+     * angle in tree.*/
```
rainwoodman: This breaks the formal agreement between the KkddkK and the code -- making it kKddkK.

Since the opening condition introduces a higher order error, we probably can either use the old PM or neglect PM force alltogether in the condition -- can we save both to the snapshot and look at the statistics?

sbird: We currently use old PM: the PM over-writes GravPM, so to use the old PM in tree we need the PM step to come after tree. Note that no drifting or kicking is done in the PM or tree code anymore: we are just computing the accelerations and potential. So the formal properties are unchanged, but obscured.

I thought about not using PM, or using new PM, but while the timestep is set so that the acceleration does not change too much, it does not make the same guarantee about GravAccel and GravPM: consider two particles separated by ~Asmth: the acceleration will be roughly the same between timesteps but it may shift quickly from GravAccel to GravPM, meaning that using one of them from a different timestep may introduce a large error.

If you are unhappy with the ordering changing, we should just close this pull request and forego the optimization.

rainwoodman: Does the change of ordering actually work? PM frees the tree (reusing the memory for mesh), but both grav and hydro needs the tree.  So after the ordering change we are building the tree twice?

On your example:
You forgot the long/short range smoothing kernel and Acut. If two particles separated by ~Asmth, then the force will be dominated by tree because Asmth << Acut. If two particles are separated by ~Acut, then the force will be dominated by PM because the short range at Acut is suppressed by erf.

rainwoodman: You are right there is a sweet spot where the force is half PM and half Tree, but it shouldn't suddenly jump between the two because of the smoothing. 

I tend to view short range and long range gravity as two separate forces (because the code says so), thus the accuracy of short range shall only depend on short range force. But the argument can also be made that we want to control the accuracy of the total gravity, because the physics says so. However the physics also says the total force includes hydro and effective forces from the feedback, and we don't take care of those here neither.

sbird: I think we rebuild the tree twice on a PM step anyway, don't we? We rebuild it at the end of the domain decomposition, use it to do PM assignment to regions, free it, and then rebuild it again.
Moving the PM later would actually allow us to only rebuild it once: we could in principle do hydro and feedback, then do PM *at the end* and avoid rebuilding it after we free. But this is probably too much of a contortion for a small gain: the tree rebuild is a quite small fraction of a PM step.

sbird: The code is allowed to make the timestep long enough that the particle can move across the Acut/Asmth boundary, especially if the particle is moving fast and has small acceleration. It is because the timestep is set by the total force.

I agree that arguably either the short timestep should just be GravAccel, or both timestep and truncation should include hydro. Actually that seems a good short paper for a beginning student...

rainwoodman: You are right. But moving all shortrange before PM looks cleaner than having some of them before and some after. Could you add a few words in the comment to say the purpose of using old PM acc is to maintain agreement with Gadget2?

sbird: Absolutely.

rainwoodman: Let's make sure the options are documented in the code such that the next person reading it knows why it is this way but not the other way.

sbird:
How is that?

sbird:
Shall I merge this?

rainwoodman:
Sure!

rainwoodman:
I wonder if we shall / can invite people to help code reviews even if they don't necessarily directly write code. I guess it is hard to find such people ...

sbird:
Probably. More people writing code would also be good...

rainwoodman:
The key issue is probably finding the right person who can code and cares. Reviewing might be a lower barrier way to get people started -- e.g. would it reasonable to ask fresh students to reviewing code changes?

sbird:
I don't know how good they would be at it...maybe we could also find them a simple feature to implement? Or get them to work on one of the open issues?

rainwoodman:
Sounds like perfect HW projects for a course like 'software engineering for computational physics'

sbird:
hehe, yes.

PR #124, opened 2017-10-04
Neutrinos

sbird:
Merge a massive neutrino implementation (since I am here).

This is based on the kspace-neutrinos implementation of Ali-Haimoud & Bird 2012, which follows neutrinos analytically, but it also incorporates hybrid neutrinos, which are initially tracers and then later gravitate. It uses code from the kspace-neutrinos repo (which has its own tests), but does not use the generic interface layer from that repo. This should have little to no effect if neutrinos are massless.


sbird:
Note that does not include neutrino support in MP-GenIC - that piece requires different transfer functions, which I didn't implement yet.

rainwoodman:
But that means we can't test it .. Shall we add something real simple to genic (could be wrong) just to drive the tests -- e.g. having the initial neutrinos on a uniform grid (with no power, since you said it doesn't matter?)

I worry if we don't merge it becomes too divergent; but it is also not good if we merge it without being able to drive it

sbird:
I can add a basic implementation of multiple transfer functions to MP-GenIC - that let's us test the important bit, the semi-linear neutrinos, and it is useful for other reasons (testing ADAPTIVE_GRAVSOFT_FORGAS, for example). It is not too hard. I'll see what I can do and let's hold off the merge until then. Do you like the general approach?

rainwoodman:
Yes. your approach address my biggest concern!

andreufont:
Hi @sbird - As you know, I'm very interested in this pull request. Please, let me know when it is ready for me to try it :-)

rainwoodman:
This is a rather large PR. I need to digest it -- one possibile approach is I do the rebase over the weekend to force myself to understand it better. But we do still want a driver otherwise there is no way for people to even run it..

sbird:
Do you still want to do the rebase? It needs to be done so I can add the GenIC integration using the different transfer function support.

rainwoodman:
I won't have the time for it -- I'll work on the CI infrastructure on nersc today. So go ahead.

sbird:
That is a basic neutrino driver for GenIC - no support for the hybrid neutrinos yet. 
@andreufont , feel free to give this a try.

sbird:
Drivers for hybrid neutrinos are now implemented. This should work and is ready to be reviewed.

allvars.h
```diff
@@ -288,6 +288,14 @@ extern struct global_data_all_processes
     int BlackHoleOn;  /* if black holes are enabled */
     int StarformationOn;  /* if star formation is enabled */
     int WindOn; /* if Wind is enabled */
+    int MassiveNuLinRespOn; /*!< flags that massive neutrinos using the linear
+                                 response code of Ali-Haimoud & Bird 2013.*/
+    int HybridNeutrinosOn; /*!< Flags that hybrid neutrinos are enabled */
+    double HybridVcrit; /*!< Critical velocity switching between particle
+                          and analytic solvers when hybrid neutrinos are on*/
+    double HybridNuPartTime; /*!< Redshift at which hybrid neutrinos switch on*/
+    char CAMBTransferFunction[100]; /*!< CAMB transfer function for initial neutrino power*/
```
rainwoodman: 100 might be too short.

begrun.c
```diff
@@ -213,13 +217,16 @@ set_units(void)
     message(0, "UnitDensity_in_cgs = %g \n", All.UnitDensity_in_cgs);
     message(0, "UnitEnergy_in_cgs = %g \n", All.UnitEnergy_in_cgs);
     message(0, "Photon density OmegaG = %g\n",All.CP.OmegaG);
-    message(0, "Massless Neutrino density OmegaNu0 = %g\n",All.CP.OmegaNu0);
+    if(!All.MassiveNuLinRespOn)
```
rainwoodman: Perhaps we can start using '.' to create long parameter names. What is the relation between MassiveNuLinResp and HybridNeutrinos?

begrun.c
```diff
@@ -213,13 +217,16 @@ set_units(void)
...
+        message(0, "Massless Neutrino density OmegaNu0 = %g\n",get_omega_nu(&All.CP.ONu, 1));
```
rainwoodman: In class they call this omega_ur, ultra relativistic species. A reason it is cleaner is because it is hard to distinguish nu,  massive nu and masslessnu. What convention are we using here? nu is massless and mnu is massive?

begrun.c
```diff
@@ -213,13 +217,16 @@ set_units(void)
...
     message(0, "Curvature density OmegaK = %g\n",All.CP.OmegaK);
     if(All.CP.RadiationOn) {
         /* note that this value is inaccurate if there is massive neutrino. */
+        double OmegaTot = All.CP.OmegaG + All.CP.OmegaK + All.CP.Omega0 + All.CP.OmegaLambda;
+        if(All.CP.MNu[0] + All.CP.MNu[1] + All.CP.MNu[2] == 0)
+            OmegaTot += get_omega_nu(&All.CP.ONu, 1);
```
rainwoodman: ONu and MNu are not independent parameters -- Onu depends on the number of massless species in MNu. From this perspective this line here is confusing -- what if I only have 1 massless species?

cosmology.c
```diff
@@ -33,17 +33,11 @@ void init_cosmology(Cosmology * CP_in)
...
-     * See Mangano et al 2005 (hep-ph/0506164)
-     * The correction is (3.046/3)^(1/4), for N_eff = 3.046 */
-    double TNu0_TCMB0 = pow(4/11., 1/3.) * 1.00328;
-
-    /* For massless neutrinos,
-     * rho_nu/rho_g = 7/8 (T_nu/T_cmb)^4 *N_eff,
-     * but we absorbed N_eff into T_nu above. */
-    CP->OmegaNu0 = CP->OmegaG * 7. / 8 * pow(TNu0_TCMB0, 4) * 3;
+    init_omega_nu(&CP->ONu, CP->MNu, TimeBegin, CP->HubbleParam, CP->CMBTemperature);
+    /* Neutrinos will be included in Omega0, if massive.
```
rainwoodman: This convention is inconvenient. I remember redshift dependency of massive neutrino is not a**3 -- due to the distribution function. So If you include Oncdm in Omega0, then special care has to be taken for the redshift dependency...

rainwoodman: We may want to drastically change the input parameters such that it takes a proper set (CLASS set) of parameters instead of the crippled Omega0.

cosmology.c
```diff
@@ -57,15 +51,12 @@ double hubble_function(double a)
...
     hubble_a += CP->OmegaK / (a * a);
-    hubble_a += CP->Omega0 / (a * a * a);
+    hubble_a += (CP->OmegaCDM + CP->OmegaBaryon) / (a * a * a);
 
     if(CP->RadiationOn) {
         hubble_a += CP->OmegaG / (a * a * a * a);
-        /* massless neutrinos are added only if there is no (massive) neutrino particle.*/
-        if(CP->MasslessNeutrinosOn)
-            hubble_a += CP->OmegaNu0 / (a * a * a * a);
+        hubble_a += get_omega_nu(&CP->ONu, a);
```
rainwoodman: What about the massive neutrino?

rainwoodman: It currently is not in OmegaCDM.

forcetree.c
```diff
@@ -753,7 +753,10 @@ force_update_node_recursive(int no, int sib, int tail, const struct TreeBuilder
             /* add all particles in this tree-node */
             int next = p;
             while(next != -1) {
-                add_particle_moment_to_node(&Nodes[no], &P[next]);
+                /*Hybrid particle neutrinos do not gravitate at early times.
+                 * So do not add their masses to the node*/
+                if(!All.HybridNeutrinosOn || All.Time > All.HybridNuPartTime || P[next].Type != All.FastParticleType)
+                    add_particle_moment_to_node(&Nodes[no], &P[next]);
```
rainwoodman: It might make more sense to move this if into add_particle_momentum_to_node.

genic/main.c
```diff
@@ -29,13 +30,22 @@ int main(int argc, char **argv)
...
+  init_cosmology(&CP, InitTime);
+
   initialize_powerspectrum(ThisTask, InitTime, UnitLength_in_cm, &CP, &PowerP);
   petapm_init(Box, Nmesh, 1);
+  /*Initialise particle spacings*/
+  const double meanspacing = Box / Ngrid;
+  const double shift_gas = -ProduceGas * 0.5 * (CP.Omega0 - CP.OmegaBaryon) / CP.Omega0 * meanspacing;
+  double shift_dm = ProduceGas * 0.5 * CP.OmegaBaryon / CP.Omega0 * meanspacing;
+  double shift_nu = 0;
+  if(!ProduceGas && NGridNu > 0) {
```
rainwoodman: Does this suggest ProduceGas and NGridNu cannot be used the same time?

genic/params.c
```diff
@@ -32,12 +32,17 @@ create_parameters()
...
     param_declare_int(ps, "Nmesh", OPTIONAL, 0, "Size of the FFT grid used to estimate displacements. Should be > Ngrid.");
     param_declare_int(ps, "Ngrid", REQUIRED, 0, "Size of regular grid on which the undisplaced particles are created.");
+    param_declare_int(ps, "NgridNu", OPTIONAL, 0, "Number of neutrino particles created for hybrid neutrinos.");
     param_declare_int(ps, "Seed", REQUIRED, 0, "");
     param_declare_int(ps, "Unitary", OPTIONAL, 0, "If non-zero, generate unitary gaussians where |g| == 1.0.");
     param_declare_int(ps, "WhichSpectrum", OPTIONAL, 2, "Type of spectrum, 2 for file ");
     param_declare_double(ps, "Omega_fld", OPTIONAL, 0, "Energy density of dark energy fluid.");
     param_declare_double(ps, "w0_fld", OPTIONAL, -1., "Dark energy equation of state");
     param_declare_double(ps, "wa_fld", OPTIONAL, 0, "Dark energy evolution parameter");
+    param_declare_double(ps, "MNue", OPTIONAL, 0, "First neutrino mass in eV.");
```
rainwoodman: What about making this a list? we shall probably add param_declare_double_array.. Let's defer it to a new PR.

genic/main.c
```diff
@@ -46,19 +56,44 @@ int main(int argc, char **argv)
...
+  if(NGridNu > 0) {
+      const int64_t TotNu = (int64_t) NGridNu*NGridNu*NGridNu;
+      const double kBMNu = 3*CP.ONu.kBtnu / (CP.MNu[0]+CP.MNu[1]+CP.MNu[2]);
+      const double v_th = NU_V0(InitTime, kBMNu, UnitVelocity_in_cm_per_s);
+      const double total_frac = init_thermalvel(v_th, Max_nuvel/v_th, 0);
+      int i;
+      message(0,"F-D velocity scale: %g. Max particle vel: %g. Fraction of mass in particles: %g\n",v_th*sqrt(InitTime), Max_nuvel*sqrt(InitTime), total_frac);
+      setup_grid(shift_nu, TotNu);
+      for(i = 0; i< TotNu; i++)
+          add_thermal_speeds(P[i].Vel);
```
rainwoodman: Is this Peculiar (a dx /dt ) or a**2 dx / dt

gravpm.c
```diff
@@ -265,6 +294,86 @@ static double sinc_unnormed(double x) {
...
+static void compute_neutrino_power() {
+    if(!All.MassiveNuLinRespOn)
+        return;
+    /*Note the power spectrum is now in Mpc units*/
+    powerspectrum_sum(&PowerSpectrum, All.BoxSize*All.UnitLength_in_cm);
+    int i;
+    /*Get delta_cdm_curr , which is P(k)^1/2, and skip bins with zero modes:*/
+    int nk_nonzero = 0;
+    for(i=0;i<PowerSpectrum.size;i++){
+        if (PowerSpectrum.Nmodes[i] == 0)
```
rainwoodman: I think we shall eliminate these Nmodes[i] == 0 modes in the powerspectrum code -- here is where the linear binning works better -- at least all modes are non-empty!

gravpm.c
```diff
@@ -265,6 +294,86 @@ static double sinc_unnormed(double x) {
...
+        /*Broadcast the transfer size*/
+        MPI_Bcast(&(transfer_init.NPowerTable), 1,MPI_INT,0,MPI_COMM_WORLD);
+        /*Allocate the memory unless we are on task 0, in which case it is already allocated*/
+        if(ThisTask != 0)
+          transfer_init.logk = (double *) mymalloc("Transfer_functions", 2*transfer_init.NPowerTable* sizeof(double));
+        transfer_init.T_nu=transfer_init.logk+transfer_init.NPowerTable;
+        /*Broadcast the transfer table*/
+        MPI_Bcast(transfer_init.logk,2*transfer_init.NPowerTable,MPI_DOUBLE,0,MPI_COMM_WORLD);
+        /*Initialise delta_tot*/
+        delta_tot_init(&delta_tot_table, nk_nonzero, PowerSpectrum.k, PowerSpectrum.P, &transfer_init, All.Time);
```
rainwoodman: It may be useful to put all k-space-neutrino functions under the same prefix. not now though.

gravpm.c
```diff
@@ -299,6 +408,25 @@ void powerspectrum_compute(const int64_t k2, const int kpos[3], pfft_complex * c
 
 }
 
+/*Just read the power spectrum, without changing the input value.*/
+static void readout_power_spectrum(int64_t k2, int kpos[3], pfft_complex *value) {
```
rainwoodman: I have been reserving readout for configuration space resampling operations. What about call this measure_power_transfer?

gravpm.c
```diff
@@ -299,6 +408,25 @@ void powerspectrum_compute(const int64_t k2, const int kpos[3], pfft_complex * c
...
+     * k_x = kpos * 2pi / L
+     *
+     * */
+    int k;
+    for(k = 0; k < 3; k ++) {
+        double tmp = (kpos[k] * M_PI) / All.Nmesh;
+        tmp = sinc_unnormed(tmp);
+        f *= 1. / (tmp * tmp);
+    }
+    powerspectrum_compute(k2, kpos, value, f);
```
rainwoodman: powerspectrum_add_mode?

petaio.c
```diff
@@ -124,6 +125,106 @@ petaio_build_selection(int * selection,
...
+    }
+    BigArray deltas = {0};
+    size_t dims[2] = {0 , 1};
+    /*The neutrino state is shared between all processors,
+     *so only write on master task*/
+    if(ThisTask == 0)
+        dims[0] = nk*ia;
+    ptrdiff_t strides[2] = {8, 8};
+    big_array_init(&deltas, delta_tot, "=f8", 2, dims, strides);
+    petaio_save_block(bf, "Neutrino/deltas", &deltas);
```
rainwoodman: Perhaps saving k as well?And using captialized column names, Delta, K?

rainwoodman: Looks like you can set dims[2] to {nk, ia} -- the result will read into python as a 2-d table.

petapm.h
```diff
@@ -22,6 +22,12 @@ typedef struct {
     void (*readout)(int i, double * mesh, double weight);
 } PetaPMFunctions;
 
+typedef struct {
+    void (*global_readout)(int64_t k2, int kpos[3], pfft_complex * value);
+    void (*global_analysis)(void);
+    void (*global_transfer)(int64_t k2, int kpos[3], pfft_complex * value);
```
rainwoodman: May want to add arguments for the analysis result and parameter of the transfer.

rainwoodman:
There will be another 'minor' conflict after merging #143.

I have a few suggestions here and there but I don't recall very critical ones. 
I still think in the long run we will benefit from embedding CLASS to provide the cosmology and transfer functions. (dropping quite some code..) So if possible at all we shall use this chance to further massage our parametrizatoin of cosmology to align the with CLASS.

andreufont:
I've compiled this branch, and I'm able to run simple runs. Are there instructions / examples for how to specify neutrino masses, or this is not yet an option? The only difference that I see in the parameter files is in paramfile.gadget, where now we have:
MassiveNuLinRespOn = 0
What does this exactly do? Shouldn't we also have modifications in the initial conditions? 

Thanks!

genic/main.c
sbird: No, that's fine. In that gas and DM are shifted, and NU sits in the middle on the grid points. As long as they do not initially have the same positions it is ok.

allvars.h
sbird: You're right, I will fix that.

begrun.c
sbird: You mean to organise the parameters for a module into a short struct, which is in turn a member of All? HybridNeutrinos is a sub-option of MassiveNuLinResp, indeed.

begrun.c
sbird: The convention here is that all neutrino masses are computed by get_omega_nu, whether massless or massive. There is no split between massless and massive neutrinos at all. MassiveNuLinRespOn controls only whether the neutrinos are supposed to be a component of Omega0.

begrun.c
sbird: Yes - this is a wrong condition. It should read 
if(!All.MassiveNuLinRespOn)
OmegaTot += get_omega_nu()
Apologies, I wrote this late last night.

sbird:
@andreufont : Look at examples/neutrinos, which should be an indicator of how to use it. You need to set MNue, MNum, MNut (although I will shortly make that a list) and MassiveNuLinRespOn = 1, both in  Gadget and in the ICs. There is documentation for the neutrino module also in kspace-neutrinos/README.

andreufont:
Thanks Simeon. The example runs (after fixing the parameter file for WindOn = 0). I'll play more with it tomorrow.

begrun.c
rainwoodman: At least a first clean up is the parameter names, `StarFormation.Enabled, StarFormation.Wind.Enabled, StarFormation.Wind.Model StarFormation.Wind; And neutrino goes to Neutrino.Enabled, Neutrino.Hybrid.Enabled, ... Using structs (like Cosmology) in the internal storage will ease later sharing the parser code between GenIC and Gadget too.

cosmology.c
rainwoodman: if get_omega_nu returns the sum of massive and massless nu, this line looks even stranger ... aren't you supposed to divide this by `a**3` or `a**4`?

sbird: No, it returns the sum of the massive neutrinos as a redshift dependent property: OmegaNu(a). For high masses we have OmegaNu(a) ~ MNu/(93.15 h^2) and for low masses OmegaNu(a) ~a^-4. Because the neutrino mass interpolates between the two (noticeably so) it seemed the least confusing thing to do

genic/params.c
sbird: Ok - that is a good idea but indeed let us defer it.

begrun.c
sbird: I like this idea but I wonder if we should do it all at once, in a separate (later) pull. It is kind of strange when only the neutrino parameter obey this convention. Also it would be nice to have some kind of support for these structs from the param reading code - you had some idea to embed lua for reading parameters at one point, didn't you?

gravpm.c
sbird: The linear binning actually produces a noticeably less accurate neutrino power spectrum, because of the excess smoothing induced (I checked this).

petaio.c
sbird: Is it ok to do that and read it back into a flat memory block in C?

petapm.h
sbird: void * pointers to preserve the generic algorithm property?

sbird:
Rebased, all but a couple of comments addressed and the paramfile updated.

petapm.h
sbird: There is no-where for them to go - let me leave it as it is. It sets some structures which are effectively private members of the gravpm object. The other way I would have to pass the structures from petapm back to gravpm which would be a bit messy.

genic/main.c
sbird: Peculiar a dx/dt, I think.

sbird:
Ok, I have addressed all the comments I want to - as far as I'm concerned this is ready to merge. Let's wait a few days and see if @andreufont finds anything wrong with it?

petaio.c
rainwoodman: It is still a flat memory block as long as the strides are correctly set.. I think you want
```
strides[2] = {8 * ia, 8}
dims[2] = {nk or 0, ia}
```
This will give you a tight representation of the data in C after reading.

gravpm.c
rainwoodman: excessive smoothing at low k? because log binning have excessive smoothing at high k.

I think what you really want is, probably something per mode at low-k but then averaged at high k.

begrun.c
rainwoodman: OK. Let's plan for a massive parameter overhual later.

petaio.c
sbird: Thanks - does what I have now look right?

gravpm.c
sbird: Excessive smoothing at low k, indeed. Excessive smoothing at high k is ok for the neutrinos (even physical!) I think what you suggest is exactly what I really want.

cosmology.c
rainwoodman: This touches how k-space-neutrino is implemented and I don't think it can be addressed in this PR; especially given that you are also embedding it in other places..

If you think of neutrino as a single fluid then yes, I see how the current get_omega_nu is the least confusing approach. massive neutrino still has a pressure component that scales like a^-4. I think that's likely what you mean by the neutrino mass interpolates between the two. 

However, after studying class somewhat carefully, I tend to agree with class where it divides neutrinos into components in terms of how they scale with a: ur (a**-4), mass-like (a**-3), pressure-like (a**-4), like here:

https://github.com/lesgourg/class_public/blob/master/source/background.c#L354

genic/main.c
rainwoodman: You have to be careful, because the unit of P[i].Vel depends on whether UsePecuiliar Velocity is set.

gravpm.c
rainwoodman: We've been talking about doing this as an improvement of the model you implemented here. (actually a student here was supposed to work on this). Perhaps getting him to look at this after it is merged?

sbird: That would be great.

petapm.h
rainwoodman: Sure. when we start to do the next round of global variables cleanups ...

rainwoodman:
We can also merge this first, then merge the fixes to master if andreu finds any. The bases for this consideration is that big PR tends to develop conflicts. If it is already merged we don't have to worry about these any more.

Another interesting dynamics we shall be aware is the submodule embedding of k-space-neutrino. It is the first 'science' module that we externally depend on. We only had utility modules previously.
Science modules actually representing the core business logic, but utility modules do not touch the core business. 

It is harder to insulate core business logic modules -- for example -- the choice of nu parametrization in k-space-neutrino dictates how nu shall be parametrized in MP-Gadget. Modifications of k-space-neutrino will have to propagate to other places, making it more costly. 

We will see how this plays out over time.

genic/main.c
sbird: Well noticed - I had not accounted for that. Should be fixed now.

petapm.h
sbird: A never-ending task

petaio.c
```diff
@@ -197,14 +198,15 @@ void petaio_read_neutrinos(BigFile * bf)
...
     /*The neutrino state is shared between all processors,
      *so only read on master task and broadcast*/
-    if(ThisTask == 0)
-        dims[0] = nk*ia;
+    if(ThisTask == 0) {
+        dims[0] = nk;
+    }
     big_array_init(&deltas, delta_tot, "=f8", 2, dims, strides);
-    petaio_read_block(bf, "Neutrino/deltas", &deltas, 1);
+    petaio_read_block(bf, "Neutrino/Deltas", &deltas, 1);
```
rainwoodman: Is there a better name than Deltas? DeltaNuTotHistory?

gravpm.c
rainwoodman: Ok. I'll start nudging him. It is relatively simple but will likely be a mixture of commits on K-space-neutrino and MP-Gadget as I see now.

cosmology.c
sbird: I think the reason I like the single fluid is that it matches how I am treating the perturbations, and it matches well with how particles are treated in neutrino particle simulations: they are a single fluid, described by particles. It also has a neat interpretation in terms of the maximum neutrino velocity for the hybrid particles: you can set Max_nuvel = C/4 (say) and then the relativistic component is taken care of automatically.

sbird:
That is indeed a bit of a concern - if this becomes too difficult we could instead import the C files into MP-Gadget directly.

cosmology.c
rainwoodman: I see. The hybrid particles deals with mass-like and pressure-like massive neutrinos the same time. What about the ultra relativistic species?

sbird: Not sure what you mean exactly? pressure-like massive neutrinos are the ultra-relativistic species?

rainwoodman:
Conflict again!

sbird:
Fortunately a trivial one - maybe we should merge this before rebase errors start cropping up. If you are happy with the current state I will run one last long test with the current master and then we could merge?

cosmology.c
rainwoodman: No. I mean there is a pressure-like component to the overall density due to massive neutrino that also scales like a**-4, in this line:
https://github.com/lesgourg/class_public/blob/master/source/background.c#L380

ultra-relativstic is already a**-4, at:
https://github.com/lesgourg/class_public/blob/master/source/background.c#L418

sbird: Ah, I see. So you are proposing adding support for an arbitrary amount of ultra-relativistic species? 

At present we have OmegaG, which is so called because it is actually photons (and the parameter we set is the CMBTemperature). I think just renaming it OmegaUR would be not the right thing to do, because it is set up to be actual photons. But I think you are suggesting adding an extra species that redshifts like a^-4 and could map onto, for example, dark photons?

rainwoodman:
no dark photons, yet. ur == massless neutrinos in CLASS convention. ncdm ==
massive neutrinos.

sbird:
So class splits massive and massless neutrinos not for aesthetics but for internal reasons related to the approximations made by Boltzmann solver; massless neutrino transfer functions are basically free for class to compute (so much so that camb computes them even if there are no massless neutrinos), whereas massive neutrinos are one of the most expensive things it does.

The advantage of a single variable for the neutrinos is that physically they are the same species, so splitting them into a massive and massless part is a bit arbitrary.

PR #126, opened 2017-10-08
Cleanups

sbird:
712ef2200c2e9a is an actual bugfix which could cause a crash after spawning a blackhole.

All the rest are cleanups done while working on something else. 
I removed SPH_GRAD_RHO, cleaned up the code tangle in domain_exchange_once a little and merged init() with begrun().

### DIFF
--- a/exchange.c
+++ b/exchange.c
@@ -103,77 +103,83 @@ int domain_exchange(int (*layoutfunc)(int p)) {
     return failure;
 }
 
+#define NSP 3
+
 static int domain_exchange_once(int (*layoutfunc)(int p), int* toGo, int * toGoSph, int * toGoBh, int *toGet, int *toGetSph, int *toGetBh)
 {
-    int count_togo = 0, count_togo_sph = 0, count_togo_bh = 0, 
-        count_get = 0, count_get_sph = 0, count_get_bh = 0;
-
-    int i, target;
+    int i, j, target;
     struct particle_data *partBuf;
     struct sph_particle_data *sphBuf;
     struct bh_particle_data *bhBuf;
 
-    int * count = ta_malloc("var", int, NTask);
-    int * count_sph = ta_malloc("var", int, NTask);
-    int * count_bh = ta_malloc("var", int, NTask);
-    int * offset = ta_malloc("var", int, NTask);
-    int * offset_sph = ta_malloc("var", int, NTask);
-    int * offset_bh = ta_malloc("var", int, NTask);
+    int *ctmem = mymalloc("cts", 4*NSP*NTask*sizeof(int));
+    int *count[NSP];
+    int *offset[NSP];
+    int *count_recv[NSP];
+    int *offset_recv[NSP];
+    int count_togo[NSP] = {0};
+    int count_get[NSP] = {0};
 
-    int * count_recv = ta_malloc("var", int, NTask);
-    int * count_recv_sph = ta_malloc("var", int, NTask);
-    int * count_recv_bh = ta_malloc("var", int, NTask);
-    int * offset_recv = ta_malloc("var", int, NTask);
-    int * offset_recv_sph = ta_malloc("var", int, NTask);
-    int * offset_recv_bh = ta_malloc("var", int, NTask);
+    memset(ctmem, 0, 4*NSP*NTask*sizeof(int));
 
-    for(i = 1, offset_sph[0] = 0; i < NTask; i++)
-        offset_sph[i] = offset_sph[i - 1] + toGoSph[i - 1];
-
-    for(i = 1, offset_bh[0] = 0; i < NTask; i++)
-        offset_bh[i] = offset_bh[i - 1] + toGoBh[i - 1];
-
-    offset[0] = 0;
-
-    for(i = 1; i < NTask; i++)
-        offset[i] = offset[i - 1] + toGo[i - 1];
-
-    for(i = 0; i < NTask; i++)
-    {
-        count_togo += toGo[i];
-        count_togo_sph += toGoSph[i];
-        count_togo_bh += toGoBh[i];
-
-        count_get += toGet[i];
-        count_get_sph += toGetSph[i];
-        count_get_bh += toGetBh[i];
+    for(j=0; j<NSP; j++) {
+        count[j] = ctmem+j*NTask;
+        offset[j] = ctmem + 4*NSP * NTask +j*NTask;
+        count_recv[j] = ctmem + 2*4*NSP * NTask +j*NTask;
+        offset_recv[j] = ctmem + 3*4*NSP * NTask +j*NTask;
     }
+
+    /*Build arrays*/
+    int * toGo_arr[NSP];
+    toGo_arr[0] = toGo;
+    toGo_arr[1] = toGoSph;
+    toGo_arr[2] = toGoBh;
+
+    int * toGet_arr[NSP];
+    toGet_arr[0] = toGet;
+    toGet_arr[1] = toGetSph;
+    toGet_arr[2] = toGetBh;
+
     int bad_exh=0;
 
-    /*Check whether the domain exchange will succeed. If not, bail*/
-    if(NumPart + count_get - count_togo> All.MaxPart){
-        message(1,"Too many particles for exchange: NumPart=%d count_get = %d count_togo=%d All.MaxPart=%d\n", NumPart, count_get, count_togo, All.MaxPart);
-        bad_exh = 1;
-    }
-    if(N_sph_slots + count_get_sph - count_togo_sph > All.MaxPart) {
-        message(1,"Too many SPH for exchange: N_sph=%d All.MaxPart=%d\n", N_sph_slots + count_get_sph, All.MaxPart);
-        bad_exh = 1;
-    }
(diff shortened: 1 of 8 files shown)


rainwoodman:
Sorry but SPH_GRAD_RHO is used as one of the starformation conditioner for molecular hydrogen. It is quite relevant at high z. We actually use that in BlueTides.

rainwoodman:
What about cherry-picking the commits that doesn't remove SPH_RHO_GRAD -- it can be converged to a ``All.*`` variable, which is computed from the star-formation criteria used.

sbird:
Oh, apologies, I didn't realise we used it. I'll drop that commit.

sbird:
How's that? I dropped the cleanup of init.c for now as it conflicts with open PRs.

exchange.c
```diff
@@ -103,29 +103,22 @@ int domain_exchange(int (*layoutfunc)(int p)) {
...
-
-    int * count_recv = ta_malloc("var", int, NTask);
-    int * count_recv_sph = ta_malloc("var", int, NTask);
-    int * count_recv_bh = ta_malloc("var", int, NTask);
-    int * offset_recv = ta_malloc("var", int, NTask);
-    int * offset_recv_sph = ta_malloc("var", int, NTask);
-    int * offset_recv_bh = ta_malloc("var", int, NTask);
+    INIT_EXTRA(count);
+    INIT_EXTRA(offset);
+    INIT_EXTRA(count_recv);
```
rainwoodman: We really should have had one per particle species -- I was thinking of introducing an array and a loop somehow. But the macro, hinders thand...

rainwoodman:
Why would it crash the code if the new BH is active? I worry there is another bug that causes the crash and by protecting the timebin you are hiding it instead of fixing it. In principle we shall be able to add any particle to the active list and the code shouldn't crash.

rainwoodman:
Note that the grad_rho removal is not dropped yet.

sbird:
If the BH is made active when the gas particle it is created from is not active, it is active but in an inactive TimeBin. After creation the BH particle receives a kick, which it should not receive (because it is in an inactive timebin). So then the Kick time is not right for this bin, and the code errors with Ti_drift != Ti_kick. I think this doesn't actually matter that much, because the BH is repositioned anyway, but it is good to check it.

sbird:
I think SPH_GRAD_RHO is still there - what is removed is the SPH_GRAD_RHO guards around the selfgravity star formation (they are not necessary)

exchange.c
sbird: good point...

rainwoodman:
I see what you mean. What we can do is to modify the timestep calculator such that if IsNewParticle (or another flag, e.g. WakeUp) is set, then the time step of the particle is reduced such that it is awake on the next active bin.

init.c
```diff
@@ -292,11 +279,9 @@ setup_smoothinglengths(int RestartSnapNum)
         myfree(olddensity);
 #endif //DENSITY_INDEPENDENT_SPH
 #pragma omp parallel for
-        for(i = 0; i < NumPart; i++) {
-            if(P[i].Type == 0) {
-                /* EgyWtDensity stabilized, now we convert from energy to entropy*/
-                SPHP(i).Entropy = GAMMA_MINUS1 * u_init / pow(SPHP(i).EOMDensity/a3 , GAMMA_MINUS1);
-            }
+        for(i = 0; i < N_sph_slots; i++) {
```
rainwoodman: I am not sure -- there are cons and pros directly accessing SphP -- This way if N_sph < N_sph_slots you may end up having some nans in SphP[i]?

sbird: At init this is ok, I think. I can check for EOMDensity == 0 if you like.

sbird:
Current master doesn't work for me on the hydro example, when starting from a snapshot (snapshot 2). The Systemstep time is zero, and it loops for two timesteps output before exiting with the message that inttimes are out of sync. This is since the garbageflag merge. Can you take a look?

init.c
rainwoodman: I slightly prefer to access SphP via the accessor than directly meddling with the underlying storage.

It is slower but the API is more consistent. We shall make garbage.c the only place where SphP / BhP is directly modified (and promote it to slotmanager.c).

If the skip by ptype bothers you, we may want to have the indirect sorting index by PType, like the one used in petaio -- the indirect sorting index did significant improved speed for preparing IO buffer of BH and STAR there.

sbird:
I don't know if the suggested fix works...the particle is mid-timestep at that point and it is usually wrong to shorten it like that. Also it only happens for BH seeding and it does not matter if the newly seeded BH waits a little to become active.

rainwoodman:
You are right. To WakeUP a particle we need to do some additional kick to move Ti_kick to Ti_drift according to the old acceleration before giving them new time steps.

rainwoodman:
The current master worked for me if I regenerate an IC file -- you probably used an old snapshot that didn't have TimeIC. I think the code shall crash instead of trying to recover.

There is a glitch that it still writes a snapshot immediately upon restart. The '-1' at IC hack is the only way to gracefully deal with this I am afraid.

init.c
sbird: Ok, no worries, I will drop the commit.

rainwoodman:
I fixed the redundant snapshot issue in #128

rainwoodman:
Consider a rebase -- the new master checks restart in travis.

sbird:
For me it was failing to restart only when I used snapshot 2, at =0.15 , even with new ICs. This is also the first snapshot with one BH in. New master works fine, probably because it is not trying to do fof/snapshot immediately.

exchange.c
```diff
@@ -103,77 +103,83 @@ int domain_exchange(int (*layoutfunc)(int p)) {
...
+        for(i = 1; i < NTask; i++)
+            offset[j][i] = offset[j][i - 1] + toGo_arr[j][i - 1];
+        /*Compute counts*/
+        for(i = 0; i < NTask; i++)
+        {
+            count_togo[j] += toGo_arr[j][i];
+            count_get[j] += toGet_arr[j][i];
+        }
+        char *nn;
+        switch (j) {
```
rainwoodman: const char * nn[3] = {"PARTICLES", "SPH", "BH"}?

then nn[j]?

exchange.c
```diff
@@ -103,77 +103,83 @@ int domain_exchange(int (*layoutfunc)(int p)) {
...
-    int * offset_recv_bh = ta_malloc("var", int, NTask);
-
-    for(i = 1, offset_sph[0] = 0; i < NTask; i++)
-        offset_sph[i] = offset_sph[i - 1] + toGoSph[i - 1];
-
-    for(i = 1, offset_bh[0] = 0; i < NTask; i++)
-        offset_bh[i] = offset_bh[i - 1] + toGoBh[i - 1];
-
-    offset[0] = 0;
+    int *ctmem = mymalloc("cts", 4*NSP*NTask*sizeof(int));
```
rainwoodman: does it hurt to keep NSP at 6 and use the ptype to index these arrays?

rainwoodman: No it is more complicated because we are dealing with the base as well.

exchange.c
```diff
@@ -254,46 +249,39 @@ static int domain_exchange_once(int (*layoutfunc)(int p), int* toGo, int * toGoS
...
-    MPI_Alltoallv_sparse(sphBuf, count_sph, offset_sph, MPI_TYPE_SPHPARTICLE,
-                 SphP, count_recv_sph, offset_recv_sph, MPI_TYPE_SPHPARTICLE,
+    MPI_Alltoallv_sparse(sphBuf, count[1], offset[1], MPI_TYPE_SPHPARTICLE,
+                 SphP, count_recv[1], offset_recv[1], MPI_TYPE_SPHPARTICLE,
                  MPI_COMM_WORLD);
     walltime_measure("/Domain/exchange/alltoall");
 
-    MPI_Alltoallv_sparse(bhBuf, count_bh, offset_bh, MPI_TYPE_BHPARTICLE,
-                BhP, count_recv_bh, offset_recv_bh, MPI_TYPE_BHPARTICLE,
+    MPI_Alltoallv_sparse(bhBuf, count[2], offset[2], MPI_TYPE_BHPARTICLE,
```
rainwoodman: With a bit more work we can move these duplicated into a loop as well. I can take care of it after merging this.

rainwoodman:
This sounds strange. But does it run all the way to produce a next snapshot?

We shall probably change / add travis test with a hydro case. But let's merge this first?

sbird:
It runs all the way to z=0 just fine...I am not sure what is going on. Possibly it is to do with all the particles being in TimeBin 0, which only happens on the first timestep.

exchange.c
sbird: Good point!

exchange.c
sbird: Yes, because 0 is "total": we would need NSP = 7 and then it gets complex.

exchange.c
sbird: Thanks. I did not see a way to do that!

rainwoodman:
https://github.com/bccp/abopt/blob/master/.travis.yml#L28

You'll need to encrypt your password with the travis command line tool though.

PR #133, opened 2017-10-11
Initialise TimeBinCount properly.

sbird:
On the first timestep, all particles are in TimeBin 0. Set TimeBinCount[0] appropriately, so that sum(TimeBinCount) == NumPart at all times.

This fixes a big: writing a FoF on the first timestep may remove particles from a processor. If TimeBinCount is initially zero, it will become negative, with causing the timestep to become zero and an endless loop.

### DIFF
--- a/timestep.c
+++ b/timestep.c
@@ -58,6 +58,7 @@ static inttime_t get_long_range_timestep_ti(const inttime_t dti_max);
 void
 init_timebins(double TimeInit)
 {
+    int i;
     All.Ti_Current = ti_from_loga(TimeInit);
     /*Enforce Ti_Current is initially even*/
     if(All.Ti_Current % 2 == 1)
@@ -67,6 +68,11 @@ init_timebins(double TimeInit)
     PM.Ti_kick = All.Ti_Current;
     PM.start = All.Ti_Current;
     update_active_timebins(0);
+    /*This is here so that TimeBinCount is accurate on the first timestep.
+     *      * Avoids potential particle exchange making counts negative.*/
+    TimeBinCount[0] = NumPart;
+    for(i=0; i<6; i++)
+        TimeBinCountType[i][0] = NLocal[i];
 }
 
 int is_timebin_active(int i) {
@@ -775,7 +781,7 @@ inttime_t find_next_kick(inttime_t Ti_Current)
     int n;
     for(n = 0; n < TIMEBINS; n++)
     {
-        if(TimeBinCount[n])
+        if(TimeBinCount[n] > 0)
             break;
     }
     MPI_Allreduce(MPI_IN_PLACE, &n, 1, MPI_INT, MPI_MIN, MPI_COMM_WORLD);


rainwoodman:
I think this a good fix. But could you explain the trigger scenario this breaks a bit more -- FOF will exchange particles, and reduce TimeBinCount. But after FOF the particles are all returned to the original rank, and TimeBinCount is reconstructed anyways -- why does it trigger a negative count then?

sbird:
I'm a little worried about that now: I think the reduction in TimeBinCount is just because domain_maintain is called during FoF. But reconstruct_timebins is called before FoF is done, and that should indeed have initialised TimeBinCount to something large itself, whereas it seems to be remaining as zero. 

So now I am worried I did not root-cause it correctly.

rainwoodman:
But you did confirm it fixed the crash -- do you have a snapshot that crashes that I can take a look?

sbird:
I did confirm that it fixed the crash - it was just ICs. I used the hydro example with 2 MPIs, one thread and 128^3 particles. It went away if I set SnapshotWithFOF = 0.

sbird:
Thinking about it the only way this patch can have fixed the patch is because it changed the order in which TimeBinCount was allocated, and perhaps thus its position on the stack. Which suggests some array is writing out of bounds.

rainwoodman:
Could be memory issue. I didn't see it when I ran the hydro (I compiled with debug).

It must be after the new allocator is merged. Disable the patch, then run GDB with a watch on the malicious TimeBinCount element?

sbird:
Here is the logic:

First rebuild_activelist() sets up TimeBinCount[0] = NumPart
Then FoF calls domain_maintain and moves particles.  TimeBinCount is decremented for each particle that moves away from the current processor. However it is not incremented for particles that move to the current processor.
find_timesteps is called and moves all particles to other timebins, again decrementing TimeBinCount, which ends up negative.

The patch fixes the crash because TimeBinCount is checked > 0, instead of != 0.

I am not sure why the same crash does not happen at later timesteps. Probably just because it is rare for a timestep to be completely emptied. 

The fix is easy: increment TimeBinCount in exchange.c.

sbird:
There are still two things I do not understand:
1) Why is domain_maintain necessary in fof? Can't we could just do force_tree_rebuild?
2) Why does domain_maintain move particles at all? The Key is not changing, as drift is not called. Is there some corruption in the TopNodes or TopLeaves structures?

rainwoodman:
I can answer 1: it is because after FOF particle ordering is wrong, even though they are returned to the correct ranks. The assumption is that domain_maintain maintains the sorted ordering of particles, which is needed for the tree build.

For 2: What about take a look at NumPart before / after FOF and domain_maintain?

sbird:
So for 1) the domain_maintain is to ensure that ReverseLink in the secondary arrays are up to date? Can we do that with a domain_garbage_collection() force_tree_rebuild()?

For 2) I will do that!

rainwoodman:
1) No; ReverseLink is only used during GC. 

But looks like force_tree_build does not require the particles to be sorted, does it?

And domain_maintain doesn't do any sorting at all anyways. then yes domain_gc and force_tree_rebuild should work. You can even skip the GC as it is slow too..

But let's see if NumPart is conserved before / after FOF.

sbird:
Confirmed: NumPart is conserved before/after FoF. The error in TimeBinCount occurs because when the domain_exchange is called in fof_save_particles, it is decremented but not incremented again. I have implemented the fix and will send a pr when a test run finishes.

rainwoodman:
Ah. It is a bug in exchange -- GC reduces the counts when called in exchange. The problem is masked when there is no FOF because we recount the bins anyhow. But in FOF it is not done. So exchange shall take care to walk through newly received particles and update the timebins.

PR #134, opened 2017-10-13
Fix time bin count

sbird:
We found that TimeBinCount was not updated correctly in domain exchange: it was decremented for particles leaving, but not incremented for particles arriving. It was also not incremented correctly in sfr and black hole. Normally this did not matter, because rebuild_activelist was called after domain, but during FoF particles are exchanged (then returned) and rebuild_activelist is not called.

I considered a fix involving updating the TimeBinCount, but I found a simpler solution was just to avoid the need to keep it up to date. Now it is used only for printing counts just after domain. The smallest timebin is returned by find_timestep. This avoids a number of atomic updates in openmp code.

Also do not do domain_maintain after FoF: it always exchanged zero particles.

### DIFF
--- a/density.c
+++ b/density.c
@@ -418,8 +418,9 @@ density_postprocess(int i, TreeWalk * tw)
                 SPHP(i).DhsmlEgyDensityFactor *= -SPHP(i).DhsmlDensityFactor;
                 SPHP(i).EgyWtDensity /= EntPred;
             } else {
-                SPHP(i).DhsmlEgyDensityFactor=0;
-                SPHP(i).EgyWtDensity=0;
+                /*Use non-weighted densities for this*/
+                SPHP(i).DhsmlEgyDensityFactor=SPHP(i).DhsmlDensityFactor;
+                SPHP(i).EgyWtDensity=SPHP(i).Density;
             }
 #endif
 
--- a/density.c
+++ b/density.c
@@ -418,7 +418,9 @@ density_postprocess(int i, TreeWalk * tw)
                 SPHP(i).DhsmlEgyDensityFactor *= -SPHP(i).DhsmlDensityFactor;
                 SPHP(i).EgyWtDensity /= EntPred;
             } else {
-                /*Use non-weighted densities for this*/
+                /* Use non-weighted densities for this.
+                 * This should never occur normally,
+                 * but will happen for every particle during init().*/
                 SPHP(i).DhsmlEgyDensityFactor=SPHP(i).DhsmlDensityFactor;
                 SPHP(i).EgyWtDensity=SPHP(i).Density;
             }
--- a/peano.h
+++ b/peano.h
@@ -15,6 +15,7 @@ morton_t morton_key(int x, int y, int z, int bits);
 
 static inline peano_t PEANO(double *Pos, double BoxSize)
 {
+    /*No reason known for the Box/2000 and 1.001 factors*/
     const double DomainFac = 1.0 / (BoxSize*1.001) * (((peano_t) 1) << (BITS_PER_DIMENSION));
     const double spos[3] = {Pos[0] + BoxSize/2000, Pos[1] + BoxSize/2000, Pos[2] + BoxSize/2000};
     return peano_hilbert_key(spos[0]*DomainFac, spos[1]*DomainFac, spos[2]*DomainFac, BITS_PER_DIMENSION);
--- a/peano.h
+++ b/peano.h
@@ -10,12 +10,17 @@ typedef uint64_t morton_t;
 #define  BITS_PER_DIMENSION 21	/* for Peano-Hilbert order. Note: Maximum is 10 to fit in 32-bit integer ! */
 #define  PEANOCELLS (((peano_t)1)<<(3*BITS_PER_DIMENSION))
 
-#define DomainFac(len) ( 1.0 / (len) * (((peano_t) 1) << (BITS_PER_DIMENSION)))
+peano_t peano_hilbert_key(int x, int y, int z, int bits);
+morton_t morton_key(int x, int y, int z, int bits);
 
-#define PEANO(Pos) peano_hilbert_key((int) ((Pos[0] + All.BoxSize/2000) * DomainFac(All.BoxSize*1.001)), \
-        (int) ((Pos[1] + All.BoxSize/2000) * DomainFac(All.BoxSize*1.001)), \
-        (int) ((Pos[2] + All.BoxSize/2000) * DomainFac(All.BoxSize*1.001)), \
-        BITS_PER_DIMENSION)
+static inline peano_t PEANO(double *Pos, double BoxSize)
+{
+    const double DomainFac = 1.0 / (BoxSize*1.001) * (((peano_t) 1) << (BITS_PER_DIMENSION));
+    const double spos[3] = {Pos[0] + BoxSize/2000, Pos[1] + BoxSize/2000, Pos[2] + BoxSize/2000};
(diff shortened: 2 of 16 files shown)


rainwoodman:
Travis caught another place for the PEANO change. Perhaps a missing header file in Makefile.rules.

peano.h
```diff
@@ -10,12 +10,17 @@ typedef uint64_t morton_t;
...
 
-#define DomainFac(len) ( 1.0 / (len) * (((peano_t) 1) << (BITS_PER_DIMENSION)))
+peano_t peano_hilbert_key(int x, int y, int z, int bits);
+morton_t morton_key(int x, int y, int z, int bits);
 
-#define PEANO(Pos) peano_hilbert_key((int) ((Pos[0] + All.BoxSize/2000) * DomainFac(All.BoxSize*1.001)), \
-        (int) ((Pos[1] + All.BoxSize/2000) * DomainFac(All.BoxSize*1.001)), \
-        (int) ((Pos[2] + All.BoxSize/2000) * DomainFac(All.BoxSize*1.001)), \
-        BITS_PER_DIMENSION)
+static inline peano_t PEANO(double *Pos, double BoxSize)
```
rainwoodman: I like keeping it a static inline function.

rainwoodman: We don't have a concrete case (other than precaution) to justify the factor of 1.001, do we? Either case lets note it in the code, and update when someone figures it out.

run.c
```diff
@@ -213,8 +218,7 @@ void run(void)
         /* more steps to go. */
 
         /* assign new timesteps to the active particles, now that we know they have synched TiKick and TiDrift */
-
-        find_timesteps();
+        minTimeBin = find_timesteps();
```
rainwoodman: I suggest using ``int find_timesteps(int * MinTimeBin)``; in the longer term we shall probably make all functions like this, because it makes it mentally easier to add more output parameters.

timestep.c
```diff
@@ -772,41 +751,106 @@ void rebuild_activelist(void)
...
+     * calculation in every timestep.
+     * */
+
+    char extra[1024] = {0};
+
+    if(is_PM_timestep(All.Ti_Current))
+        strcat(extra, "PM-Step");
+
+    z = 1.0 / (All.Time) - 1;
+    message(0, "Begin Step %d, Time: %g, Redshift: %g, Nf = %014ld, Systemstep: %g, Dloga: %g, status: %s\n",
```
rainwoodman: I've been thinking it would be nice to see the integer step, especially now that it is consistent between runs.

timestep.c
```diff
@@ -772,41 +751,106 @@ void rebuild_activelist(void)
...
+        int j;
+        for(j=0; j<6; j++)
+            tot_count[i] += tot_count_type[j][i];
+        if(is_timebin_active(i))
+            tot_num_force += tot_count[i];
+    }
+
+    /* let's update Tot counts in one place tot variables;
+     * at this point there can still be holes in SphP
+     * because rearrange_particle_squence is not called yet.
```
rainwoodman: The comments here can be updated. It is GC, no longer rearrange_particle_sequence.
And the comment below is no longer applicable. 

Also additional comment: if the numbers are only used for printing, we can perhaps mark them static ?

timestep.c
```diff
@@ -772,41 +751,106 @@ void rebuild_activelist(void)
...
-            break;
-    }
-    MPI_Allreduce(MPI_IN_PLACE, &n, 1, MPI_INT, MPI_MIN, MPI_COMM_WORLD);
     /* Current value plus the increment for the smallest active bin. */
-    return Ti_Current + dti_from_timebin(n);
+    return Ti_Current + dti_from_timebin(minTimeBin);
 }
 
 /* mark the bins that will be active before the next kick*/
 int update_active_timebins(inttime_t next_kick)
```
rainwoodman: This function perhaps shall take MinTimeBin sounded like anything below mintimbin will be activate.

peano.h
sbird: We do not; I just kept it because that is what it always has been. I'll add a comment

timestep.c
sbird: ?

rainwoodman: Sorry I typed this in a rush. I mean the logic of update_active_timebins can be simplified if minTimeBin is passed in. The code apparently is just trying to rebuild that information from next_kick, which is inferred from minTimeBin.

sbird: It can be simplified out of existence

timestep.c
```diff
@@ -65,11 +64,15 @@ init_timebins(double TimeInit)
...
-    update_active_timebins(0);
 }
 
-int is_timebin_active(int i) {
-    return TimeBinActive[i];
+int is_timebin_active(int i, inttime_t current) {
+    /*Bin 0 is always active and at time 0 all bins are active*/
+    if(i == 0 || current == 0)
+        return 1;
+    if(current % dti_from_timebin(i) == 0)
```
rainwoodman: perhaps it shal be == dti_from_timebin(i) / 2?

sbird: No, because dti_from_timebin is the delta t?

rainwoodman: I see. 
Our scheme is D f k t k. 

- before t, due to k, the particle has synced ti_kick and ti_drift with ti_current.
Therefore it is safe to move particle to a new time step size. 

- before f, due to  D ti_drift and ti_current are the same.
- f needs to know if a particle is active. At this time ti_kick must be earlier than ti_drift. the particle is active iif ti_drift % dti == 0.

A particle is active == force will be computed, kick will be applied

sbird: Right!

density.c
```diff
@@ -418,8 +418,11 @@ density_postprocess(int i, TreeWalk * tw)
                 SPHP(i).DhsmlEgyDensityFactor *= -SPHP(i).DhsmlDensityFactor;
                 SPHP(i).EgyWtDensity /= EntPred;
             } else {
-                SPHP(i).DhsmlEgyDensityFactor=0;
-                SPHP(i).EgyWtDensity=0;
+                message(1, "Something weird: Particle %d (ID %ld): Entropy %g, EgyWtDensity %g\n",
```
rainwoodman: Not that weird -- the first initialization density EntPred and EgyWtDensity are always zero. This protects against that -- do you see a horde of warning messages after this when hydro starts?

sbird: Ah indeed! Well noticed. I will amend the commit.

sbird: When I am ready to merge - for now it is proving a useful warning.

density.c
```diff
@@ -418,8 +418,9 @@ density_postprocess(int i, TreeWalk * tw)
                 SPHP(i).DhsmlEgyDensityFactor *= -SPHP(i).DhsmlDensityFactor;
                 SPHP(i).EgyWtDensity /= EntPred;
             } else {
-                SPHP(i).DhsmlEgyDensityFactor=0;
-                SPHP(i).EgyWtDensity=0;
+                /*Use non-weighted densities for this*/
+                SPHP(i).DhsmlEgyDensityFactor=SPHP(i).DhsmlDensityFactor;
+                SPHP(i).EgyWtDensity=SPHP(i).Density;
```
rainwoodman: Doesn't it usually signals a bigger problem? How often do you see it zero? Perhaps add a comment about this?

sbird: I used it to find the bug in b5d1d95, but once that was fixed it is zero only on the first iteration of the first timestep. If we could make the first iteration non-zero it might indeed be better to do endrun() here instead, but I'm not sure how to do that. I'll add a comment.

sbird:
This pull definitely fixes several bugs, and passes tests - I think I'm ready to merge it. What do you think?

rainwoodman:
Yes. I agree. We have probably extracted all possible juice from this PR. I'll click the button. Thanks!

PR #135, opened 2017-10-17
Add gravity to BH, and save drifting position in snapshots.

### LINKED ISSUE #125: blackholes drifting is numerically unstable
Currently blackholes drifting is numerically unstable if the restarts are too frequent. For two reasons:

1. The min-pot-pos and min-pot-vel attributes are not stored in a snapshot file, thus first step (PM step) does not know about where to drift the blackholes.

2. The blackholes do not receive gravity, so if the hosting galaxy moves too fast, the BH will not be able to keep up.

The solution appears to be

a) to enable gravity on BH, and
b) store the min-pot-pos and min-pot-vel in snapshot files.

rainwoodman:
This should fix #125. Do you see problems Ananth?

### DIFF
--- a/examples/travis/paramfile.gadget
+++ b/examples/travis/paramfile.gadget
@@ -64,7 +64,7 @@ GravitySoftening = 0.03333
 GravitySofteningGas = 0.00 # enable adaptive softening of gas
 
 #----------------------BH Stuff-------------------------
-BlackHoleOn = 0
+BlackHoleOn = 1
 BlackHoleFeedbackFactor = 0.05
 BlackHoleFeedbackRadius = 0.
 BlackHoleFeedbackRadiusMaxPhys = 0.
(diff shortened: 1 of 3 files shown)


sbird:
Do we really want to produce BHs in genic? Can't we just reduce the seed mass to a low value so that travis always spawns the BHs quickly?

rainwoodman:
It is a quick hack to drive the code -- It perhaps can evolve into some stochastic biasing scheme (EZMock does this I believe).

I feel this is more direct than changing the seeding condition -- we can test the motion of BH alone without relying on the seeding of BH.

sbird:
Ok, fair enough.

rainwoodman:
This will conflict after neutrino branch is merged. But it will be minor.

rainwoodman:
OK. I removed the bh generation code. Will worry about it later.

sbird:
Should we merge this, or are you waiting for some test to complete on BlueTides? I am happy with the pr.

rainwoodman:
The issue with BlueTides is bigger than this PR.. We'll have to cherry-pick this anyways...

PR #137, opened 2017-10-18
Star part

### LINKED ISSUE #19: particle data structure lifting
1. We shall stop insisting sphp and p are in the same order.
2. Lifting star data from P will be useful for a chemical network / stellar evolution model (@sbird) increasing memory efficiency.
3. Then the duplicated code in domain.c can be replaced largely by loops.

I am slightly concerned the garbage collection may become slow -- There currently is a log NumPart sorting every step regardless particles are active -- but we don't need to do GC every step. maybe just each PM-step or each time we do domain, or when the lifted data pool is full.

sbird:
This pull request moves star-specific fields out of particle_data into a struct star_particle_data
I move Generation, SwallowID Metallicity and FormationTime. METALS is always on because 
now it has very small cost. I added a BirthDensity field based like the one found in EAGLE to the stars.
The save macros are extended to allow for arrays that are only found in some types. 

WINDS was removed too, because we have the WindModel = nowind parameter anyway.

Because my eyes hurt when implementing this, I also substantially cleaned up the domain exchange. It now has loops. I will probably add a unit test for it in a subsequent pull request.

I thought of making SwallowID be a temporary array, but since we want to save swallowed particles to the snapshot eventually I decided not to.

This has minimal performance impact and minimal effect on output, except that garbage is a little slower (but that needs updating anyway).

Fixes #19

### DIFF
--- a/allvars.h
+++ b/allvars.h
@@ -168,8 +168,7 @@ extern int NTask;		/*!< number of processors */
 
 extern int NumPart;		/*!< number of particles on the LOCAL processor */
 
-/* Local number of particles; this is accurate after a GC */
-extern int64_t NLocal[6];
+/* Local number of particles; this is accurate after domain */
 extern int64_t NTotal[6];
 
 /* Number of used BHP slots */
--- a/allvars.h
+++ b/allvars.h
@@ -174,6 +174,7 @@ extern int64_t NTotal[6];
 
 /* Number of used BHP slots */
 extern int N_bh_slots;
+extern int N_star_slots;
 extern int N_sph_slots;
 
 /* variables for input/output , usually only used on process 0 */
@@ -516,7 +517,8 @@ extern struct particle_data
         signed char TimeBin; /* Time step bin; -1 for unassigned.*/
     };
 
-    unsigned int PI; /* particle property index; used by BH. points to the BH property in BhP array.*/
+    unsigned int PI; /* particle property index; used by BH, SPH and STAR.
+                        points to the corresponding structure in **P array.*/
     MyIDType ID;
 
     MyFloat Vel[3];   /* particle velocity at its current time */
@@ -525,12 +527,6 @@ extern struct particle_data
     MyFloat GravPM[3];		/* particle acceleration due to long-range PM gravity force */
 
     MyFloat Potential;		/* gravitational potential. This is the total potential after gravtree+gravpm is called. */
-
-    MyFloat StarFormationTime;		/*!< formation time of star particle: needed to tell when wind is active. */
-
-#ifdef METALS
-    MyFloat Metallicity;		/*!< metallicity of gas or star particle */
-#endif				/* closes METALS */
 
     MyFloat Hsml;
 
@@ -578,6 +574,7 @@ struct bh_particle_data {
     MyFloat Entropy;
     MyFloat Pressure;
     MyFloat SurroundingGasVel[3];
+    MyFloat FormationTime;		/*!< formation time of black hole. */
 
     MyFloat accreted_Mass;
     MyFloat accreted_BHMass;
@@ -590,6 +587,13 @@ struct bh_particle_data {
     short int TimeBinLimit;
 } * BhP;
 
+/*Data for each star particle*/
+extern struct star_particle_data
+{
+    struct particle_data_ext base;
+    MyFloat FormationTime;		/*!< formation time of star particle: needed to tell when wind is active. */
+    MyFloat Metallicity;		/*!< metallicity of star particle */
+} * StarP;
 
 /* the following structure holds data that is stored for each SPH particle in addition to the collisionless
  * variables.
@@ -606,6 +610,7 @@ extern struct sph_particle_data
 #define EOMDensity Density
 #endif
 
+    MyFloat Metallicity;		/*!< metallicity of gas particle */
     MyFloat Entropy;		/*!< current value of entropy (actually entropic function) of particle */
     MyFloat MaxSignalVel;           /*!< maximum signal velocity */
     MyFloat       Density;		/*!< current baryonic mass density of particle */
@@ -638,6 +643,7 @@ extern struct sph_particle_data
 
 #define SPHP(i) SphP[P[i].PI]
 #define BHP(i) BhP[P[i].PI]
+#define STARP(i) StarP[P[i].PI]
 
 #define MPI_UINT64 MPI_UNSIGNED_LONG
 #define MPI_INT64 MPI_LONG
--- a/allvars.h
+++ b/allvars.h
@@ -514,7 +514,7 @@ extern struct particle_data
     };
 
     unsigned int PI; /* particle property index; used by BH, SPH and STAR.
-                        points to the corresponding structure in **P array.*/
+                        points to the corresponding structure in (SPH|BH|STAR)P array.*/
     MyIDType ID;
 
     MyFloat Vel[3];   /* particle velocity at its current time */
--- a/allvars.h
+++ b/allvars.h
@@ -592,6 +592,7 @@ extern struct star_particle_data
 {
     struct particle_data_ext base;
     MyFloat FormationTime;		/*!< formation time of star particle */
+    MyFloat BirthDensity;		/*!< Density of gas particle at star formation. */
     MyFloat Metallicity;		/*!< metallicity of star particle */
 } * StarP;
 
--- a/allvars.h
+++ b/allvars.h
@@ -530,11 +530,6 @@ extern struct particle_data
 
     MyFloat Hsml;
 
-#ifdef BLACK_HOLES
-    /* SwallowID is not reset in blackhole.c thus cannot be in a union */
-    MyIDType SwallowID; /* who will swallow this particle, used only in blackhole.c */
-#endif
-
     /* The peano key is a hash of the position used in the domain decomposition.
      * It is slow to generate so we store it here.*/
     peano_t Key; /* only by domain.c and forcetre.c */
@@ -584,6 +579,9 @@ struct bh_particle_data {
     MyFloat MinPotVel[3];
     MyFloat MinPot;
 
+    MyIDType SwallowID; /* Allows marking of a merging particle. Used only in blackhole.c.
+                           Set to -1 in init.c and only reinitialised if a merger takes place.*/
+
     short int TimeBinLimit;
 } * BhP;
 
@@ -626,6 +624,8 @@ extern struct sph_particle_data
                    indirectly ionization state and mean molecular weight. */
 
 #ifdef BLACK_HOLES
+    MyIDType SwallowID; /* Allows marking of a particle being eaten by a black hole. Used only in blackhole.c.
+                           Set to -1 in init.c and only reinitialised if a merger takes place.*/
     MyFloat       Injected_BH_Energy;
 #endif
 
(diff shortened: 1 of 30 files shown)


allvars.h
```diff
@@ -510,13 +506,11 @@ extern struct particle_data
...
 
-        unsigned char Generation; /* How many particles it has spawned; used to generate unique particle ID. 
-                                     may wrap around with too many SFR/BH if a feedback model goes rogue */
-
         signed char TimeBin; /* Time step bin; -1 for unassigned.*/
     };
 
-    unsigned int PI; /* particle property index; used by BH. points to the BH property in BhP array.*/
+    unsigned int PI; /* particle property index; used by BH, SPH and STAR.
+                        points to the corresponding structure in **P array.*/
```
rainwoodman: ??P. ** looked like pointer deference.

sbird: Better now?

allvars.h
```diff
@@ -571,13 +554,16 @@ struct bh_particle_data {
 
     int CountProgs;
 
+    unsigned char Generation; /* Generation of formation from an SPH particle */
```
rainwoodman: We can probably increase the width of this to int, now that it is lifted.

allvars.h
```diff
@@ -606,6 +604,10 @@ extern struct sph_particle_data
 #define EOMDensity Density
 #endif
 
+    unsigned char Generation; /*!< How many particles it has spawned; used to
+                                 generate unique particle ID. may wrap around
```
rainwoodman: It almost feels like ID shall be a union of int64 and char [8].

blackhole.c
```diff
@@ -717,17 +712,17 @@ void blackhole_make_one(int index) {
...
 
-    P[child].StarFormationTime = All.Time;
+    BHP(child).FormationTime = All.Time;
     /*Ensure that mass is conserved*/
     double BHmass = All.SeedBlackHoleMass;
     if(P[index].Mass <= All.SeedBlackHoleMass) {
         P[index].IsGarbage = 1;
         BHmass = P[index].Mass;
     }
+    BHP(child).Generation = SPHP(index).Generation;
```
rainwoodman: I still feel Generation shall be a common variable in P. It is part of book keeping in forking particles -- and forking should be OK on any particles.

garbage.c
```diff
@@ -46,8 +46,8 @@ int domain_fork_particle(int parent) {
         ActiveParticle[childactive] = child;
     }
 
-    P[parent].Generation ++;
-    uint64_t g = P[parent].Generation;
+    SPHP(parent).Generation++;
```
rainwoodman: This will crash if parent is not SPH.

rainwoodman: By keeping Generation in P forking doesn't need to be aware of particle type. I think we may be able to merge generation with ID, which could have automatically done the ID modification.

petaio.c
```diff
@@ -201,16 +247,16 @@ void petaio_read_internal(char * fname, int ic) {
         }
         sprintf(blockname, "%d/%s", ptype, IOTable.ent[i].name);
         petaio_alloc_buffer(&array, &IOTable.ent[i], NLocal[ptype]);
-        petaio_read_block(&bf, blockname, &array);
-        petaio_readout_buffer(&array, &IOTable.ent[i]);
+        if(0 == petaio_read_block(&bf, blockname, &array, IOTable.ent[i].fatal))
```
rainwoodman: .required?

petaio.c
```diff
@@ -745,19 +785,19 @@ static void register_io_blocks() {
...
+    IO_REG_TYPE(Generation,       "u1", 1, 4);
     IO_REG_WRONLY(StarFormationRate, "f4", 1, 0);
-#ifdef WINDS
-    IO_REG(StarFormationTime, "f4", 1, 4);
-#endif
-#ifdef METALS
-    IO_REG(Metallicity,       "f4", 1, 0);
-    IO_REG(Metallicity,       "f4", 1, 4);
-#endif /* METALS */
+    IO_REG_NONFATAL(BirthDensity, "f4", 1, 4);
```
rainwoodman: The handling of fatal can be improved. I think we need some overall flag to store if any block is successfully read. This shall be set during read_internal(). Then init.c can make use of that information to either die or generate some fake data to make it up. 

Also, when a block is not required, but data is provided, and corruption is detected during IO, we shall die instead of proceeding as if the block is not supplied.

init.c
```diff
@@ -80,7 +80,6 @@ void init(int RestartSnapNum)
         P[i].Ti_drift = P[i].Ti_kick = All.Ti_Current;
 
         P[i].IsGarbage = 0;
-        P[i].IsNewParticle = 0;
```
rainwoodman: I am not sure about this. The okamoto feedback model is applied only to the newly formed star particles. Without the IsNewParticle flag, how do you ensure the semantics is still correct?

allvars.h
rainwoodman: yes.

rainwoodman:
Big concerns:

1. removal of IsNewParticle may break the feedback -- the wind only applied once when the star is formed -- strength proportional to mass.

2. lifting Generation couples domain_fork_particle with Type; Generation (and Type) could have been unioned into ID.

3. Fatal (Required) and default value generation could make use of a table that indicates if a block is successfully read. The only necessary blocks are the IC blocks. All the rest are non-required. We can probably refactor the special handling of IC with this table -- if a block doesn't exist, we'll run the routines in init.c to initialize it with fake data, and print a message. Also shall always die if corruption of data/container is detected, regardless of 'fatal/require'.

petaio.c
sbird: ok

petaio.c
sbird: How do we detect corruption? Through the bigfile error code?

sbird:
1. IsNewParticle is removed, but it just checks STARP.FormationTime == All.Time instead, which should have the same effect. I can bring it back though, it is not a big deal.
2. All callers of domain_fork_particle already check that Type = 0, so it is ok in practice. 
I think of "Generation" as meaning "this is how many star/BH particles an SPH particle has spawned", which is a useful thing to save for SPH particles. Both Generation and Type could be unioned with ID, indeed, but that makes the code less clear. Why not use:
domain_fork_particle(i, int generation)
and then let the caller figure it out (using SPHP(i).Generation for now), and if we start forking stars or BHs it is easy to implement?

sbird:
For 3: do you mean provide an extra function like property_getter and property_setter that sets up the array if not read? And call them all from init? So something like (pseudocode):
foreach(p in Particles) {
foreach(block in Blocks) {
if(block.not_read)
     block.init(p)
}
}

Alternatively we could have an extra function that gets called in the loop in petaio.c once a read call fails.

rainwoodman:
1. FormationTime == All.Time doesn't work, because when the precision of two numbers mismatch they never equal. I was bitten by this once. 

2. I thought of Generation as a way to keep track of 'generic' particle forking. You see that now you have it in all particle types except DM already.. And who knows if we want to start spawning tracer particles on DM at some point later...? At that point the logic for figuring out the ID/Generation has to be refactored out from SPH again.

3. No. I mean read_internal shall have a table as an out parameter. I think having an init function is an overkill -- and the init procedure of some data is not strictly a loop. 

Here is a doodling of the interface I am thinking:

```
int petaio_read_snapshot(...., IOStatus * status);
int petaio_is_provided(IOStatus * status, char * blockname, int ptype);
```

in init.c, we can do
```
if (petaio_is_provided("SmoothingLength", 0)) {
    setup_smoothing_length();
}
...
```

petaio.c
rainwoodman: Yes. If the file opens (exists) but the read fails, then it is a corruption.

rainwoodman:
3) almost feels like another PR. -- 'IO clean up'. 

I came up with the getter and setter in a deep night and I have this sense that it is not the most clean thing to have -- It just feels like perhaps we can shrinking down the use of macro magics.. BigFile handles the casting of types already, but we have to do a permutation for the slots; only very few variables must be converted before saving.

rainwoodman:
```
typedef struct {
char name; int ptype; /* this is the primary key of the struct; shall add interface to look it up.*/

io_scatter_func scatter_func;
io_gather_func gather_func;
io_init_func init_func;

struct {
ptrdiff_t offset;
size_t elsize; /* length of an element*/
void * slots; /* if NULL, write to the slots */
size_t slot_strides; /* strides of the slots to get slots[PI] */
} generic;

} IOBlockSpec;

typedef int (* io_scatter_func)(const void * buffer, IOBlock * spec);
typedef int (* io_gather_func)(void * buffer, IOBlock * spec);
typedef int (* io_init_func)(IOBlock * spec);

The default scatter and gather func will use `generic` interface to directly copy.
Then we can have special ones for SFR etc.
```

petaio.c
sbird: Not so - the open succeeds even if no file exists. It just creates an empty directory and it is the read that returns an error.

sbird:
1. Ah. Ok, in that case, let me drop the IsNewParticle change.
2. The problem I was really trying to solve was carrying around all these Generation arrays in the snapshots for pure DM simulations (when they are always zero). How about I put Generation back in P, but only write/read it for type 0,4 and 5 particles?
3. Ah, I see. That sounds indeed like a good interface. But I think I agree that should be another PR.

sbird:
Moved IO cleanup to Issue #138

rainwoodman:
For 2; The bottom is that Generation and Type + lower bits of ID == Primary Key of particles. But we are now tracking them separately. If we merge Generation as higher bits of ID (primary key) then there is no need to keep another array at all.

sbird:
But then we would have to change the ID of the SPH particle every time it spawns a star, whereas now IDs are persistent. Do we really want to change that?

rainwoodman:
OK. This is a place whatever we do it is going to be suboptimal. What about going your second proposal, Generation in P but IO only does it for those that have changed. Then if we enable DM forking we can add that block to IO.

sbird:
Ok! That is what is currently implemented.

petaio.c
rainwoodman: That's a bigfile bug. I don't think it is supposed to ever do that...

rainwoodman: big_block_open will try to open 'header' file -- and it if doesn't exist it should throw an error. (line 315 of bigfile.c.)  big_file_open do work even if there is no dataset inside the file, but it is not the function you are calling here..

sfr_eff.c
```diff
@@ -715,31 +702,29 @@ static int make_particle_star(int i) {
...
+    STARP(newstar).base.ID = P[newstar].ID;
+    /* set ptype */
+    P[newstar].Type = 4;
+    /*Set properties*/
+    sum_mass_stars += P[newstar].Mass;
+    STARP(newstar).FormationTime = All.Time;
+    STARP(newstar).BirthDensity = SPHP(i).Density;
+    /*Copy metallicity*/
+    STARP(newstar).Metallicity = SPHP(i).Metallicity;
 #ifdef WINDS
```
rainwoodman: Straying ifdef WINDS.

sbird: Bah, rebase errors...

petaio.c
sbird: I checked and indeed you are right  - I think I must have had a left-over folder from some earlier crashing run before. I will fix the error case.

rainwoodman: This is the only remaining part of this PR.

sbird: Current head of this branch only returns if the open fails, and does endrun if read or seek fail. So I think it is good.

PR #139, opened 2017-10-20
Darkenergy

sbird:
Add parameters for a simple (w0, wa) dark energy parametrisation, so we can run simulations with dark energy models of the type most commonly used by cosmologists.

### DIFF
--- a/cosmology.c
+++ b/cosmology.c
@@ -10,6 +10,7 @@
 #define  STEFAN_BOLTZMANN 5.670373e-5
 
 static Cosmology * CP = NULL;
+static inline double OmegaFLD(const double a);
 
 void init_cosmology(Cosmology * CP_in)
 {
@@ -53,7 +54,8 @@ double hubble_function(double a)
 
     /* first do the terms in SQRT */
     hubble_a = CP->OmegaLambda;
-
+    
+    hubble_a += OmegaFLD(a);
     hubble_a += CP->OmegaK / (a * a);
     hubble_a += CP->Omega0 / (a * a * a);
 
@@ -130,6 +132,16 @@ double F_Omega(double a)
     double dD1da=0;
     double D1 = growth(a, &dD1da);
     return a / D1 * dD1da;
+}
+
+/*Dark energy density as a function of time:
+ * OmegaFLD(a)  ~ exp(-3 int((1+w(a))/a da)_a^1
+ * and w(a) = w0 + (1-a) wa*/
+static inline double OmegaFLD(const double a)
+{
+    if(CP->Omega_fld == 0.)
+        return 0;
+    return CP->Omega_fld * pow(a, 3 * (1 + CP->w0_fld + CP->wa_fld))*exp(3*CP->wa_fld*(1-a));
 }
 
 static double sigma2_int(double k, void * p)
--- a/genic/params.c
+++ b/genic/params.c
@@ -23,10 +23,10 @@ create_parameters()
     param_declare_string(ps, "OutputDir", REQUIRED, 0, "");
     param_declare_string(ps, "FileBase", REQUIRED, 0, "");
 
-    param_declare_double(ps, "Omega0", REQUIRED, 0, "");
-    param_declare_double(ps, "OmegaLambda", REQUIRED, 0, "");
-    param_declare_double(ps, "OmegaBaryon", REQUIRED, 0, "");
-    param_declare_double(ps, "HubbleParam", REQUIRED, 0.7, "Hubble parameter");
+    param_declare_double(ps, "Omega0", REQUIRED, 0.2814, "");
+    param_declare_double(ps, "OmegaBaryon", REQUIRED, 0.0464, "");
+    param_declare_double(ps, "OmegaLambda", REQUIRED, 0.7186, "Dark energy density at z=0");
+    param_declare_double(ps, "HubbleParam", REQUIRED, 0.697, "Hubble parameter");
     param_declare_int(ps,    "ProduceGas", REQUIRED, 0, "Should we create baryon particles?");
     param_declare_double(ps, "BoxSize", REQUIRED, 0, "");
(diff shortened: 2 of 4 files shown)


cosmology.c
```diff
@@ -43,6 +44,9 @@ void init_cosmology(Cosmology * CP_in)
      * rho_nu/rho_g = 7/8 (T_nu/T_cmb)^4 *N_eff,
      * but we absorbed N_eff into T_nu above. */
     CP->OmegaNu0 = CP->OmegaG * 7. / 8 * pow(TNu0_TCMB0, 4) * 3;
+
+    /*Check whether the dark energy sector is a cosmological constant.*/
+    CP->EvolvingDEOn = (fabs(CP->w0 + 1) < 1e-7) && (fabs(CP->wa) < 1e-7);
```
rainwoodman: Is this also called the Dark Energy Fluid in CLASS? Perhaps call it that? When Fluid DE is on, OmegaLambda shall be usually set to zero for a 'reasonable' cosmology -- may want to catch that and issue a warning.

I think Class called the parameters just w0_fld and wa_fld -- even if you'd avoid Fluid, DarkEnergy is always more readable.

cosmology.c
```diff
@@ -52,7 +56,7 @@ double hubble_function(double a)
     double hubble_a;
 
     /* first do the terms in SQRT */
-    hubble_a = CP->OmegaLambda;
+    hubble_a = OmegaDE(a);
```
rainwoodman: Omega_DEFluid(a) + OmegaLambda?

cosmology.c
```diff
@@ -132,6 +136,16 @@ double F_Omega(double a)
...
 }
 
+/*Dark energy density as a function of time:
+ * Omega_DE(a)  ~ exp(-3 int((1+w(a))/a da)_a^1
+ * and w(a) = w0 + (1-a) wa*/
+static inline double OmegaDE(const double a)
+{
+    if(!CP->EvolvingDEOn)
+        return CP->OmegaLambda;
+    return CP->OmegaLambda * pow(a, 3 * (1 + CP->w0 + CP->wa))*exp(3*CP->wa*(1-a));
```
rainwoodman: OmegaFLD * pow(a, ....) ?

cosmology.c
rainwoodman: The idea there is you can have a Lambda-like DE and a Fluid like DE component, and they can coexist.

cosmology.c
rainwoodman: When OmegaFluid is set to zero the component is turned off.

cosmology.c
sbird: Not sure - I was following CAMB and PDG which use definitions more like the ones I did here. class also uses a different definition of wa than I have: w = w0 + wa (1-a0/a), whereas I have w = w0 + wa(1-a). Let me check with a DE cosmologist which one they expect.

sbird:
Yes - I have talked to a DE cosmologist and it seems that following CLASS is the right thing to do. Let's do it!

cosmology.c
sbird: (this is a CLASS documentation typo)

sbird:
There we go. Now interface matches that of CLASS.

cosmology.c
```diff
@@ -132,6 +134,16 @@ double F_Omega(double a)
     return a / D1 * dD1da;
 }
 
+/*Dark energy density as a function of time:
+ * Omega_DE(a)  ~ exp(-3 int((1+w(a))/a da)_a^1
```
rainwoodman: Omega_FLD

sbird: fixed

genic/params.c
```diff
@@ -87,6 +90,11 @@ void read_parameterfile(char *fname)
     CP.OmegaLambda = param_get_double(ps, "OmegaLambda");
     CP.OmegaBaryon = param_get_double(ps, "OmegaBaryon");
     CP.HubbleParam = param_get_double(ps, "HubbleParam");
+    CP.Omega_fld = param_get_double(ps, "Omega_fld");
+    CP.w0_fld = param_get_double(ps,"w0_fld");
+    CP.wa_fld = param_get_double(ps,"wa_fld");
+    if(CP.OmegaLambda > 0 && CP.Omega_fld > 0)
+        endrun(0, "Cannot have OmegaLambda and Omega_fld (evolving dark energy) at the same time!\n");
```
rainwoodman: What about "Having OmegaLambda and Omega_fld the same time is highly unusual." Because the code runs just fine.

param.c
```diff
@@ -354,6 +357,11 @@ void read_parameter_file(char *fname)
         All.CP.Omega0 = param_get_double(ps, "Omega0");
         All.CP.OmegaBaryon = param_get_double(ps, "OmegaBaryon");
         All.CP.OmegaLambda = param_get_double(ps, "OmegaLambda");
+        All.CP.Omega_fld = param_get_double(ps, "Omega_fld");
+        if(All.CP.OmegaLambda > 0 && All.CP.Omega_fld > 0)
```
rainwoodman: We shall look into sharing the cosmology construction code in GenIC and the main code..

sbird: One thing we could do is to specify the cosmology in GenIC, store it in the header, and then just read it in the main code.

genic/params.c
sbird: That works too, but doesn't match what CLASS does: class errors. OmegaLambda is really just a special case of Omega_fld so there is no *reason* to have them both at the same time, unless you accidentally used the wrong value in the parameter file.

rainwoodman: OK then.

param.c
rainwoodman: It just creates another parser -- from bigfile attr to cosmology..

rainwoodman:
I think this is good to be merged.

sbird:
thanks

PR #140, opened 2017-10-22
Multitransfer

sbird:
Implement using different transfer functions for CDM and baryons into MP-GenIC, with test (ported from S-GenIC). This is a prerequisite for the IC generation part of massive neutrinos, as well as enabling AdaptiveGravsoft by default.


sbird:
The travis test is rather broken at the moment. It should agree to better than 5%. One reason is that the Planck 2015 cosmology has a non-zero neutrino mass of 0.06 eV, which induces scale-dependent growth not accounted for in the growth function rescaling.

sbird:
The (small) change in output power comes about because of the change in the way ProduceGas is handled.

sbird:
Yet the discrepancy persists and is large even with WMAP and no ProduceGas. Odd.

sbird:
This appears to be a problem with the power spectrum estimation in NBodyKit. See here:
powercompare
This is a plot of the power spectrum estimated from the same IC box using the same grid size in NBodyKit and my own personal power spectrum estimator, GenPK. You can see that on small scales NBodyKit loses a lot of power.

Is it perhaps using nearest grid point interpolation, or forgetting to divide by the CIC window function?

In any case, despite the test failure GenIC seems fine.

examples/travis/check-results.py
```diff
@@ -47,15 +50,25 @@ def test_power(output, ref, ref_z):
...
-pklin = LinearPower(Planck15, redshift=ref_z)
+pklin = LinearPower(WMAP9, redshift=ref_z)
+#This checks that the power spectrum loading and rescaling code is working.
+genpk = numpy.loadtxt("output/inputspec_IC.txt")
+assert_allclose(pklin(genpk[:,0]), genpk[:,1], rtol=5e-4, atol=0.0)
+#This checks that the output is working
 print(ref_z, ref.power['power'][2:8])
 print(pklin(ref.power['k'][2:8]))
-assert_allclose(abs(ref.power['power'])[2:8],
+#Deconvolve by the CIC window. This is not exact.
```
rainwoodman: There is no need to deconvolve CIC for the measurements -- nbodykit does its own corrections. 
Were you correcting for the CIC in ZA?

genic/power.c
```diff
@@ -72,113 +116,93 @@ void read_power_table(void)
...
+int initialize_powerspectrum(int ThisTask, double InitTime, double UnitLength_in_cm_in, Cosmology * CPin, struct power_params * ppar)
 {
-    if(WhichSpectrum == 2)
-        read_power_table();
+    if(ppar->WhichSpectrum == 2) {
+        read_power_table(ThisTask, ppar->FileWithInputSpectrum, 1, &power_table, ppar->SpectrumLengthScale, parse_power);
+        transfer_table.Nentry = 0;
+        if(ppar->DifferentTransferFunctions)
+            read_power_table(ThisTask, ppar->FileWithTransferFunction, MAXCOLS, &transfer_table, ppar->SpectrumLengthScale, parse_transfer);
+        message(0, "Power spectrum rows: %d, Transfer: %d\n", power_table.Nentry, transfer_table.Nentry);
```
rainwoodman: Is it more convenient  to have one multi column file or is it more convinient to have several files?

genic/power.c
```diff
@@ -250,86 +274,8 @@ double sigma2_int(double k, void * params)
       w = 1./3. - kr2/30. +kr2*kr2/840.;
   else
       w = 3 * (sin(kr) / kr - cos(kr)) / kr2;
-  x = 4 * M_PI / (2 * M_PI * 2 * M_PI * 2 * M_PI) * k * k * w * w * PowerSpec(k);
+  x = 4 * M_PI / (2 * M_PI * 2 * M_PI * 2 * M_PI) * k * k * w * w * PowerSpec(k, 7);
```
rainwoodman: a type of -1 or perhaps using a bitmask might be cleaner than 7.

tests/test_genic-power.c
```diff
@@ -0,0 +1,93 @@
...
+static void
+test_read_rescale_sigma8(void ** state)
+{
+    /*Do setup*/
+    struct power_params PowerP;
+    /* Test rescaling to an earlier time
+     * (we still use the same z=99 power which should not be rescaled in a real simulation)*/
+    PowerP.InputPowerRedshift = 0;
+    PowerP.Sigma8 = -1;
+    PowerP.FileWithInputSpectrum = "examples/camb_matterpow_99.dat";
```
rainwoodman: might want to use a path converted from  __FILE__ -- then you don't have to run the test from a specific directory.

genic/power.c
sbird: CAMB produces one multi-column file. Or do you mean the internal structure in the code?
One structure is more convenient because the transfer functions have the same k (but the matter power does not) and are normalised the same.

genic/params.c
```diff
@@ -39,12 +39,14 @@ create_parameters()
     param_declare_double(ps, "w0_fld", OPTIONAL, -1., "Dark energy equation of state");
     param_declare_double(ps, "wa_fld", OPTIONAL, 0, "Dark energy evolution parameter");
 
+    param_declare_int(ps, "DifferentTransferFunctions", OPTIONAL, 0, "Use species specific transfer functions for baryon and CDM.");
```
rainwoodman: For the `MultipleTransferFunction` case, does it make more sense to have an abstraction of the primordial power rather than saving the transfer ratios? Because physically that's how the power spectrum of each species come about.

genic/power.c
sbird: Traditionally I have used 7 to mean "total power" in transfer function code, but I can make it -1 if you prefer.

tests/test_genic-power.c
sbird: I did not realise FILE was not a C++ ism! Thanks.

genic/params.c
sbird: It really makes more sense to save the transfer ratios, because the interpolation has to be able to deal with different k values for power and transfer functions.

examples/travis/check-results.py
sbird: Possibly. I was just following what GenPK does, which I worked it out many years ago and have since forgotten. It uses sinc^{-2} for each k dimension, so sinc^{-6} in total, applied to delta. So for P(k) it should be sinc^{-12}. Looks like one factor of sinc is missing here.

tests/test_genic-power.c
sbird: On my system FILE gives the path of the current source file relative to the directory the compiler is invoked in, and so is unfortunately no better than just hard-coding "examples"

examples/travis/check-results.py
sbird: I looked through the nbodykit code but got a little lost in the python - it looks like the right factor of sinc is applied, but in real space before the fourier transform, which I don't understand fully. Since the factor applied is the same I also do not understand why it differs from GenPK unless there is a bug somewhere.

rainwoodman: I think you just need to look at this figure: http://nbodykit.readthedocs.io/en/latest/cookbook/interlacing.html ; nbodykit uses is interlacing and TSC -- you can think of the nbodykit power spectrum estimator as 'unbiased' at all scales.


The CIC window in GenIC does 'smooth-out' ZA; but since the displacment invokes non-linear coupling, I don't see why you'll need a correction -- did your correction improve agreement against linear power?  ZA is not even supposed to be the same as the linear power -- at k~0.05 it is already 2% difference even in the limit where the CIC window is very small.

genic/power.c
rainwoodman: I see. the idea is to read the output of CAMB directly.

tests/test_genic-power.c
rainwoodman: I guess we can do some magic with config.h. Let me add something to the master.

rainwoodman: If you rebase, there is GADGET_TESTDATA_ROOT which points to the source code root. The tests can also directly include header in the source root (no need to use '../')

genic/params.c
rainwoodman: Aren't these just functions that return value of any give k -- I tend to wipe the data representation and interpolation algorithm under the carpet.You will need to interpolate both functions anyways because they don't match the mesh-k. But I think we can worry about these things later when the mood is right..

rainwoodman:
I looked up and the default setting in nbodykit does a CIC with compensation -- I somehow thought the default is the recommended setting in the notebook.

examples/travis/check-results.py
rainwoodman: There might be a factor of 2 in your sinc kernel -- did you check your GenPK give a flat power back with a uniform field? The SINC correction is supposed to give you a flat power on shot-noise like input, see here:
[mozilla.pdf](https://github.com/rainwoodman/MP-Gadget/files/1406083/mozilla.pdf)

You can also see that the TSC+interlacing gives a slight up-turn for pure shotnoise, because it is supposed work better on LCDM-like power(not dominated by shotnoise at high-k). The default (CIC+compensation) in nbodykit is a good estimator when Nmesh is so great that the field is already shot-noise. I somehow misremembered that the default is the TSC+interlacing setup.

rainwoodman: Also the compensation is applied in fourier space, as declared here: https://github.com/bccp/nbodykit/blob/d554caf8b6db15b1d7e71fa9a684dae7834e814b/nbodykit/base/catalogmesh.py#L502

sbird: Right - so it looks like NBodykit and GenPK are doing the same thing. I double-checked that nbodykit gives exactly the same answer with tsc once the window function is corrected for (impressively so!). The window function in the linked document is the one implemented in GenPK. GenPK also gives a totally flat spectrum for white noise down to the Nyquist frequency of the particles. Also *both* codes give results that are invariant to the grid size. So then I wonder why they differ from each other.

sbird:
If I compute the DM power individually, or I compute the gas power individually, both agree basically exactly with GenPK. It is only when I use a MultipleSpeciesCatalog that I get disagreement. Something is wrong!

rainwoodman:
I did find a bug in MultipleSpeciesCatalog when using interlaced painting -- but sounds like you are not using the interlaced painter?

sbird:
I have no idea. But it definitely looks like a bug in MultipleSpeciesCatalog. I think we want to check the power of different species individually anyway.

sbird:
New commit fixes it.

sbird:
Rebased. Boom. Let's merge this sucker.

rainwoodman:
Yeah. Travis is green. Merging!

PR #142, opened 2017-10-24
Allow merging deeperly diverged subtrees in domain.

### LINKED ISSUE #141: Code crashing on three different machines
For the last few days, my standard test run (hydro run, L=20 Mpc/h, N=256, z_ini=99) is crashing. It happens in Hypatia (I attach relevant files to reproduce the crash), but similar smaller runs also crash on my desktop, and @andersdot is having also crashes at CCA. 

It would be great to have stable versions that we can use for larger runs. Do you have a suggestion?

[crash_mpgadget.tar.gz](https://github.com/rainwoodman/MP-Gadget/files/1411084/crash_mpgadget.tar.gz)

The actual error message here starts with: 

[compute-0-24:08864] *** Process received signal ***
[compute-0-24:08864] Signal: Floating point exception (8)
[compute-0-24:08864] Signal code: Integer divide-by-zero (1)
[compute-0-24:08864] Failing at address: 0x41eb1e
[compute-0-24:08864] [ 0] /lib64/libpthread.so.0() [0x35f7e0f7e0]

and it consistently happens at z=2.63687 (I ran three times the code over the last few days, using slightly updated versions of the code, and it always crashed at this z).

rainwoodman:
print a warning when they disagree too strongly with each other.

Intended to fix #141

### DIFF
--- a/domain.c
+++ b/domain.c
@@ -1364,20 +1364,27 @@ domain_toptree_merge(struct local_topnode_data *treeA,
     else if(treeB[noB].Shift > treeA[noA].Shift)
     {
         /* Since we only know how to split A, here we simply add a spatial average to A */
-        int n = 1L << (treeB[noB].Shift - treeA[noA].Shift);
+        uint64_t n = 1L << (treeB[noB].Shift - treeA[noA].Shift);
+
+        if(treeB[noB].Shift - treeA[noA].Shift > 60) {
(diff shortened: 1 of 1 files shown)


rainwoodman:
I am merging this but keeping the branch on in case we'll want a separate test.

Issue #123, opened 2017-10-04
Softening cleanup


sbird:
We will have three parameters:
1 - Collisionless softening (which is comoving) and set as a fraction of mean particle separation.
2 - We will have a collisional softening, which is adaptive if zero, and otherwise a fraction of mean DM separation. Probably adaptive is the default.
3 - Parameter giving Ngrid (optional) for multi-grid simulations. If 0, compute from DM type 1 totnumpart, otherwise use it.

Everything is always comoving.

andreufont:
Hi @sbird, 

Somewhat related to this. It would be great to have a bit of documentation / recommendations in the configuration files about what values to use. Right now in my sims I'm using the settings that you sent us at some point (see below), but I have no idea whether I should be scaling these with box size, mean particle separation, or whether these should be thought as universal. 

SofteningGas = 2.60417333333
SofteningGasMaxPhys = 2.60417333333
SofteningHalo = 2.60417333333
SofteningHaloMaxPhys = 2.60417333333
SofteningDisk = 2.60417333333
SofteningDiskMaxPhys = 2.60417333333
SofteningBulge = 2.60417333333
SofteningBulgeMaxPhys = 2.60417333333
SofteningStars = 2.60417333333
SofteningStarsMaxPhys = 2.60417333333
SofteningBndry = 2.60417333333
SofteningBndryMaxPhys = 2.60417333333

sbird:
You want all those settings to be 1/30 of the mean interparticle spacing.

andreufont:
Great. Should this be set as a default? As in: if you don't provide any value for these, they are automatically set to 1/30 times mean particle spacing.

sbird:
Yes - they should also just be one parameter. if you feel like implementing the plan described in this issue that would be great :)

rainwoodman:
I'll take care of this while setting up the NERSC CI.

PR #143, opened 2017-10-24
Soften cleanup

rainwoodman:
This largely removes the twisted Gadget logic in the handling of softening length by

1. decouple type from the softening, allowing a more natural flow of adaptive softening of gas
2. reparametrize the softening to use two parameters; eliminating the non-sense Phys softening limiter -- we argued using adaptive softening made more sense.
3. Use a single array for softenings, not two off only by a factor of 2.8.

Also fixed a potential divide by zero bug of potential calculation when adaptive softening is enabled.

### DIFF
--- a/param.c
+++ b/param.c
@@ -196,7 +196,7 @@ create_gadget_parameter_set()
     param_declare_int(ps, "FastParticleType", OPTIONAL, 2, "Particles of this type will not decrease the timestep. Default neutrinos.");
 
     param_declare_double(ps, "GravitySoftening", OPTIONAL, 1./30., "Softening for collisionless particles; units of mean separation of DM. ForceSoftening is 2.8 times this.");
-    param_declare_double(ps, "GravitySofteningGas", OPTIONAL, 1./30., "Softening for collisionless particles; units of mean separation of DM; 0 to use Hsml of last step. ");
+    param_declare_double(ps, "GravitySofteningGas", OPTIONAL, 1./30., "Softening for collisional particles (Gas); units of mean separation of DM; 0 to use Hsml of last step. ");
 
     param_declare_double(ps, "BufferSize", OPTIONAL, 100, "");
     param_declare_double(ps, "PartAllocFactor", REQUIRED, 0, "");
--- a/param.c
+++ b/param.c
@@ -195,8 +195,8 @@ create_gadget_parameter_set()
     param_declare_int(ps, "RadiationOn", OPTIONAL, 0, "Include radiation density in the background evolution.");
     param_declare_int(ps, "FastParticleType", OPTIONAL, 2, "Particles of this type will not decrease the timestep. Default neutrinos.");
 
-    param_declare_double(ps, "GravitySoftening", OPTIONAL, 0.05, "Gravitational softening for collisionless particles; units of mean separation of DM.");
-    param_declare_double(ps, "GravitySofteningGas", OPTIONAL, 0.05, "Gravitational softening for collisionless particles; units of mean separation of DM; 0 for adaptive ");
+    param_declare_double(ps, "GravitySoftening", OPTIONAL, 1./30., "Softening for collisionless particles; units of mean separation of DM. ForceSoftening is 2.8 times this.");
+    param_declare_double(ps, "GravitySofteningGas", OPTIONAL, 1./30., "Softening for collisionless particles; units of mean separation of DM; 0 to use Hsml of last step. ");
 
     param_declare_double(ps, "BufferSize", OPTIONAL, 100, "");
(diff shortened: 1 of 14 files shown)


rainwoodman:
This PR replaces #123

param.c
```diff
@@ -195,19 +195,8 @@ create_gadget_parameter_set()
...
-    param_declare_double(ps, "SofteningStars", REQUIRED, 0, "");
-    param_declare_double(ps, "SofteningBndry", REQUIRED, 0, "");
-    param_declare_double(ps, "SofteningHaloMaxPhys", REQUIRED, 0, "");
-    param_declare_double(ps, "SofteningDiskMaxPhys", REQUIRED, 0, "");
-    param_declare_double(ps, "SofteningBulgeMaxPhys", REQUIRED, 0, "");
-    param_declare_double(ps, "SofteningGasMaxPhys", REQUIRED, 0, "");
-    param_declare_double(ps, "SofteningStarsMaxPhys", REQUIRED, 0, "");
-    param_declare_double(ps, "SofteningBndryMaxPhys", REQUIRED, 0, "");
+    param_declare_double(ps, "GravitySoftening", OPTIONAL, 1./30., "Softening for collisionless particles; units of mean separation of DM. ForceSoftening is 2.8 times this.");
+    param_declare_double(ps, "GravitySofteningGas", OPTIONAL, 1./30., "Softening for collisionless particles; units of mean separation of DM; 0 to use Hsml of last step. ");
```
sbird: "Softening for collisional particles"

timestep.c
```diff
@@ -91,40 +91,12 @@ set_softenings(const double time)
 {
```
sbird: Only need to call this function at init(), not in set_global_time()

timestep.c
```diff
@@ -91,40 +91,12 @@ set_softenings(const double time)
...
-    for(i = 0; i < 6; i++)
-        All.ForceSoftening[i] = 2.8 * All.SofteningTable[i];
-
-    All.MinGasHsml = All.MinGasHsmlFractional * All.ForceSoftening[0];
+    for(i = 0; i < 6; i ++)
+        All.ForceSoftening[i] = 2.8 * All.GravitySoftening * All.MeanSeparation[1];
+
+    All.ForceSoftening[0] = 2.8 * All.GravitySofteningGas * All.MeanSeparation[1];
+
+    All.MinGasHsml = All.MinGasHsmlFractional * All.ForceSoftening[1];
```
sbird: Maybe better to do:
All.MinGasHsml = All.MinGasHsmlFractional * All.GravitySoftening * All.MeanSeparation[0]?

sbird:
Actually, I wonder if we are really correct to not use the physical units softening parameters. There may be something we are missing. I will ask someone.

timestep.c
rainwoodman: It depends on the definition; I changed it to be a fraction of DM softening -- so DM Softening sort of acts like the master control block of all 'softening' related parameters.

rainwoodman:
I remember you said the adaptive softening mimics softening in proper units. But we agreed that for DM there is no Hsml -- proper unit softening acted as a poor man's version of adaptive.

rainwoodman:
But it goes the wrong direction (proper units gives a bigger smoothing per mean separation at later time).

sbird:
I did say that but I am worried that I said something untrue or missed something. Whatever, I don't use proper softening myself, and it is easy enough to bring back if we want to.

sbird:
Fixed comoving softening gives a larger proper softening at late times. So the purpose of fixed proper softening is to prevent the proper softening from increasing in large halos, which have constant proper size with redshift after virialising.

sbird:
But the adaptive softening should have the same effect (except better): as the halo becomes more dense in comoving units the adaptive softening will shrink. Since on these scales the dynamics are dominated by gas this should have the desired effect, but be somewhat more rigourous.

sbird:
The DM softening will continue to increase, but we argued that this was ok, especially as it increases by a factor of 2-3 only.

timestep.c
sbird: I was worried about odd setups where there are more or less gas particles than DM particles. But since we set this parameter to zero anyway by default I think what you have is fine.

rainwoodman:
If a child node has MixedGravSmoothingLength, then probably all of its parents shall have the flag too? Currently I don't think the code guarentees this -- if the child's MaxSmoothing is the same as the parent the flag is not set -- it doesn't sound right...

sbird:
Yes, I think that's true. Parents should have MixedGravSoft = OR(children MixedGravSoft). The old code I think did guarantee this.

rainwoodman:
I found a central place to do this -- does this look reasonable?

sbird:
Makes sense to me.

PR #144, opened 2017-10-25
Less bh param

rainwoodman:


sbird:
The only one I am concerned about is WindModel, as it implicitly includes "nowind": Everything else has a "XXXOn" parameter which is required.

rainwoodman:
The twist is that wind is part of SFR. nowind is a special type of wind feedback. We can also think of wind as a separate module that depends on SFR -- but it introduces a module dependency and I think it is not easy to handle..

rainwoodman:
I think we shall make the WindModel default sh03 -- which is the canonical model.

sbird:
Ok, fair enough.

I would make the default what is in BlueTides and Illustris (which I think is what you have?). sh03 is canonical but a bad fit to reality.

rainwoodman:
I read the code again and agree with adding WindOn. The default shall be something like ojft10 then, and the criteria shall have H2..

sbird:
Sounds good to me.

sbird:
I am happy for this to be merged.

rainwoodman:
Sure. I'm waiting for travis to turn green.

PR #146, opened 2017-10-25
updated lya config files after changes in softening and winds

andreufont:
I modified the configuration files in examples/lya and examples/dm-only to work well with the new softening settings and WindOn option.

Also, I commented out the BlackHole options from the configuration files, although I'm not sure this is the right thing to do... Do you expect most people to compile the code with the BlackHole option? 

PS: this is also my first attempt to merge from a fork, I hope I don't break the whole internet.

### DIFF
--- a/examples/dm-only/paramfile.gadget
+++ b/examples/dm-only/paramfile.gadget
@@ -25,8 +25,6 @@ StarformationCriterion = density
 RadiationOn = 0
 HydroOn = 0
 WindOn = 0
-# add this line if MP-Gadget was compiled with BlackHole option
-#BlackHoleOn = 0
 
 # Accuracy of time integration
 MaxSizeTimestep = 0.1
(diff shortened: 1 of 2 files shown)


examples/lya/paramfile.gadget
```diff
@@ -63,9 +52,9 @@ QuickLymanAlphaProbability = 1 # Set to 1.0 to turn dense gas directly into star
 
 SnapshotWithFOF = 1
 WindModel = nowind
-BlackHoleOn = 0
-BlackHoleFeedbackFactor = 0.05
-BlackHoleFeedbackRadius = 0.
-BlackHoleFeedbackRadiusMaxPhys = 0.
-BlackHoleFeedbackMethod = spline | mass
-SeedBlackHoleMass = 5.0e-5
+# if compiled with BlackHole option, add these lines
```
rainwoodman: I suppose these blackhole stuff are all optional now too... Do you want to get rid of them while you are still on it?

andreufont:
I've added a new commit where I remove the BlackHole options alltogether.

examples/dm-only/paramfile.gadget
```diff
@@ -29,7 +24,7 @@ StarformationOn = 0
 StarformationCriterion = density
 RadiationOn = 0
 HydroOn = 0
-BlackHoleOn = 0
```
sbird: I think BlackHoleOn is still required, isn't it? Otherwise looks good.

sbird:
I want to merge this so I can rebase the neutrino branch. I will restore the BlackHoleOn parameters afterwards.

sbird:
I see that BlackHoleOn is required only when BLACK_HOLES is enabled: we should probably change the code so that the same parameters are needed irrespective of the compile-time options (they would just do nothing if the compile option was not enabled). We can do that later though.

rainwoodman:
Yes. I'll merge this one now.

andreufont:
Yes, as Simeon says BlackHoleOn = 0 is required if you compile the code with BlackHole option, but if you did not, then it would crash if you have the option. It would be good to modify the code so that it doesn't crash in any case. I'll start an issue.

sbird:
No worries, I'll fix it now.

PR #148, opened 2017-10-25
Make BLACK_HOLE parameters exist even if compile flag is off.

sbird:
Make the code accept the same parameters irrespective of the SFR and
BLACK_HOLES compile flags. If these compile flags are not enabled the
(optional) black hole and star parameters just do nothing.

    Fixes: #147

Issue #147, opened 2017-10-25
Unify parameter files regardless of compilation options


andreufont:
Right now, if you compile MP-Gadget with the flag:
OPT += -DBLACK_HOLES
you will need to specify in paramfile.gadget the option:
BlackHoleOn = 0

However, if you compile MP-Gadget without the flag, the code will crash if you have BlackHoleOn defined in your parameter file. 

Ideally, if you have BlackHoleOn = 0 on your parameter file, but you did not compile with DBLACK_HOLES, it would just ignore it.

rainwoodman:
Ideally the parser shall ignore module specific parameters when a flag is set to disabled. Currently we are making all module specific parameters 'optional' to make it 'apparently working'. So if any REQUIRED parameter leaks into param.c it will be broken again.

Let's keep this Open till we can actually handle the dependency of parameter existence.

sbird:
Ok, I'm closing this since we don't have any compile flags anyway anymore. We still don't have module specific parameters but I don't think we're going to.

PR #150, opened 2017-10-26
hide BH options when not compiled with BH flag

andreufont:
This PR is related to PR #148 and issue #147 .

It hides BH options when MP-Gadget was compiled without the BH flag. 

I've tested that the code compiles both when the flag is ON and OFF.

### DIFF
--- a/param.c
+++ b/param.c
@@ -224,6 +224,7 @@ create_gadget_parameter_set()
 
     param_declare_double(ps, "BlackHoleFeedbackRadiusMaxPhys", OPTIONAL, 0, "");
 
+#ifdef BLACK_HOLES
     static ParameterEnum BlackHoleFeedbackMethodEnum [] = {
         {"mass", BH_FEEDBACK_MASS},
         {"volume", BH_FEEDBACK_VOLUME},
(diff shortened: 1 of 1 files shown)


sbird:
Oops, well noticed.

rainwoodman:
With this change if the code is not compiled with BH option dangling BH options in the parameter file will crash the code. Is this really a desired behavior?

sbird:
Maybe not. You think a better fix would be to delete the guards around the definition of BlackHoleFeedbackMethod in allvars.h?

andreufont:
I'd say:
- make BlackHoleOn option compulsory, always.

If code compiled with -DBLACKHOLE:
- read BH related options

If code compiled w/o -DBLACKHOLE:
- crash if BlackHoleOn = 1
- ignore BH related options

Andreu


- if set to 1, read BH related options
- if set to

rainwoodman:
OK. We have a consensus here! I think by dropping the guard and carefully marking all options OPTIONAL could achieve this. If a parameter is really 'REQUIRED' we have to do something else. (it needs changes in the parser I am afraid). Luckily most per-module parameters are optional.

sbird:
@andreufont : Do you want to send a follow-up pull request dropping the guards on BlackHoleFeedbackMethod?

andreufont:
I can do this (and other tests) tomorrow.
Andreu

andreufont:
I believe I know how to do it, but I'm a bit surprised that we are treating BH different than SFR. Should we do the same thing for SFR options?

rainwoodman:
Without thinking too deep, I say yes.

PR #151, opened 2017-10-26
Allocate more memory for stars with quick lyman alpha star formation

sbird:
So we don't run out of memory. This is a short-term hack. The longer term is to allocate star and BH memory with mymalloc2 and increase it dynamically.


rainwoodman:
They shall be allocated from another memory pool i think..

PR #152, opened 2017-10-26
Move 'time is forward' check into run.c

### LINKED ISSUE #149: Lya simulation crashing with: New timestep < old
Hello everyone, my name is Chris and I've just started a PhD with Andreu at UCL. I'm looking at simulating Lyman alpha, and I get the following error:

`[ 000006.78 ] Task 20: Time error: New timestep (0.0243902) < old (0.0243902)!`

This error only crops up at certain starting redshifts (z=10, z=40 and z=60) and the run crashes immediately. I tried a range of sims with different box sizes and starting [redshfits shown here and the red boxes are the sims which crash.](https://imgur.com/a/yF7tR). I have attached the Options.mk and paramfiles I am using (had to convert them to .txt to upload them).

[Options.txt](https://github.com/rainwoodman/MP-Gadget/files/1417874/Options.txt)
[paramfile_gadget.txt](https://github.com/rainwoodman/MP-Gadget/files/1417875/paramfile_gadget.txt)
[paramfile_genic.txt](https://github.com/rainwoodman/MP-Gadget/files/1417876/paramfile_genic.txt)

Is this an issue or am I doing something wrong? Also happy to investigate with other parameters - if you have access to hypatia I've put the sims that failed in:

`/home/chrisp/Projects/testSims/borkedJobs`

Thanks!

sbird:
The check that time is moving forward was prone to floating point and
other problems during simulation startup. Move this into run.c and only
use it after the first timestep completes.

Fixes #149

### DIFF
--- a/run.c
+++ b/run.c
@@ -71,6 +71,9 @@ void run(void)
 
         /*Convert back to floating point time*/
         set_global_time(exp(loga_from_ti(All.Ti_Current)));
+        /*1.0 check for rate setting in sfr_eff.c*/
+        if(NumCurrentTiStep > 0 && All.TimeStep < 0)
+            endrun(1, "Negative timestep: %g New Time: %g!\n", All.TimeStep, All.Time);
 
         int is_PM = is_PM_timestep(All.Ti_Current);
 
--- a/examples/lya/paramfile.genic
+++ b/examples/lya/paramfile.genic
@@ -1,9 +1,9 @@
 OutputDir = output # Directory for output
 FileBase = IC              # Base-filename of output files
 
-Ngrid = 64 # Size of cubic grid on which to create particles.
+Ngrid = 128 # Size of cubic grid on which to create particles.
 
-BoxSize = 32000   # Periodic box size of simulation
+BoxSize = 10000   # Periodic box size of simulation
 
 
 
--- a/init.c
+++ b/init.c
@@ -39,9 +39,6 @@ void init(int RestartSnapNum)
 {
     int i, j;
 
-    /* Important to set the global time before reading in the snapshot time as it affects the GT funcs for IO. */
-    set_global_time(All.TimeInit);
-
     /*Add TimeInit and TimeMax to the output list*/
     if (RestartSnapNum < 0) {
         /* allow a first snapshot at IC time; */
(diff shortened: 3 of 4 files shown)


run.c
```diff
@@ -71,6 +71,9 @@ void run(void)
 
         /*Convert back to floating point time*/
         set_global_time(exp(loga_from_ti(All.Ti_Current)));
```
rainwoodman: What if set_global_time takes the integer time instead? Then there is no rounding off error.

sbird: This is a little tricky (because pesky globals). We must do set_global_time before read_snapshot, because we need All.cf.a, to get the velocity units when reading the snapshot. But before we use an integer time we must do init_timebins. And before we do that we need to do setup_sync_points. I *think* there is no need for setup_sync_points to be after set_global_time...so yes, we can probably do that.

sbird: Except we must do set_global_time(1.0) in sfr_eff.c, and 1.0 may not be on the integer timeline. (also *before* we have init_timebins). That code wants cleaning up at some point...but not today

sbird: But I can do set_global_time(exp(loga_from_ti(All.Ti_Current))) instead of All.TimeInit anyway, which should get rid of the roundoff.

rainwoodman:
Is it possible to add a test case for this?

sbird:
Tricky - it would be a test case for init() which should be tested by the travis run. But this only appears on some machines, I think.

sbird:
I think no: this is an integration failure that appears only on some machines (not our test machines). There is no way we can add a test for it that I can see.

rainwoodman:
Currently set_global_time is not covered by test_timebinmgr.c. Perhaps it shall; but it depends on other things..

For unit testing we need to invert the dependency by defining a
```
TimeStepEvent
```
Where different modules can register event handlers that gets called per step..

Then in the test case no modules have been initialized and we are only testing the barebone (All.Time)...

rainwoodman:
So this is good to merge then?

sbird:
I was waiting for the student to confirm that it works for them, but it should be low peril if you want to merge it now.

Plaetean:
Yes it works, the same sim that crashed earlier has been running for ~1 hour now.

sbird:
Ok! Let's merge it when travis goes green then.

PR #154, opened 2017-10-27
add fixed/paired flags

andersdot:
Sorry, there are spurious changes to spaces and such, kind of hides the real changes, but hopefully it's readable


allvars.c
```diff
@@ -30,6 +30,10 @@ int RestartFlag;		/*!< taken from command line used to start code. 0 is normal s
 				   marks a restart from a snapshot file. */
 int RestartSnapNum;
 
+/* paired/fixed simulation flags*/
+int PairedSim;
+int FixedAmpSim;
+
```
sbird: You don't need to add these parameters to allvars.[ch] if you have already added them to genic/allvars.[ch]: they are only used when generating initial conditions.

genic/allvars.c
```diff
@@ -9,6 +9,8 @@ int ProduceGas;
 int Seed;
 int Unitary;
 int UsePeculiarVelocity;
+int PairedSim;
+int FixedAmpSim;
```
sbird: FixedAmpSim does the same as Unitary, so you can drop it. My guess is that this is a rebase issue...

rainwoodman:
I do not like the names "PairedSim" and "FixedAmpSim"; they describe the goal but not the essense.

Also be aware that PairedSim doesn't actually give you a pair.. 

What about InvertedPhase UnitaryAmplitude?

andersdot:
@rainwoodman uh oh, what do you mean PairedSim doesn't give you a pair? 
It looks like a flag for Unitary is already in zeldovich.c so I'll use/test that, and I like the InvertedPhase keyword

rainwoodman:
Yes -- that's what I mean. Unitary will look kind of out of place after you add InvertedPhase, so I do suggest changing the name to UnitaryAmplitude. [a name change will break the test case and I believe you can correct it in example/travis/paramfile.genic]

sbird:
Oops, you need to pull out the changes from param.c as well. Other than that looks good to me once Yu's name change goes in - we should be sure to merge this squashed into one commit.

andersdot:
Ok, this got a bit messy, but I committed changes to update the flags to InvertPhase and UnitaryAmplitude

genic/zeldovich.c
```diff
@@ -420,7 +420,13 @@ pmic_fill_gaussian_gadget(PM * pm, double * delta_k, int seed, int unitary)
                 /* we want two numbers that are of std ~ 1/sqrt(2) */
                 ampl = sqrt(- log(ampl));
 
-                if (unitary) ampl = 1.0; /* cos and sin gives 1/sqrt(2)*/
+                if (UnitaryAmplitude) ampl = 1.0; /* cos and sin gives 1/sqrt(2)*/
```
emassara: UnitaryAmplitude and InvertPhase are a global variables, in my version I passed them as parameters changing the following lines:

static void gaussian_fill(PetaPMRegion * region, pfft_complex * rho_k, int unitaryAmplitude, int invertPhase)

gaussian_fill(petapm_get_fourier_region(), petapm_get_rho_k(), UnitaryAmplitude, InvertPhase);

pmic_fill_gaussian_gadget(PM * pm, double * delta_k, int seed, int unitaryAmplitude, int invertPhase)

if (unitaryAmplitude) ampl = 1.0; /* cos and sin gives 1/sqrt(2)*/

if (invertPhase) phase = phase + M_PI;  /* invert the sign of the delta */

rainwoodman:
Would it be too much to save these two parameters, together with the random seed to the IC file in this PR?

andersdot:
@rainwoodman if you can be more specific, I don't mind making the changes, just not sure what you mean here

genic/zeldovich.c
andersdot: I set UnitaryAmplitude and InvertPhase in the paramfile for mp-genic. Are you suggesting a change? I think I'm not understanding your comment

rainwoodman: There is an unused parameter `unitary` in the argument list of this function. You can either get rid of that; or you can make sure no global variables is referenced in this function by passing both in as parameters, like Elena's version. Since this piece of code has been somewhat shared between FastPM[1], pmesh[2], and MP-Gadget, the no-global variable approach will produce slightly less headache in the future. But it's really up to you to decide which way to go.

[1] https://github.com/rainwoodman/fastpm/blob/master/libfastpm/initialcondition.c#L141
[2] https://github.com/rainwoodman/pmesh/blob/master/pmesh/_whitenoise_generics.h#L2

rainwoodman:
You can add a few lines here to save additional global variables like Seed, UnitaryAmplitude and InvertedPhase.
https://github.com/rainwoodman/MP-Gadget/blob/680d8d69adb3776416719d3adb339bef168304e9/genic/save.c#L78

andersdot:
I think I made the proper updates, but this is my first time coding in C o.O

rainwoodman:
And you were hacking Gadget -- it's not too bad coding in C isn't it. 

This looks good to me now. I think we can merge it.

rainwoodman:
@andersdot  This is merged. You can remove your branch if you'd like to keep your repo clean.

PR #155, opened 2017-10-30
Starmalloc2

sbird:
This pull request allocates memory for all secondary arrays (star, BH and sph) in one allocation, and makes it resizeable if necessary. This is done by adding realloc to memory.c and allocating a few things in the top allocator, so that the particles are upmost of the bottom allocator once the exchange starts. This should fix the problems found in Issue #141 permanently.

Resizing may be expensive as it needs a memmove on the stars and BH arrays, but with the current defaults I tested that it happens very rarely.

The first few commits are cleanups to memory allocations: alloca->ta_malloc and malloc->mymalloc.
I also update a couple of parameter files and fix a bug in the timeline that gets hit if you change the snapshot outputs when resuming. 

Then I add realloc, with tests. I move a few allocations around so that SphP/StarP is topmost.

I add the dynamical resizing, first to stars then expanded to Sph. 

End with a few small fixes and cleanups.


sbird:
I considered making struct particle_data also reallocable, but All.MaxPart is so embedded in the code (particularly in the treewalk) that I opted not to: it is less important anyway because the domain can always switch to load-balancing instead of work-balancing

rainwoodman:
Before I start to read the code -- do you ask for new memory from OS or it's from our preallocated pages?

sbird:
From our pre-allocated pages

allvars.h
```diff
@@ -197,7 +197,11 @@ extern struct global_data_all_processes
 
     int MaxPart;			/*!< This gives the maxmimum number of particles that can be stored on one
                               processor. */
-    int MaxPartBh;		/*!< This gives the maxmimum number of BH particles that can be stored on one
+    int MaxPartSph;		/*!< This gives the maxmimum number of Sph particles that can be stored on this
```
rainwoodman: Use an array?

blackhole.c
```diff
@@ -714,6 +714,8 @@ void blackhole_make_one(int index) {
 
     P[child].PI = atomic_fetch_and_add(&N_bh_slots, 1);
     P[child].Type = 5;	/* make it a black hole particle */
+    if(P[child].PI >= All.MaxPartBh)
```
rainwoodman: Shall we change the API for forking particles to take a particle type and make the assertion there?

exchange.c
```diff
@@ -47,8 +50,8 @@ int domain_exchange(int (*layoutfunc)(int p), int failfast) {
     /* flag the particles that need to be exported */
     int * toGo_arr[NSP];
     int * toGet_arr[NSP];
-    toGo_arr[0] = (int *) mymalloc("toGo", (NSP * sizeof(int) * NTask));
-    toGet_arr[0] = (int *) mymalloc("toGet", (NSP * sizeof(int) * NTask));
+    toGo_arr[0] = (int *) mymalloc2("toGo", (NSP * sizeof(int) * NTask));
```
rainwoodman: keep in mind NTask can fetch to millions in the near future.

exchange.c
```diff
@@ -233,14 +236,20 @@ static int domain_exchange_once(int (*layoutfunc)(int p), int** toGo_arr, int **
...
 
-    if(N_sph_slots > All.MaxPart)
-        endrun(787878, "Task=%d N_sph=%d All.MaxPart=%d\n", ThisTask, N_sph_slots, All.MaxPart);
-
-    if(N_bh_slots > All.MaxPartBh)
-        endrun(787878, "Task=%d N_bh=%d All.MaxPartBh=%d\n", ThisTask, N_bh_slots, All.MaxPartBh);
-
-    if(N_star_slots > All.MaxPartBh)
-        endrun(787878, "Task=%d N_star=%d All.MaxPartBh=%d\n", ThisTask, N_star_slots, All.MaxPartBh);
+    int newSph = All.MaxPartSph;
```
rainwoodman: An array will really help -- the sizeof() of each element data type can be stored in an array to simplify the interface of realloc_secondary_data too.

exchange.c
```diff
@@ -496,3 +506,31 @@ static int domain_countToGo(ptrdiff_t nlimit, int (*layoutfunc)(int p), int fail
     }
     return ret;
 }
+
+static void
+realloc_secondary_data(int newMaxPartSph, int newMaxPartBh, int newMaxPartStar)
+{
+    size_t bytes = newMaxPartSph * sizeof(struct sph_particle_data) +
+        newMaxPartStar * sizeof(struct star_particle_data) + newMaxPartBh * sizeof(struct bh_particle_data);
+    SphP = myrealloc(SphP, bytes);
```
rainwoodman: It feels a bit uncomfortable with these memmove's and shifting of positions. 
Remember that we only use a PI to locate the secondary data. So one possibility to avoid memcopy is bucketing:

1. keep a linked list of allocations
2. when resolving PI, go though the buckets and till the correct one, find the relative offset in the bucket and return the ptr.

When there are too many buckets the code will slow down badly; a warning shall be issued. The main point of the growing mechanism is to ensure that it doesn't crash out-of-slots right away.

petaio.c
```diff
@@ -189,7 +218,9 @@ void petaio_read_internal(char * fname, int ic) {
         endrun(0, "Failed to close block: %s\n",
                     big_file_get_error_message());
     }
-    allocate_memory(NTotal[0] > 0);
+
+    /*Allocate the particle memory*/
+    petaio_alloc_particle_memory();
```
rainwoodman: I see the old code was doing this too. But when I look at this, It is seems to be clearer if the memory is allocated before read, since the read function doesn't return the allocated pointer as a return value; perhaps it is was impossible to do this previously -- the read_header function (which results All.MaxPart) was decoupled from read_internal only recently.

memory.c
```diff
@@ -263,6 +264,83 @@ allocator_print(Allocator * alloc)
     }
 }
 
+void *
+allocator_realloc_int(Allocator * alloc, void * ptr, size_t new_size, char * fmt, ...)
```
rainwoodman: resize instead of realloc, since the ptr is guaranteed not to move.

blackhole.c
sbird: I think we agreed last round to keep fork_particle type-agnostic. If DM is being forked the check needs to be here.

memory.c
sbird: Only for bottom allocations.

exchange.c
sbird: Sure, but why does this matter? The memory is the same no matter whether it is top or bottom

memory.c
rainwoodman: I see...

blackhole.c
rainwoodman: Assigning slot for the type can be made type agnostic. I'll do a PR on this PR.

exchange.c
sbird: I bet you any money that the memmove is faster.

Note that the BH is only moved when SPH gets larger (I could add a (if(src!=dest) memmove(dest,src), but I'm assuming that any competent memmove implementation does that itself anyway) and the initial allocation for Sph is the same as it always was - since we have only once had a situation where Sph ran out of memory, I expect that in practice the BH will never move.

So really it is just the stars, which are very small - sizeof(StarP[0]) is 32 bytes, I think. We are talking memmoving a total of ~30MB. memmove is very fast. Also they get moved max once or twice a simulation. Whereas the bucket access penalty will be every timestep during GC.

exchange.c
sbird: Note that the particle number increase is slightly different for different particles, so we still need a nest of if statements.

allvars.h
sbird: It is possible, but I am not so sure it is cleaner, honestly. We use bare variables for the slots as well...

rainwoodman:
Let's merge this as is and I'll merge my changes on top of this.

PR #159, opened 2017-10-31
SlotsManager

rainwoodman:
The SlotsManager module deals with the slots.

    I wrote a simpler version of realloc. since it only works on the last item, the operation can be built with dealloc and alloc. Also properly handled the vaarg.

    All 6 particle types now can have their own secondary slots. This is declared in slot_init(). All specilized treatments are eliminated ; except bh and sfr_eff which may need additional API from slotsmanager to cleanup.

    GC should be slightly faster (if compiler optimization does its work) now that we only loop over NumPart once. code may need some refactoring -- to many if's conditioning on whether a slot type is enabled.

    old names, SphP, BhP StarP are now macros. Variable version of these are internally maintained by slotsmanager -- for debugging, prefixed with GDB_

    slots_fork() handles the consistency of secondary slots and base.

    the strange throttling code is removed -- without it the code works just fine. It seems to handle an extremely low memory case (package > limit) where we shall probably just crash anyways.


petaio.c
```diff
@@ -222,57 +223,63 @@ void petaio_read_internal(char * fname, int ic) {
...
+        newSlots[4] +=  0.01 * All.MaxPart;
     }
-    if(All.BlackHoleOn || N_bh_slots > 0) {
-        All.MaxPartBh = All.PartAllocFactor * N_bh_slots + 0.01 * All.MaxPart;
+    if(All.BlackHoleOn || NTotal[5] > 0) {
+        newSlots[5] +=  0.01 * All.MaxPart;
     }
+
+    /* make sure identical numbers are for all ranks; move this to grow? */
+    MPI_Allreduce(MPI_IN_PLACE, newSlots, 6, MPI_INT, MPI_MAX, MPI_COMM_WORLD);
```
sbird: Why is this needed? Initially we guarantee it is the same (eg, line 221) and much later we don't want them to be the same, because the stars may be very inhomogenous.

petaio.c
```diff
@@ -222,57 +223,63 @@ void petaio_read_internal(char * fname, int ic) {
...
-        message(0, "Allocated %g MB for %d SPH, %d Stars and %d BHs.\n", bytes / (1024.0 * 1024.0),All.MaxPartSph,
-                All.MaxPartStar, All.MaxPartBh);
+
+    slots_reserve(newSlots);
+
+    /* set up the memory topology */
+
+    int offset = 0;
+    for(ptype = 0; ptype < 6; ptype ++) {
+        /* set use this many slots; FIXME: encapuslate this */
```
sbird: encapsulate :)

sfr_eff.c
```diff
@@ -262,8 +262,9 @@ void cooling_and_starformation(void)
 
     double totsfrrate, localsfr=0;
     int i;
+    /* FIXME: this is inaccurate if some particles are made into garbage. may want to propagate garbage flag to the slots.*/
```
sbird: I think this is fine, because we want the SFR from the garbage particles. Generally they will have been turned into stars...

slotsmanager.c
```diff
@@ -0,0 +1,427 @@
...
+    P[child].Mass = 0;
+    P[child].Type = ptype;
+
+    if(SlotsManager->info[ptype].enabled) {
+        /* if enabled, alloc a new Slot for secondary data */
+        int PI = atomic_fetch_and_add(&SlotsManager->info[ptype].size, 1);
+
+        if(PI >= SlotsManager->info[ptype].maxsize) {
+            /* rare case, use an expensive critical section */
+            #pragma omp critical
```
sbird: Right, this will not work with the memory allocator. Also the critical section is not sufficient, because not all accesses to SlotsManager are guarded by a similar critical section. 

This case is very rare: until recently we did not even check for it and we have never seen a problem. I would drop the code.

slotsmanager.c
```diff
@@ -0,0 +1,427 @@
...
+        if(!SlotsManager->info[P[i].Type].enabled) continue;
+
+        if(P[i].PI >= SlotsManager->info[P[i].Type].size) {
+            endrun(1, "slot PI consistency failed2\n");
+        }
+        if(BASESLOT(i)->ID != P[i].ID) {
+            endrun(1, "slot id consistency failed2\n");
+        }
+#pragma omp atomic
+        used[P[i].Type] ++;
```
sbird: This variable is unused. Did you mean to use it for something?

slotsmanager.c
```diff
@@ -0,0 +1,427 @@
...
+    int good = 1;
+
+    if(SlotsManager->Base == NULL)
+        SlotsManager->Base = (char*) mymalloc("SlotsBase", 0);
+
+    for(ptype = 0; ptype < 6; ptype ++) {
+        newMaxSlots[ptype] = SlotsManager->info[ptype].maxsize;
+        while(newMaxSlots[ptype] < atleast[ptype]) {
+            int add = 0.2 * newMaxSlots[ptype];
+            if (add < 128) add = 128;
```
sbird: You have changed the increase amount here (made it smaller). I am worried that the new condition does not ensure that there are enough slots available for the next step's star formation, if, e.g. newMaxSlots = 1.21 * atleast. (although I agree that the old condition was overly complex, given that the SPH slot needs increasing very very rarely).

I think a fix would be to scale this with total particle load: add = 0.005 * All.MaxPart. That way we don't risk running out of slots in highly threaded situations.

slotsmanager.c
```diff
@@ -0,0 +1,427 @@
...
+            newMaxSlots[ptype] += add;
+            good = 0;
+        }
+    }
+    /* no need to grow, already have enough */
+    if (good) {
+        return;
+    }
+    /* FIXME: do a global max; because all variables in All.* are synced between ranks. */
```
sbird: I don't think we want to fix this fixme: stars, SPH and BH can be inhomogeneous between ranks (because stars are very clustered). I was toying with the idea of one day allowing SPH memory to shrink sometimes to allow for this

slotsmanager.c
```diff
@@ -0,0 +1,427 @@
...
+            MPI_COMM_WORLD);
+
+    /* if any rank is too fragmented, add a sorting to defrag. */
+    if(defrag)  {
+        /* sort the used ones
+         * by their location in the P array */
+        qsort_openmp(SlotsManager->info[ptype].ptr,
+                     SlotsManager->info[ptype].size,
+                     SlotsManager->info[ptype].elsize, 
+                     slot_cmp_reverse_link);
```
sbird: Looks good. I did some profiling (perhaps you did too) and qsorting the SPH slots was taking 90% of the GC time. So avoiding this if defrag is too high is a good idea. We should probably profile this a bit though: if the sort is run too rarely the tree will become slow (and if too often the gc takes too much time). But we can mess with that later.

sbird:
One other fix: because the garbage collection does not run every timestep now you need to do 
```
if(P[i].IsGarbage) continue
```
in the find_timesteps loop and in rebuild_activelist. Otherwise you can have the smallest timestep contain only a garbage particle, which causes an empty timebin when it gets collected and a crash.

sbird:
And there are a few other places where the persistence of a garbage particle causes problems: maybe better to make the defrag unconditional and deal with them all in a separate pull request later.

sbird:
Wait, the above does not make sense - the base gc is unconditional, so all GC particles should be collected at the same time. I suspect the domain exchange is getting corrupted somehow. Tests needed!

Also starformation() is run after domain(), so we should have the IsGarbage check anyway I think. It is just that it is quite unlikely to cause a problem.

slotsmanager.c
rainwoodman: I though the critical is sufficient, for all access to info[ptype].maxsize is protected by it. The current code will endrun if a problem occurs -- so probably just drop the critical and grow?

slotsmanager.c
rainwoodman: Yes. Later. I was thinking adding a separate defrag that is called only in full domain decomposition.

slotsmanager.c
rainwoodman: Is 0.5% enough if we have very fast SFR? This is probably point to the need to grow in fork.

slotsmanager.c
rainwoodman: It was a left over. My fix was to move MaxSlots out of All. I think I removed the code to handle shrinking. in the case of shrinking the memmove have to be done in a different order.

slotsmanager.c
rainwoodman: I thought it was to be printed in the DEBUG mode...

rainwoodman:
Yes. The IsGarbage check is needed whenever a particle is touched. setting IsGarbage shall be made a method though.

slotsmanager.c
sbird: The critical isn't enough because we have the memmove in slots_reserve: this will be bad if we are accessing BhP at the same time. I don't know if there is any reasonable way around this without a stop_the_world() routine...

slotsmanager.c
sbird: Oh, I see. That's fine.

slotsmanager.c
sbird: Yeah, I don't know, but it's probably ok. The last version I had increased the star allocation every time N_star_slots > MaxPartStar - 0.005*MaxPart, so it had to exceed 0.5% of the particles in every timestep, which seemed pretty safe? I would guess that the timestep would shrink if star formation got faster than that.

rainwoodman:
OK looks like a unit test is important. (per #141 ) I'll try to address the issues and see how far I can go with the unit tests. Hopefully this PR can solve the issue there.

petaio.c
rainwoodman: It was a left over -- previously MaxSlots were under All and shall be idential cross ranks. Now they are not.

sbird:
Many apologies that I do not have time to work on this until the end of next week. Especially since I broke it in the first place with the star-part branch.

slotsmanager.c
rainwoodman: That's correct. The access to the secondary data is not atomic. I'll change this to death spot on.

petaio.c
rainwoodman: It's now actually moved to slotsmanager.

exchange.c
```diff
@@ -80,89 +89,67 @@ int domain_exchange(int (*layoutfunc)(int p), int failfast) {
...
+                NumPart, plan->toGetSum.base, plan->toGoSum.base, All.MaxPart);
         bad_exh = 1;
     }
 
     MPI_Allreduce(MPI_IN_PLACE, &bad_exh, 1, MPI_INT, MPI_LOR, MPI_COMM_WORLD);
+
     if(bad_exh) {
-        myfree(ctmem);
+        myfree(toGetPtr);
+        myfree(toGoPtr);
```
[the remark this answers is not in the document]
rainwoodman: changed

exchange.c
```diff
@@ -80,89 +89,67 @@ int domain_exchange(int (*layoutfunc)(int p), int failfast) {
...
-        for(i = 0; i < NTask; i++)
-        {
-            count_togo[j] += toGo_arr[j][i];
-            count_get[j] += toGet_arr[j][i];
-        }
-    }
+
     /*Check whether the domain exchange will succeed. If not, bail*/
-    if(NumPart + count_get[0] - count_togo[0] > All.MaxPart){
+    if(NumPart + plan->toGetSum.base - plan->toGetSum.base > All.MaxPart){
```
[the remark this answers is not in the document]
rainwoodman: fixed

exchange.c
```diff
@@ -6,56 +6,65 @@
...
+{
+    int i;
+    for(i = 0; i < NTask; i ++) {
+        if(ptype == -1) {
+            count[i] = entries[i].base;
+        } else {
+            count[i] = entries[i].slots[ptype];
+        }
+    }
+}
```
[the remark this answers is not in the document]
rainwoodman: added a comment

sfr_eff.c
rainwoodman: I am propagating IsGarbage to the secondary slots. So we have the option to skip them.

slotsmanager.c
rainwoodman: So We'll always set MaxSlots to 0.5 % + the requested reservation.

sbird: That's not quite as aggressive as it was before: before it increased the allocation if N_star_slots was within 0.5% of the total. Now it is possible for N_star_slots to get arbitrarily close to the total before being increased. If a burst of star formation occurs at this point we are in trouble.

rainwoodman:
I  wonder if we shall merge neutrino first before doing this. What do you say?

sbird:
Sure, I'm happy to resolve the conflicts and push the button if you are ready. Is there some infrastructure in neutrino you would particularly like to re-use?

rainwoodman:
No - I worry this will break neutrinos and cause another massive rebase there.

sbird:
Thanks for the merge. Do you want help fixing this up? I have time from tomorrow.

rainwoodman:
I am looking into the zero SFR issue from travis. But take a look if the simpler unit test covers the same cases you had -- I realized exchange doesn't care about particle position at all -- and asserting on ID is a lot simpler.

sbird:
Something like this approach is fine, but I explicitly wanted to test some of the other cases:
1) Where there are different numbers of different particles (I've seen bugs that only manifest under these conditions before).
2) Where a realloc happens
3) Where there is DM only but no active slots (so that if there is a bug, we know roughly where it is).

I also wanted a benchmark test case where there were a lot of SPH particles, so we could test the performance of our progressively-smarter gc. But that is less important.

rainwoodman:
1. Where there are different numbers of different particles (I've seen
   bugs that only manifest under these conditions before).

This is no longer relevant because the slotsmanager deals with all 6
species the same way. I added a test where some particle species have zero
particles.

   1. Where a realloc happens

I added a case for this -- It triggered a hanging bug in slotsmanager
related to collective printing (if used[] is zero on some ranks but not on
others). I am too lazy to add a regression test.

The test introduces an explicit between test_exchange and slotsmanager --
which I am not that happy about.

   1. Where there is DM only but no active slots (so that if there is a
   bug, we know roughly where it is).

The test already covers slot types that are enabled and not enabled --
currently ptype of 1 2 3 are treated like traditional DM. Would the 'zero'
particle test work for this? Again be aware slots manager knows nothing
special about DM or Gas. If elsize is zero the slot type is disabled. If it
is not zero then it is enabled; every species is treated like a loop -- I
think the code may actually even work for different number of species..

rainwoodman:
I am not so sure about benchmarks -- it usually brings the imagination of having some report generating mechanism.. But currently the parallel cmoka output is already very messy..

sbird:
Thanks - I am happy with the tests as currently written.

examples/travis/paramfile.gadget
```diff
@@ -81,9 +81,9 @@ TimeBetweenSeedingSearch 1.03
 CritPhysDensity = 0       #  critical physical density for star formation in
 #  hydrogen number density in cm^(-3)
 
-CritOverDensity = 57.7   #  overdensity threshold value
+CritOverDensity = 0   #  overdensity threshold value
 
-QuickLymanAlphaProbability = 0 # Set to 1.0 to turn dense gas directly into stars.
+QuickLymanAlphaProbability = 0.001 # Set to 1.0 to turn dense gas directly into stars.
```
sbird: Did you make these changes on purpose?

exchange.c
```diff
@@ -6,56 +6,72 @@
...
 int domain_exchange(int (*layoutfunc)(int p), int failfast) {
```
sbird: The failfast argument isn't needed anymore: I can remove it if you like.

rainwoodman:
Sure. Please.

rainwoodman:
Yes. I wanted the integration
 Travis test to cover the slots.

sbird:
There we go. Some other cleanups fell out when I removed failfast: no point looping now, since we don't do anything with the loop.

rainwoodman:
nlimit is actually used. The logic was
```
move at most nlimit particles
loop till all particles are moved
```

rainwoodman:
in memory bound situation we do need several iterations in the exchange.

sbird:
I see - in that case I think we want something like this? Since all the small arrays are ta_malloc now, I think we can also increase nlimit to FreeBytes - 2 *NTask * sizeof(MPI_Request)?

sbird:
The MPI_Request array in system.c should probably be on the heap too...

sbird:
Shall I merge neutrinos as discussed above, to avoid conflict? I would quite like to avoid a big rebase, it's true, since I am finishing off the paper right now.

rainwoodman:
Sure.

On Nov 16, 2017 3:10 PM, "Simeon Bird" <notifications@github.com> wrote:

Shall I merge neutrinos as discussed above, to avoid conflict? I would
quite like to avoid a big rebase, it's true, since I am finishing off the
paper right now.

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/rainwoodman/MP-Gadget/pull/159#issuecomment-345079222>,
or mute the thread

rainwoodman:
Green. Merging.

sbird:
Alas - a long-running test just found a bug. A clean run of run.sh from examples/lya crashes with this:

[ 003171.13 ] Begin Step 426, Time: 0.0753588 (1d3000), Redshift: 12.2698, Nf = 00000000287123, Systemstep: 0.00011835, Dloga: 0.00157172, status: 
[ 003171.13 ] TotNumPart: 0000004194304 SPH 0000002097090 BH 0000000000 STAR 0000000000062 
[ 003171.13 ] Occupied:            0            1            2            3            4            5 dt
[ 003171.13 ]    bin=14       260775       215659            0            0            0            0 0.0251475
[ 003171.13 ]    bin=13      1644906      1785841            0            0            0            0 0.0125737
[ 003171.13 ]  X bin=12       182646        93693            0            0            0            0 0.00628687
[ 003171.13 ]  X bin=11         8746         1929            0            0           42            0 0.00314343
[ 003171.13 ]  X bin=10           17           30            0            0           20            0 0.00157172
[ 003171.13 ]                -----------------------------------
[ 003171.13 ] Total:          191409        95652            0            0           62            0  Sum:        287123
[ 003171.13 ] Start force computation...
[ 003171.13 ] Begin tree force.  (presently allocated=356.297 MB)
[ 003171.90 ] tree is done.
[ 003171.90 ] Start density computation...
[ 003172.51 ] ngb iteration 1: need to repeat for 3798 particles.
[ 003172.58 ] Hmax exchange: 0 toplevel tree nodes out of 71
[ 003172.58 ] Start hydro-force computation...
[ 003173.73 ] SFR: spawned 0 stars, converted 6 gas particles into stars
[ 003173.74 ] force computation done.
[ 003173.88 ] Attempting a domain exchange
[ 003173.92 ] Using 2220388224 bytes for exchange.
[ 003173.94 ] iter=0 exchange of 0000000000455 particles
[ 003173.96 ] GC : Reducing Particle slots from 4194310 to 4193849
[ 003173.99 ] GC: Used slots for type 0 is 283884
[ 003173.99 ] GC: Used slots for type 4 is 0
[ 003174.00 ] Task 4: IsGarbage flag inconsistent between secondary and base: P[465].Type=4

PR #162, opened 2017-11-12
Memory sort

sbird:
I noticed that system qsort, which for glibc is a merge sort, allocated O(n) temporary storage using malloc. This is not good for our memory fragmentation, especially as it does it under heavy thread contention. It is also unnecessary, as we already have temporary memory allocated for the parallelization, which we can re-use. 

I fixed this by copying glibc's qsort implementation of merge sort into openmpsort.c and renaming the temp variable. While I was there I also extended indirect sorting for large structures to the merge phase, which should help for the sort of struct particle_data in domain.c and reduces the maximum memory allocation. Another advantage of this is that a smart compiler with lto could possibly inline the comparison function.

We could probably make the sorting much faster by using introsort (which is in every C++ standard library) instead of glibc's mergesort, but I didn't bother (yet).

### DIFF
--- a/fof.c
+++ b/fof.c
@@ -14,6 +14,7 @@
 #include "drift.h"
 #include "domain.h"
 #include "mpsort.h"
+#include "openmpsort.h"
 #include "mymalloc.h"
 #include "forcetree.h"
 #include "endrun.h"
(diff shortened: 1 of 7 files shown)


rainwoodman:
Looks good first glance. Go ahead as long as tests are passed!

Yu

sbird:
Merged, thanks!

PR #163, opened 2017-11-14
Test exchange

sbird:
This adds a unit test for the exchange code. 

It fixes a bug: reallocing was broken if both Stars and BHs were there, because pointer arithmetic is not associative for different pointer types (duh).
It also fixes a minor suboptimality: garbage particles were put in the exchange buffer.

I did this on master instead of the slotmanager branch: this means a rebase is necessary but it also means that one can bisect if a commit in slotmanager breaks the test.

### DIFF
--- a/exchange.c
+++ b/exchange.c
@@ -327,6 +327,7 @@ domain_build_plan(ptrdiff_t nlimit, int (*layoutfunc)(int p), ExchangePlan * pla
     {
         if(package >= nlimit) {insuf = 1; break; }
         if(!P[n].OnAnotherDomain) continue;
+        if(P[n].IsGarbage) continue;
 
         int target = layoutfunc(n);
         if (target == ThisTask) continue;
--- a/Makefile.tests
+++ b/Makefile.tests
@@ -25,8 +25,11 @@ TESTED = \
 	openmpsort \
 	genic-power
 
-TESTBIN = $(TESTED:%=tests/test_%)
-SUITE?= $(TESTBIN)
+MPI_TESTED = exchange
+
+TESTBIN = $(TESTED:%=tests/test_%) $(MPI_TESTED:%=tests/test_%)
+SUITE?= $(TESTED:%=tests/test_%)
+MPI_SUITE?= $(MPI_TESTED:%=tests/test_%)
 
 .PHONY: test run-tests build-tests
 
@@ -37,6 +40,7 @@ depends/lib/libcmocka.a:
 build-tests: $(TESTBIN)
 
 test : build-tests
+	trap 'err=1' ERR; for tt in $(MPI_SUITE) ; do mpirun -np 4 $$tt ; done; exit $$err
 	trap 'err=1' ERR; for tt in $(SUITE) ; do $$tt ; done; exit $$err
 
 clean-tests:
--- a/tests/test_exchange.c
+++ b/tests/test_exchange.c
@@ -27,6 +27,68 @@ struct particle_data *P;
 struct global_data_all_processes All;
 int NumPart;
 /*Dummy*/
+static int
+order_by_id(const void *a, const void *b)
+{
+    const MyIDType * pa  = (const MyIDType *) a;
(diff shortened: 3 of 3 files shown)


sbird:
Having now done the rebase, it was a waste of time to do it on master: the code has changed a lot in slotsmanager and bisection is too hard anyway. I have rebased the code onto the slotsmanager branch. If that doesn't land soon we should commit 89332e2 to master as that fixes a bug, but it is also fixed in the slots branch.

rainwoodman:
I'll try to land it tomorrow.

Yu

sbird:
Here is a rebased version, to be merged into starmalloc2. For some reason the test fails now.

rainwoodman:
Looks like the code is not compiled with -DDEBUG.

exchange.c
```diff
@@ -162,6 +162,7 @@ static int domain_exchange_once(int (*layoutfunc)(int p), ExchangePlan * plan)
     for(i = 0; i < NumPart; i++)
     {
         if(!(P[i].OnAnotherDomain && P[i].WillExport)) continue;
+        if(P[i].IsGarbage) continue;
```
rainwoodman: Why would we want to export an garbage? Maybe the planning routine shall be able to know garbage is not supposed to be exported.

exchange.c
```diff
@@ -327,6 +328,7 @@ domain_build_plan(ptrdiff_t nlimit, int (*layoutfunc)(int p), ExchangePlan * pla
     {
         if(package >= nlimit) {insuf = 1; break; }
         if(!P[n].OnAnotherDomain) continue;
+        if(P[n].IsGarbage) continue;
```
rainwoodman: If we have this line then the line above can be an assertion?

exchange.c
sbird: Yes - that is what this commit does, isn't it? It skips exporting any garbage particle, also in the planning routine.

rainwoodman:
It needs to link against exchange.c

Shall we introduce libgadget.a that archives all of the object files to fix this once for all?

Some tests have dummy global variables that can be eliminated as well.  If you don't see any shortcomings I'll go ahead and do that.

exchange.c
sbird: My understanding is that build_plan counts particles and domain_exchange copies them into the buffer: so you need to skip garbage in both places or the counts will not correspond to the buffer sizes. Is that not right?

sbird:
I think I was avoiding doing that so one could still provide dummy "mock" versions that were guaranteed to work and simplify testing (eg, by removing MPI components). I cant remember if we still do that to any extent though.

But anyway, that's a different problem - we should merge starmalloc2 first. And even if the compile time thing is fixed the tests still fail locally at runtime, which needs to be fixed too.

exchange.c
rainwoodman: But WillExport of garbage is always False.

rainwoodman:
Most(if not all) of the tests are running though the MPI enabled stub -- we in principle are just starting MPI apps with 1 rank for non-MPI tests (allows the MPI based logging routines to run)

exchange.c
sbird: Ah, indeed! Absolutely - this line is an assertion. I will fix that.

sbird:
Right - sorry, I meant that we could provide a dummy routine which, eg, instead of broadcasting a struct to all processors would just be a null-op. Or that skipped touching a large global array. 
I remember doing this for one iteration of the forcetree tests (a simpler version of domain_get_topleaf), but I don't think we do it anymore, which suggests it is probably indeed fine to just link everything into libgadget.a 

Why not commit that change to master and then we rebase this and starmalloc2 on top of that?

rainwoodman:
I sort of feel it is better to do that after these land to master. I'll merge this PR to starmalloc2,  then fix it and merge it

rainwoodman:
The tests are quite convoluted....

rainwoodman:
I made a simpler version because the exchange doesn't care about particle positions at all. Could you take a look if it is sufficient?

PR #164, opened 2017-11-17
Fixfork

sbird:
Fix the crash I was seeing in the master branch: the slot was not zerod after allocation, so IsGarbage could persist. Still running a long test on this.

### DIFF
--- a/slotsmanager.c
+++ b/slotsmanager.c
@@ -27,7 +27,7 @@ slots_gc_slots();
 
 /*Initialise a new slot for the particle at index i.*/
 static void
-slots_new_slot(int i, size_t size)
+slots_connect_new_slot(int i, size_t size)
 {
     memset(BASESLOT(i), 0, size);
     /* book keeping ID: debug only */
@@ -92,7 +92,7 @@ slots_fork(int parent, int ptype)
             endrun(1, "Assertion Failure more PI than available slots : %d > %d\n",P[child].PI, SlotsManager->info[ptype].maxsize);
         }
 
-        slots_new_slot(child, SlotsManager->info[ptype].elsize);
+        slots_connect_new_slot(child, SlotsManager->info[ptype].elsize);
     }
 
     /*! When a new additional star particle is created, we can put it into the
--- a/slotsmanager.c
+++ b/slotsmanager.c
@@ -25,6 +25,16 @@ slots_gc_base();
 static int
 slots_gc_slots();
 
+/*Initialise a new slot for the particle at index i.*/
+static void
+slots_new_slot(int i, size_t size)
+{
+    memset(BASESLOT(i), 0, size);
+    /* book keeping ID: debug only */
+    BASESLOT(i)->ID = P[i].ID;
+    BASESLOT(i)->IsGarbage = P[i].IsGarbage;
+}
+
 int
 slots_fork(int parent, int ptype)
 {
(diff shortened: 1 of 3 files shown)


slotsmanager.c
```diff
@@ -82,6 +82,9 @@ slots_fork(int parent, int ptype)
             endrun(1, "Assertion Failure more PI than available slots : %d > %d\n",P[child].PI, SlotsManager->info[ptype].maxsize);
         }
 
+        /*Remember to zero the new slot*/
+        memset(BASESLOT(child), 0, SlotsManager->info[ptype].elsize);
```
rainwoodman: This is right. But what about calling the loop body of this (refactoring it out as a function first):
https://github.com/rainwoodman/MP-Gadget/blob/master/slotsmanager.c#L479

rainwoodman: I ran into this when writing the test case. So basically whenever a slot is attached to a particle it shall be initialized.

sbird:
Sure - how about that?

slotsmanager.c
sbird: I left the loop body intact because we don't want to memset the slot to zero in slots_setup_id: we just read data into those particles.

rainwoodman: I don't think the data is read in yet when the loop is called.

rainwoodman:
yep line 382 of petaio.c -- it occured before reading -- so I suppose it is OK to memset in the loop as well. The function name 'new_slot' lacks a verb. What about slots_connect_new_slot()?

sbird:
That's the wrong function - slots_setup_id is called on line 429, after reading. It has to be, because othrwise there is no ID to set up. slots_connect_new_slot sounds fine.

rainwoodman:
My bad -- so setup_id is for diagnostics but connect_new_slot is for data consistency. How did your long test go?

sbird:
Running fine to z=4.15 at the moment - let me run it to z=4 and check the output is the same.

rainwoodman:
Sure. What about getting rid of memset in connect_new_slot? Because we shall initialize the slot properly after the connection anyways.

sbird:
My concern is that at some point in the future we might forget to initialise some variable, maybe one that is set with a +=, and then we get wrong results. But I'm open to skipping the memset anyway if you like - it's fine at the moment

rainwoodman:
If that's the case, we shall probably then somehow fill it with 0xBE or some other strange numbers that likely certainly will crash the code very quickly?

sbird:
Hey, that's a good idea. Is there a series of bytes which do not decode to a valid FP value?

sbird:
I guess not - can use 0xBE.

rainwoodman:
http://www.brycevandyk.com/why-does-mozilla-use-e4-and-e5-to-fill-alloced-and-freed-memory/

sbird:
Yeah, e4 would be a good choice.

sbird:
Ok, the long test has reached z=2.8 fine, and looks about the same as before. Slightly lower star formation, but within the variance expected for a threaded run at the moment. This looks ok to merge to me.

rainwoodman:
Memory.c shall probably set e4 and e5 too. What do you think?

sbird:
That is a lot more expensive - how about doing it only in debug mode?

PR #165, opened 2017-11-21
A bit more garbage

sbird:
Continuing in our series of improving the garbage collection - this is a series with a few more tweaks. The differences are actually rather slight. The main commit is 9ba9d25, which adds a new garbage collection algorithm which is O(n) when compacting the slots, and preserves the (peano) order of the particles. This also lets the non-slot-compacting part avoid changing the slot array entirely, but because more data is moved, the base gc is slower. On my machine this works out to exactly the same time taken doing GC. The slot gc might be faster but with the default parameters it doesn't run very often. 
I also moved the peano sort into slotmanager and added a slot sort after it.

The main advantage is that treebuild is 20% faster on average, because the ordered particles make it easier for the optimizations to kick in. The better ordering makes no difference to the treewalk. Overall this made surprisingly little difference, which may suggest we have optimized as much as we can.

### DIFF
--- a/exchange.c
+++ b/exchange.c
@@ -25,8 +25,8 @@ typedef struct {
     ExchangePlanEntry toGoSum;
     ExchangePlanEntry toGetSum;
 } ExchangePlan;
-/* 
- * 
+/*
+ *
  * exchange particles according to layoutfunc.
  * layoutfunc gives the target task of particle p.
 */
@@ -173,8 +173,7 @@ static int domain_exchange_once(int (*layoutfunc)(int p), ExchangePlan * plan)
     ta_free(toGoPtr);
     walltime_measure("/Domain/exchange/makebuf");
 
-    /* FIXME: This needs some benchmarking to pick a good value!
-     * Arguably this should be type-specific, but that feels like too much complexity.*/
+    /* This is a reasonable value which usually skips gc on the slots.*/
     slots_gc(0.1);
 
     walltime_measure("/Domain/exchange/garbage");
@@ -253,7 +252,7 @@ static int domain_exchange_once(int (*layoutfunc)(int p), ExchangePlan * plan)
             }
         }
         for(ptype = 0; ptype < 6; ptype ++) {
-            if(newPI[ptype] != 
+            if(newPI[ptype] !=
                 SlotsManager->info[ptype].size + plan->toGetOffset[src].slots[ptype]
               + plan->toGet[src].slots[ptype]) {
                 endrun(1, "N_slots mismatched\n");
--- a/exchange.c
+++ b/exchange.c
@@ -77,6 +77,8 @@ int domain_exchange(int (*layoutfunc)(int p)) {
 #pragma omp parallel for
     for(i = 0; i < NumPart; i++)
     {
+        if(P[i].IsGarbage)
+            continue;
         int target = layoutfunc(i);
         if(target != ThisTask)
             P[i].OnAnotherDomain = 1;
@@ -319,7 +321,6 @@ domain_build_plan(int (*layoutfunc)(int p), ExchangePlan * plan)
             break;
         }
         if(!P[n].OnAnotherDomain) continue;
-        if(P[n].IsGarbage) continue;
 
         int target = layoutfunc(n);
         if (target == ThisTask) continue;
--- a/exchange.c
+++ b/exchange.c
@@ -173,10 +173,14 @@ static int domain_exchange_once(int (*layoutfunc)(int p), ExchangePlan * plan)
     ta_free(toGoPtr);
     walltime_measure("/Domain/exchange/makebuf");
 
-    /* gc the slots here if we are low on slot memory. */
+    /*Find which slots to gc*/
     int compact[6] = {0};
     for(ptype = 0; ptype < 6; ptype++) {
+        /* gc if we are low on slot memory. */
         if (SlotsManager->info[ptype].size + plan->toGetSum.slots[ptype] > 0.95 * SlotsManager->info[ptype].maxsize)
+            compact[ptype] = 1;
+        /* gc if we had a very large exchange. */
+        if(plan->toGoSum.slots[ptype] > 0.1 * SlotsManager->info[ptype].size)
             compact[ptype] = 1;
     }
     slots_gc(compact);
--- a/exchange.c
+++ b/exchange.c
@@ -154,7 +154,6 @@ static int domain_exchange_once(int (*layoutfunc)(int p), ExchangePlan * plan)
         P[i].WillExport = 0;
         target = layoutfunc(i);
 
-        /* mark this particle as a garbage for removal later */
         int ptype = P[i].Type;
 
         /* watch out thread unsafe */
@@ -167,7 +166,6 @@ static int domain_exchange_once(int (*layoutfunc)(int p), ExchangePlan * plan)
         /* now copy the base P; after PI has been updated */
         partBuf[plan->toGoOffset[target].base + toGoPtr[target].base] = P[i];
         toGoPtr[target].base ++;
-
         /* mark the particle for removal. Both secondary and base slots will be marked. */
         slots_mark_garbage(i);
     }
@@ -175,11 +173,10 @@ static int domain_exchange_once(int (*layoutfunc)(int p), ExchangePlan * plan)
     ta_free(toGoPtr);
     walltime_measure("/Domain/exchange/makebuf");
 
-    /* now remove the garbage particles because they have already been copied.
-     * eventually we want to fill in the garbage gap or defer the gc, because it breaks the tree.
-     * invariance . */
+    /* FIXME: This needs some benchmarking to pick a good value!
+     * Arguably this should be type-specific, but that feels like too much complexity.*/
+    slots_gc(0.1);
 
-    slots_gc();
     walltime_measure("/Domain/exchange/garbage");
 
     int newNumPart;
@@ -299,7 +296,7 @@ domain_build_plan(int (*layoutfunc)(int p), ExchangePlan * plan)
     int n;
     size_t package;
     int ptype;
-    ptrdiff_t nlimit = FreeBytes - NTask * 2 * sizeof(MPI_Request);
+    ptrdiff_t nlimit = FreeBytes - NTask * 2 * sizeof(MPI_Request) - SlotsManager->garbage * sizeof(int);
 
     memset(plan->toGo, 0, sizeof(plan->toGo[0]) * NTask);
 
@@ -327,6 +324,8 @@ domain_build_plan(int (*layoutfunc)(int p), ExchangePlan * plan)
 
         plan->toGo[target].base += 1;
         nlimit -= sizeof(P[0]);
+        /* For the GC*/
+        nlimit -= sizeof(int);
 
         int ptype = P[n].Type;
         plan->toGo[target].slots[ptype] += 1;
--- a/exchange.c
+++ b/exchange.c
@@ -280,12 +280,12 @@ static int domain_exchange_once(int (*layoutfunc)(int p), ExchangePlan * plan)
         SlotsManager->info[ptype].size = newSlots[ptype];
     }
 
-    walltime_measure("/Domain/exchange/finalize");
-
 #ifdef DEBUG
     domain_test_id_uniqueness();
     slots_check_id_consistency();
 #endif
+    walltime_measure("/Domain/exchange/finalize");
+
     return 0;
 }
 
--- a/exchange.c
+++ b/exchange.c
@@ -183,6 +183,8 @@ static int domain_exchange_once(int (*layoutfunc)(int p), ExchangePlan * plan)
         if(plan->toGoSum.slots[ptype] > 0.1 * SlotsManager->info[ptype].size)
             compact[ptype] = 1;
     }
+    /*Make the slot compaction collective*/
+    MPI_Allreduce(MPI_IN_PLACE, &compact, 6, MPI_INT, MPI_LOR, MPI_COMM_WORLD);
     slots_gc(compact);
 
     walltime_measure("/Domain/exchange/garbage");
--- a/exchange.c
+++ b/exchange.c
@@ -173,8 +173,13 @@ static int domain_exchange_once(int (*layoutfunc)(int p), ExchangePlan * plan)
     ta_free(toGoPtr);
     walltime_measure("/Domain/exchange/makebuf");
 
-    /* This is a reasonable value which usually skips gc on the slots.*/
-    slots_gc(0.1);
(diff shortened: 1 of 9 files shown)


slotsmanager.c
```diff
@@ -301,28 +295,32 @@ slots_gc_slots()
...
+        if(SlotsManager->info[ptype].garbage >= defrag)
+            disabled = 0;
     }
 
+    if(!disabled) {
+        slots_gc_mark();
+
+        for(ptype = 0; ptype < 6; ptype++) {
+            int defrag = 1 + defrag_frac * SlotsManager->info[ptype].size;
+            if(SlotsManager->info[ptype].garbage < defrag)
```
rainwoodman: It looks strange this is done twice.

slotsmanager.c
```diff
@@ -210,62 +232,34 @@ static int
 slots_gc_sweep(int ptype)
```
rainwoodman: sweep removes

slotsmanager.c
```diff
@@ -210,62 +232,34 @@ static int
...
+    for(i = 0; i < used; i++)
+        if(BASESLOT_PI(i,ptype)->IsGarbage)
+            slotBufcnt[sc++] = i;
+    slotBufcnt[sc] = used;
+    for(i = 0; i < sc; i++) {
+        int src = slotBufcnt[i]+1;
+        int dest = slotBufcnt[i] - i;
+        int nmove = slotBufcnt[i+1] - slotBufcnt[i]-1;
+//             message(1,"ptype = %d i = %d, PI = %d-> %d, nm=%d\n",ptype, i, src, dest, nmove);
+        /* TODO: in principle we can track this change and modify the tree nodes;
```
rainwoodman: So the variable slotBufcnt is the 'permutation' array between the old PI and the new PI of the slot -- could you write a comment 'slotBufcnt[i] = j where i is XXX and j is XXX'? I don't understand track changes and modifying the tree nodes: tree doesn't care about PI -- it only care about the base particle index.

slotsmanager.c
```diff
@@ -141,32 +141,41 @@ slots_gc(void)
...
+        for(i = 0; i < NumPart; i++)
+            if(P[i].IsGarbage) {
+                /* mark the particle for removal in base slot.*/
+                partBufcnt[pc++] = i;
+            }
+
+        /*Set final elements*/
+        partBufcnt[pc] = NumPart;
+
+        /*Compactify the P array: this invalidates the ReverseLink.*/
```
rainwoodman: .. invalidates the ReverseLink causing ReverseLink only valid within slots_gc_slots.

slotsmanager.c
```diff
@@ -346,7 +409,7 @@ slots_reserve(int atleast[6])
     if(SlotsManager->Base == NULL)
         SlotsManager->Base = (char*) mymalloc("SlotsBase", 0);
 
-    /* FIXME: change 0.005 to a parameter. The expericence is 
+    /* FIXME: change 0.005 to a parameter. The experience is
```
rainwoodman: 0.01, not 0.005..

slotsmanager.c
```diff
@@ -434,7 +497,10 @@ slots_mark_garbage(int i)
     P[i].IsGarbage = 1;
     if(SLOTS_ENABLED(P[i].Type)) {
         BASESLOT(i)->IsGarbage = 1;
+        atomic_add_and_fetch(&SlotsManager->info[P[i].Type].garbage, 1);
     }
+    /*Increment the counter*/
+    atomic_add_and_fetch(&SlotsManager->garbage, 1);
```
rainwoodman: Why count for garbage here?

It's usually more efficient to avoid atomic operation and count the number of garbages if we have to run though the full particle storage anyways during GC.

If all we need is whether there is any garbage, then we don't need an atomic write -- just set has_garbage to 1.

rainwoodman: In either case you do as many as +1 operations -- the atomic version is hell a lot worse than the non-atomic version.

slotsmanager.c
sbird: The TODO comment was left over from an earlier version - I will remove it.

slotsmanager.c
sbird: Fixed.

slotsmanager.c
sbird: The per-type atomics are there because we need to know if more than 10% of the slot is garbage to skip doing a gc.  The only other way to know that is to loop over the BASESLOT array. If we need to do that even if we are skipping the slot (which in practice we always are), gc becomes 50% slower and that completely dominates over the extra locking cost.

mark_slot_garbage is not called very often except during exchange, and the loop there is not threaded. Also when it is called there is another atomic access just before in fork, which will serialise.

 I can pretty easily avoid the overall garbage count though, if you think that is worth it. We could also just comment out the slots_gc_slots() - it is never called in practice because every PM step domain_decompose_full takes care of the garbage. What do you think?

rainwoodman: Indeed during tree walk, not often. Though I recall during exchange we might call this quite often -- since we mark all outgoing particles as garbage there. Though I am not sure how bad a performance hit it really will be since the exchange is likely a lot more expensive anyways. So I agree with you now on keeping the atomic locks.

Avoid having an overall garbage count reduces the amount of book keeping between GC cycles -- which sounds like a cleaner design -- remember the relief when we get rid of the book keeping code for TotNumPart and TimBinCount? However it appears that we need some of these to keep track if it is necessary to do a GC, so let's keep them. Could you change the name of variable to ngarbage to indicate it is the number of garbage and add a comment in the struct to say it is maintained by using mark() during cycles?

I am a bit supprised slots_gc_slots is no longer used. I need to digest this.

rainwoodman: The exchange algorithm uses gc to purge the out-going particles before receiving. So you can't really remove the routine -- otherwise we will run out of the slots quickly -- in cases where multiple iterations of exchange is required. 

Is the difference betweens slots_gc and slots_gc_sorted only that the latter sorts the particles before GC? If so perhaps it is cleaner to have a slots_sort_base(), and call both functions rather than having a larger function that does both.

sbird: We do call it often during exchange - but that loop is single-threaded, so there is zero contention.

slots_gc_sorted() does a sort of the particles, moving garbage to the end. It sorts the base array and then it sorts the slots array. It doesn't use the O(n) garbage algorithm in slots_gc() at all, because it reorders rather than just compacting. I think I will add a comment to this effect.

So slots_gc_slots() is no longer called often in normal circumstances because to be called we need > 10% of the particles exchanged between two PM timesteps, and the displacement constraint  means that this doesn't happen often. 

You make a good point though that under low memory conditions with multiple exchange iterations we do need it - the slots could use up all the memory otherwise. In fact, I will make it explicit that it always runs when multiple iterations happen, even if it would compact by < 10% of the slots.

sbird:
Ok. This new version took me far too long to figure out, but it avoids the counters. Instead it gc's the slots in an exchange only when we are short on memory. It also loops just once, so that it avoids allocating memory and is somewhat faster.

run.c
```diff
@@ -186,18 +186,20 @@ void run(void)
...
 
-            /* the accel may have created garbage -- collect them before checkpointing! */
-            force_tree_free();
-            slots_gc();
-            force_tree_rebuild();
+            /* The accel may have created garbage -- collect them before checkpointing!
+             * Tree will be auto-rebuilt if gc collected particles,
+             * but we should rebuild the active list.*/
+            int compact[6] = {0};
+            if(slots_gc(compact))
```
rainwoodman: `slots_gc(need_compact, &base_moved);`?

sbird: What does the base_moved argument do?

slotsmanager.c
```diff
@@ -127,46 +130,90 @@ slots_gc(void)
...
     tree_invalid |= slots_gc_base();
-    tree_invalid |= slots_gc_slots();
+    tree_invalid |= slots_gc_slots(compact_slots);
 
     MPI_Allreduce(MPI_IN_PLACE, &tree_invalid, 1, MPI_INT, MPI_SUM, MPI_COMM_WORLD);
 
-    EISlotsAfterGC event = {0};
+    EISlotsAfterGC event = {{tree_invalid}};
 
     event_emit(&EventSlotsAfterGC, (EIBase*) &event);
```
rainwoodman: What if we add rebuild_activelist as an event handler?

slotsmanager.c
```diff
@@ -127,46 +130,90 @@ slots_gc(void)
...
     return tree_invalid;
 }
 
+
+#define GARBAGE(i, ptype) (ptype >= 0 ? BASESLOT_PI(i,ptype)->IsGarbage : P[i].IsGarbage)
+#define PART(i, ptype) (ptype >= 0 ? (void *) BASESLOT_PI(i, ptype) : (void *) &P[i])
+
+/*Compaction algorithm*/
+static int
+slots_gc_compact(int used, int ptype, size_t size)
```
rainwoodman: It feels a bit strange using ptype < 0 for the base slots -- I sort of felt baseslots are different from secondary slots. Perhaps we shall have both inheriting from GCFriendly, which is a 4 bytes or 8 byte struct with IsGarbage flag?

run.c
rainwoodman: as a named return value ..

rainwoodman: But it sounds like rebuild_activelist can be an eventhandler registered by the timebin manager module. Then you get get rid of the call to rebuild_activelist here.

slotsmanager.c
sbird: We could - I'd rather not have it in a whole handler of it's own. The point of adding it is to ensure that when the tree gets rebuilt, the active list also gets rebuilt, so we should have them in the same function. But adding it to the existing handler would introduce a cross-module dependency of forcetree.c on timestep.c, and I'm not sure this is worth it, since at the moment this is only done in one place. What do you think?

sbird: Hmm, no, on second thoughts lets have it in its own handler.

rainwoodman:
Do we trigger the GC collectively on all ranks? I think we shall probably do that because if one rank hits GC, others will have to wait anyways.

slotsmanager.c
sbird: If it is in its own handler, then the active list will also be rebuilt after slot_gc in domain_exchange: so we would remove the rebuild_activelist call in the mainloop. But then we have a problem, which is that if the exchange does not generate garbage we would still need to rebuild the active list.

Also hiding something that is really a mainloop action in an event handler makes things harder to follow: working out what happens in event handlers is a bit tricky and over-using them can lead to crazy spaghetti code (parts of KDE in particular suffer from this). So I think on balance I prefer not to add it as an event handler.

slotsmanager.c
rainwoodman: This is better to be left on a future PR though.

slotsmanager.c
sbird: The better fix might be to move force_tree_rebuild back into the main loop and send the event handler back to being an error.

run.c
sbird: I can do this instead of the event handler, but returning the number of items moved is a fairly standard C library convention, so I quite like it like this.

sbird:
The base is collective, but the slot compaction is not - I will fix that.

slotsmanager.c
sbird: Yes - let's leave it like this for now and make it a future PR if we find ourselves doing this in more than one place.

slotsmanager.c
sbird: I have done this - the event handler is now gone.

rainwoodman: Sure. The event handlers are to decouple interaction between modules -- the missing part is probably a diagram that charts who emits what and who listen to what -- in UI toolkits it's been too frequently abused as a way to avoid the overhead cost of subclassing -- I suppose that's what makes you dislike Events.

I agree that if the mainloop already depend on all modules then there is no need to use it. I thought it might be cleaner to keep all forcetree logic in forcetree and timebin logic in timebin -- where a very thin mainloop does almost nothing. So another approach is to say the event handler has an argument about number of baseslots moved. Then we can decide if to rebuild the tree in the force tree handler, and decide to rebuild the activelist in the timebin handler.

slotsmanager.c
```diff
@@ -335,6 +346,72 @@ slots_gc_slots()
...
+            SlotsManager->info[ptype].size--;
+        }
+        slots_gc_collect(ptype);
+    }
+#ifdef DEBUG
+    slots_check_id_consistency();
+#endif
+
+    /*Rebuild the tree if still allocated*/
+    EISlotsAfterGC event = {{1}};
```
rainwoodman: If we no longer need the event -- then shall remove it as well.. the comment looked out dated.

exchange.c
```diff
@@ -165,19 +166,27 @@ static int domain_exchange_once(int (*layoutfunc)(int p), ExchangePlan * plan)
...
 
     ta_free(toGoPtr);
     walltime_measure("/Domain/exchange/makebuf");
 
-    /* now remove the garbage particles because they have already been copied.
-     * eventually we want to fill in the garbage gap or defer the gc, because it breaks the tree.
-     * invariance . */
+    /*Find which slots to gc*/
+    int compact[6] = {0};
+    for(ptype = 0; ptype < 6; ptype++) {
```
rainwoodman: Shall we make this a function, e.g. part of the plan?

slotsmanager.c
sbird: I don't dislike events per se - I think the current post-fork event is an excellent use of them, because it does remove cross-module dependencies and it describes something that could conceptually happen at any time in the code, so maps well onto what I understand an event to be. 

I just don't like over-using them: I saw once a mainloop which was implemented by several events chaining themselves together, which was just really hard to follow. Because they are attached at runtime it can become hard to follow which event handlers are actually running and debugging becomes tricky. I fixed a bug in nepomuk once where an event was being disconnected under peculiar circumstances deep within some other handler - it took me forever to figure it out, because I didn't have a way to find the function disconnecting it.

So if the main loop depends on all modules then it should explicitly state that in the header files - hiding the dependency in an event is just confusing.

One good reason for events is optional extensibility - I can see several of our feedback modules being implemented that way eventually. For example, SPH_GRAD_RHO could be replaced with an event handler registered by the H2 star formation criterion.

slotsmanager.c
sbird: Ok, done.

exchange.c
sbird: Done

sbird:
Travis is red but this seems to be a transient failure to install a dependent module (bigfile). Shall we merge?

rainwoodman:
Yes. Please merge.

The compilation error is because the new mpich requires the compiler
packages from conda in order to build binaries (because it was built with
the compilers from conda).

PR #167, opened 2017-11-26
Reordered

sbird:
A few minor cleanups from a flight. The only functional change is that the density computation gets moved earlier, so that the adaptive softening uses the updated smoothing length.

### DIFF
--- a/hydra.c
+++ b/hydra.c
@@ -237,13 +237,7 @@ hydro_ngbiter(
     {
         double Pressure_j = PressurePred(other);
         double p_over_rho2_j = Pressure_j / (SPHP(other).EOMDensity * SPHP(other).EOMDensity);
-        double soundspeed_j;
-
-#ifdef DENSITY_INDEPENDENT_SPH
-        soundspeed_j = sqrt(GAMMA * Pressure_j / SPHP(other).EOMDensity);
-#else
-        soundspeed_j = sqrt(GAMMA * p_over_rho2_j * SPHP(other).Density);
-#endif
+        double soundspeed_j = sqrt(GAMMA * Pressure_j / SPHP(other).EOMDensity);
 
         double dv[3];
         int d;
--- a/run.c
+++ b/run.c
@@ -41,7 +41,7 @@ enum ActionType {
 };
 static enum ActionType human_interaction(double lastPM, double TimeLastOutput);
 static int should_we_timeout(double TimelastPM);
-static void compute_accelerations(int is_PM, int FirstStep);
+static void compute_accelerations(int is_PM, int FirstStep, int GasEnabled);
 static void update_IO_params(const char * ioctlfname);
 static void write_cpu_log(int NumCurrentTiStep);
 
@@ -51,6 +51,8 @@ void run(void)
     int NumCurrentTiStep = 0;
     /*Minimum occupied timebin. Initially (but never again) zero*/
     int minTimeBin = 0;
+    /*Is gas physics enabled?*/
+    int GasEnabled = All.NTotalInit[0];
 
     /*To compute the wall time between PM steps and decide when to timeout.*/
     double lastPM = All.CT.ElapsedTime;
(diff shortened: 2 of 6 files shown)


hydra.c
```diff
@@ -237,13 +237,7 @@ hydro_ngbiter(
...
         double Pressure_j = PressurePred(other);
         double p_over_rho2_j = Pressure_j / (SPHP(other).EOMDensity * SPHP(other).EOMDensity);
-        double soundspeed_j;
-
-#ifdef DENSITY_INDEPENDENT_SPH
-        soundspeed_j = sqrt(GAMMA * Pressure_j / SPHP(other).EOMDensity);
-#else
-        soundspeed_j = sqrt(GAMMA * p_over_rho2_j * SPHP(other).Density);
-#endif
+        double soundspeed_j = sqrt(GAMMA * Pressure_j / SPHP(other).EOMDensity);
```
rainwoodman: This looks a bit suspicious -- doesn't the removal of ifdef break the density-energy sph force?

run.c
```diff
@@ -339,12 +339,29 @@ human_interaction(double TimeLastPM, double TimeLastOut)
  * be outside the allowed bounds, it will be readjusted by the function ensure_neighbours(), and for those
  * particle, the densities are recomputed accordingly. Finally, the hydrodynamical forces are added.
  */
-void compute_accelerations(int is_PM, int FirstStep)
+void compute_accelerations(int is_PM, int FirstStep, int GasEnabled)
```
rainwoodman: Mention that the subroutine assumes a valid tree before and after -- Perhaps add an assertion?

hydra.c
sbird: Did I misread it? I think the two expressions are the same, because EOMDensity is EgyWtDensity for DENSITY_INDEPENDENT_SPH and Density otherwise.

run.c
```diff
@@ -339,12 +339,29 @@ human_interaction(double TimeLastPM, double TimeLastOut)
...
+        message(0, "Start density computation...\n");
+
+        density();		/* computes density, and pressure */
+
+        /***** update smoothing lengths in tree *****/
+        force_update_hmax(ActiveParticle, NumActiveParticle);
+        /***** hydro forces *****/
+        message(0, "Start hydro-force computation...\n");
+
+        hydro_force();		/* adds hydrodynamical accelerations  and computes du/dt  */
```
rainwoodman: Looks like hydroforce can go after gravity? Gravity doesn't depend on it at all.

run.c
sbird: It is not valid after - starformation may invalidate the tree. Actually we should make this explicit.

run.c
```diff
@@ -383,35 +400,23 @@ void compute_accelerations(int is_PM, int FirstStep)
     if(FirstStep)
         grav_short_tree();
 
-    if(All.NTotalInit[0] > 0)
+    /* Note this must be after gravaccel and hydro,
+     * because new star particles are not in the tree,
```
rainwoodman: Can we move PM before all other forces? If so perhaps we can find a way to avoid PM depending on the tree, then defer tree build after PM on a PM step.

Be aware that we currently have to rebuild the tree twice in a PM step. I recall in Gadget-2/3 it was a big fuzz that the tree and PM are overlapped and tree is not built more than once per step. 

I am not sure how much of a gain it really is though.

hydra.c
rainwoodman: That's right -- I still remember there was some trickiness here. Perhaps it is integrated into pressurepred.

run.c
sbird: That isn't a big gain, true. But moving PM first might be a good idea anyway. Then the tree builds could all be explicitly in run.c/compute_accelerations() rather than hidden elsewhere.

sbird: The PM depends on the tree because of the accelerated Region assignment. Does that really help that much, since we have to rebuild the tree to do it?

run.c
sbird: It could - I just thought conceptually density + hydro go together and the sfr/bh go together too. But not very important.

run.c
rainwoodman: I think it can be modified to depend only on the TopTree (for domain). You can keep a call to build the tree and write a FIXME, till I make the change.

sbird: Ok, I will do that.

sbird: Oh, my mistake. We cannot move PM first because the tree opening angle uses GravPM from the last timestep.

rainwoodman: Uggh. Please add a comment so we don't run into this next time looking at the code! We can then merge the PR.

sbird: There is a comment! See line 369. I just hadn't read it, and there is no defence against that.

PR #169, opened 2017-12-04
split ioctl stuff to HCI(human-computer-interaction).

rainwoodman:
Updated timeout strategy that's likely more reliable:

1. queries occur on every PM step.
2. measure the longest time duration between two queries == longest
   time spent between two PM steps.
3. If there isn't enough wallclocktime between now and the limit
   to arrive to the next query, then we have a timeout.

### DIFF
--- a/hci.c
+++ b/hci.c
@@ -38,6 +38,8 @@ hci_init(HCIManager * manager, char * prefix, double WallClockTimeLimit, double 
     manager->LongestTimeBetweenQueries = 0;
 }
 
+/* override the result of hci_now; for unit testing -- we can't rely on MPI_Wtime there! 
+ * this function can be called before hci_init. */
 int
 hci_override_now(HCIManager * manager, double now)
 {
@@ -116,7 +118,7 @@ hci_query_timeout(HCIManager * manager)
         return NULL;
     }
 
-    /* any empty string would work. */
+    /* any freeable string would work. */
     return calloc(32, 1);
 }
 
@@ -128,6 +130,7 @@ hci_query_auto_checkpoint(HCIManager * manager)
 
     double now = hci_get_elapsed_time(manager);
     if(now - manager->TimeLastCheckPoint >= manager->AutoCheckPointTime) {
+        /* any freeable string would work */
         return calloc(32, 1);
     }
     return NULL;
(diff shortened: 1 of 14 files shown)


hci.c
```diff
@@ -0,0 +1,217 @@
...
+        message(0, "Stopping due to TimeLimitCPU, dumping a CheckPoint.\n");
+        action->type = HCI_TIMEOUT;
+        action->write_snapshot = 1;
+        free(request);
+        return 1;
+    }
+
+    if(request = hci_query_filesystem(manager, "ioctl"))
+    {
+        //update_IO_params(request);
```
sbird: This should not be commented!

hci.c
```diff
@@ -0,0 +1,217 @@
...
+static char *
+hci_query_auto_checkpoint(HCIManager * manager)
+{
+    /*How long since the last checkpoint?*/
+    if(manager->AutoCheckPointTime <= 0) return NULL;
+
+    double now = hci_get_elapsed_time(manager);
+    if(now - manager->TimeLastCheckPoint >= manager->AutoCheckPointTime) {
+        /* any freeable string would work */
+        return calloc(32, 1);
```
sbird: These two functions could return an int

rainwoodman: I wanted them to have the same interface -- int would break that.

hci.c
rainwoodman: I disabled it because I think this needs to be replaced by something more general, e.g. rerunning the configuration parser, and copy over the useful new parameters.

hci.c
sbird: In that case I would suggest having all functions return int, 1 == do something, 0 == do nothing, and hci_query_filesystem take an optional char * file_contents argument to store the file contents.

hci.c
sbird: I think that would be a different pull request. I don't use this feature personally, but it seems strange to disable it by a commented line without replacement. Adding a fixme comment would be better.

rainwoodman: Yes. It shall be a fixme. I added a warning too. I don't think anyone actually use this feature anymore.

hci.c
rainwoodman: Yes. I'll do that. Is it has to be a `char**`, though.

PR #170, opened 2017-12-04
Fixes

sbird:
A few fixes to bugs: the hdf5 reading should be malloc not mymalloc, there was a memory error, and an example file was broken at some point. Also fix runtests.


rainwoodman:
The last one is suggesting we haven't incorporated the knowledge that rebuild active list is needed after domain full properly. Perhaps we can't..

PR #174, opened 2017-12-09
Slotsfix

### LINKED ISSUE #171: Job terminated with signal 11 at PC=xxx error
Myself and @andreufont have had some Lya sims crash with the error:

`
MP-Gadget:33262 terminated with signal 11 at PC=3a96089f7e SP=7ffc8bfc3ba8.
`

The Makefile and full output files are here:
[test_err.txt](https://github.com/rainwoodman/MP-Gadget/files/1531637/test_err.txt)
[test_out.txt](https://github.com/rainwoodman/MP-Gadget/files/1531639/test_out.txt)
[Optionsmk.txt](https://github.com/rainwoodman/MP-Gadget/files/1531655/Optionsmk.txt)

If I increase PartAllocFactor to 4 the sims work ok, but these sims were running fine a few weeks ago with it set to 2. Should changing PartAllocFactor be necessary for the params (L=10Mpc, Ngrid=256) we are using?

Thanks!

sbird:
WIP: This makes some fixes to the slotsmanager which I do not think will actually fix #171.


begrun.c
```diff
@@ -51,13 +51,16 @@ void begrun(int RestartSnapNum)
...
 
     hci_init(HCI_DEFAULT_MANAGER, All.OutputDir, All.TimeLimitCPU, All.AutoSnapshotTime);
-    slots_init();
 
     petaio_init();
     walltime_init(&All.CT);
 
     petaio_read_header(RestartSnapNum);
 
+    int enabled[6] = {All.NTotalInit[0], 0, 0, 0, All.StarformationOn || All.NTotalInit[4], All.BlackHoleOn || All.NTotalInit[5]};
```
rainwoodman: It might be cleaner to use
```
slots_set_enabled(ptype, elsize)
```
after init.

slotsmanager.c
```diff
@@ -406,25 +406,29 @@ slots_gc_sorted()
...
 
 void
-slots_reserve(int atleast[6])
+slots_reserve(int atleast[6], int collective)
 {
     int newMaxSlots[6];
     int ptype;
     int good = 1;
+    /*Ensure all processors have initially the same number of particle slots*/
+    if(collective)
```
rainwoodman: It appears that other than the allreduce collective only affects the message output.. This will probably look awkward when we revisit it in a year or two.

sbird: What do you suggest?

begrun.c
sbird: But I would want enabled to be const-after-init if this were C++ :)

slotsmanager.c
rainwoodman: I couldn't think of anything better. Do we really need uncollective growth?

begrun.c
rainwoodman: We probably end up with the convention that struct members are read-only to the outside anyways. So it doesn't really matter. I am in general getting uncomfortable about having an array as an input of a function where the function is only doing a loop over the array (with no cross talk between iterations) -- the interface offers less flexibility than direct accepting the index and defering the loop to the caller.

sbird: I was tempted to read All.StarFormationOn, etc, inside of slots_init, but that makes the tests more messy. slots_set_enabled is a few more lines of code, but I can use it if you like.

slotsmanager.c
sbird: We do - some processors may be over-weight black holes and others may be over-weight stars. It is *collective* growth we don't really need.

begrun.c
rainwoodman: I'd say let's go with `slots_set_enabled` or s`lots_enable_type`, and be explicit about the `if-conditions`. It's verbose but cleaner.

And yes we shall try avoid using All.* variables in new code.

slotsmanager.c
rainwoodman: So we shall perhaps get rid of collective growth and wrap the message in DEBUG. for the special case where we explicitly need a collective growth then we do the allreduce before calling the function.

sbird:
There we go. The other thing we could do is to move *_particle_data into slotsmanager.h, since they are anyway only supposed to be accessed by the SPHP macros and those are already there.

sbird:
I did that move.

slotsmanager.c
sbird: How about do the allreduce before but keep an argument like "int where" which controls the message (same name as in message() and endrun())? Having the allocated message is fairly useful even if DEBUG is off (although we could reword it a little).

rainwoodman:
Sure. do you want to move the structs as well? Then we can close this. I feel we are close to a release. I did a few 2048 nbody runs and they finished like breeze.

sbird:
Already done (db2e807). Merging, and looking forward to a release indeed.

PR #175, opened 2017-12-12
Fix some undefined behaviour

sbird:
This was causing problems with icc for me. Found with gcc's undefined behaviour sanitizer.

### DIFF
--- a/exchange.c
+++ b/exchange.c
@@ -269,7 +269,7 @@ static int domain_exchange_once(int (*layoutfunc)(int p), ExchangePlan * plan)
             if(!SlotsManager->info[ptype].enabled) continue;
 
             if(BASESLOT(i)->ID != P[i].ID) {
-                endrun(1, "ID mismatched\n");
+                endrun(1, "Exchange: P[%d].ID = %ld != SLOT ID = %ld\n",i,P[i].ID, BASESLOT(i)->ID);
             }
         }
         for(ptype = 0; ptype < 6; ptype ++) {
--- a/exchange.c
+++ b/exchange.c
@@ -208,7 +208,7 @@ static int domain_exchange_once(int (*layoutfunc)(int p), ExchangePlan * plan)
     }
 
     if(newNumPart > All.MaxPart) {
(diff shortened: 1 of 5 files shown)


slotsmanager.c
```diff
@@ -412,8 +414,13 @@ slots_reserve(int where, int atleast[6])
     int ptype;
     int good = 1;
 
-    if(SlotsManager->Base == NULL)
+    if(SlotsManager->Base == NULL) {
         SlotsManager->Base = (char*) mymalloc("SlotsBase", 0);
+        /* This is so the ptr is never null! Avoid undefined behaviour. */
+        for(ptype = 5; ptype >= 0; ptype--) {
+            SlotsManager->info[ptype].ptr = SlotsManager->Base;
```
rainwoodman: Wouldn't it be better to set the ptr to NULL such that access on uninitialized .ptr  immediately lead to segfault?

exchange.c
```diff
@@ -269,7 +269,7 @@ static int domain_exchange_once(int (*layoutfunc)(int p), ExchangePlan * plan)
             if(!SlotsManager->info[ptype].enabled) continue;
 
             if(BASESLOT(i)->ID != P[i].ID) {
```
rainwoodman: the direct access of `.enabled` looked suspicious is it the same as the macro?

rainwoodman: It is the same.

sbird: Yes, but the macro is only in slotsmanager.c

slotsmanager.c
sbird: No, because of the memmove. The ptr is null at this point already, but we should not call memmove on a null pointer, even with size 0, because the compiler may optimise the pointer away.

sbird:
Alas, this does not fix my problem. We should merge it anyway, but I still have a crash (different to Chris')

slotsmanager.c
rainwoodman: We are supposed to avoid memmove on a null pointer by checking `enabled` flags in `_reserve`.

sbird: There is a hole when we reserve the first block: at that point even the enabled types are still NULL.

rainwoodman: I see.

PR #183, opened 2017-12-28
Globals

sbird:
Some more global variable cleanups. The main things are to remove the per-variable sfr counter, which was basically unused and to move struct particle_data into its own file, and to combine NumPart and All.MaxPart into a PartManager struct. It also enables some cleanups in the tree, but I am deferring them to a future PR. 

Everything here should make zero semantic difference - it is mostly just renames. Even so, since we are wanting a stable branch now, shall we create one, tag it as release-1 and merge this to master?

### DIFF
--- a/drift.c
+++ b/drift.c
@@ -9,6 +9,7 @@
 #include "timefac.h"
 #include "timestep.h"
 #include "slotsmanager.h"
+#include "partmanager.h"
 #include "endrun.h"
 
 static int drift_particle_full(int i, inttime_t ti1, int blocking);
--- a/drift.c
+++ b/drift.c
@@ -11,7 +11,6 @@
 #include "slotsmanager.h"
 #include "endrun.h"
 
-
 static int drift_particle_full(int i, inttime_t ti1, int blocking);
 
 static void real_drift_particle(int i, inttime_t ti1);
@@ -27,9 +26,6 @@ void drift_particle(int i, inttime_t ti1) {
 int drift_particle_full(int i, inttime_t ti1, int blocking) {
     if(P[i].Ti_drift == ti1) return 0 ;
 
-#pragma omp atomic
-    TotalParticleDrifts ++;
-
 #ifdef OPENMP_USE_SPINLOCK
     int lockstate;
     if (blocking) {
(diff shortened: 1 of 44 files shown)


drift.c
```diff
@@ -62,8 +56,6 @@ int drift_particle_full(int i, inttime_t ti1, int blocking) {
     {
         if(P[i].Ti_drift != ti1) {
             real_drift_particle(i, ti1);
-        } else {
-            BlockedParticleDrifts ++;
```
rainwoodman: yes get rid of these!

forcetree.c
```diff
@@ -989,8 +991,8 @@ void force_update_hmax(int * activeset, int size)
     /* FIXME: actually only TOPLEVEL nodes contains the local mass can potentially be dirty,
      *  we may want to save a list of them to speed this up.
      * */
-    for(i = All.MaxPart; i < All.MaxPart + MaxNodes; i ++) {
-        NodeIsDirty[i - All.MaxPart] = 0;
+    for(i = 0; i < MaxNodes; i ++) {
+        NodeIsDirty[i] = 0;
```
rainwoodman: Is this the only variable per node? Did we settle on names for SingleParticleNode/TerminalNode and TreeNode? Perhaps consider start making this a struct member here?

types.h
```diff
@@ -5,6 +5,9 @@
 
 /*Define some useful types*/
 
+#define MPI_UINT64 MPI_UNSIGNED_LONG
+#define MPI_INT64 MPI_LONG
```
rainwoodman: I recall in a few places MPI_LONG and MPI_LONG_LONG are still used instead MPI_INT64 / MPI_UINTT64.

Perhaps also define MPI_PTRDIFF and MPI_SIZET and replace them all?

forcetree.c
sbird: Yes, that one can be cleaned up substantially. I'll make a note though and defer it to a longer cleanup of the tree - because these commits split All.MaxPart from MaxNodes it enables some more tree cleanups. I also want to make the tree build more thread-scalable.

types.h
sbird: Should it be #define MPI_INT64 MPI_LONG_LONG ? I don't think MPI_LONG is guaranteed to be 64-bit

sbird: I will leave it as a comment - we are probably more likely to run on a 128-bit architecture than a 32-bit one right now.

sbird:
What do you think about making a stable branch before merging this?

rainwoodman:
No objections here. I can do it tonight.

PR #186, opened 2018-01-03
Driftentropy

sbird:
This pull request evolves entropy at drift time instead of kick time. It also fixes multiple bugs in DENSITY_INDEPENDENT_SPH that I found along the way. 

These should definitely go to stable.
e692317 - fixes a bug that the entropy change is limited in EntropyPred. This is silly because it also applies to backward predictions and thus limits the heating that can happen.
7fdcda0 - This moves the change in the units of DtEntropy into reduction rather than postprocessing. This is because EntropyPred = Entropy + DtEntropy * Dt is used to generate DtEntropy for neighbouring particles. If this happened after DtEntropy was reduction but before postprocessing the units of Dtentropy were different and so EntropyPred would give a wrong result. Also for stable.

3e1b223 and earlier commits change the Entropy to be computed at drift time. Because of the above, computing it at kick time means that EntropyPred gives a different result depending on whether DtEntropy was updated yet, ie, EntropyPred and thus DtEntropy for a given particle depends on whether the treewalk was already run for a neighbour particle (because it does prediction using ither DtEntropy from the current or previous step). So our output becomes dependent on the number of threads, which is not good. This should probably go to stable, maybe in a squashed version as the intermediate commits won't apply.

The final two commits are cleanups which should make no difference in practice.

### DIFF
--- a/hydra.c
+++ b/hydra.c
@@ -138,7 +138,7 @@ hydro_copy(int place, TreeWalkQueryHydro * input, TreeWalk * tw)
     input->Density = SPHP(place).Density;
 #ifdef DENSITY_INDEPENDENT_SPH
     input->EgyRho = SPHP(place).EgyWtDensity;
-    input->EntVarPred = EntropyPred(place);
+    input->EntVarPred = pow(EntropyPred(place), 1/GAMMA);
     input->DhsmlDensityFactor = SPHP(place).DhsmlEgyDensityFactor;
 #else
     input->DhsmlDensityFactor = SPHP(place).DhsmlDensityFactor;
@@ -304,7 +304,7 @@ hydro_ngbiter(
 #ifdef DENSITY_INDEPENDENT_SPH
         double hfc = hfc_visc;
         /* leading-order term */
-        double EntPred = EntropyPred(other);
+        double EntPred = pow(EntropyPred(other), 1/GAMMA);
         hfc += P[other].Mass *
             (dwk_i*iter->p_over_rho2_i*EntPred/I->EntVarPred +
              dwk_j*p_over_rho2_j*I->EntVarPred/EntPred) / r;
--- a/hydra.c
+++ b/hydra.c
@@ -25,6 +25,13 @@
  *  computed, and where the rate of change of entropy due to the shock heating
  *  (via artificial viscosity) is computed.
  */
+
+double
+Pressure(int i)
+{
+    return pow(SPHP(i).Entropy, 1/GAMMA) * pow(SPHP(i).EOMDensity, GAMMA);
+}
+
 typedef struct {
     TreeWalkQueryBase base;
 #ifdef DENSITY_INDEPENDENT_SPH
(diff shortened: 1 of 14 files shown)


rainwoodman:
If the entropy was synced with Kick then 7fdce is not as bad of an issue is
it?

sbird:
If the Entropy is synced to *drift* then 7fdce is not an issue. If it is synced to Kick it is.

rainwoodman:
Now I see the problem -- it is because we are no longer saving entvarpred as a variable. Therefore, as we update DtEntropy with the treewalk some particles would have had partially updated DtEntropy -- it would be wrong even with this fix.

I therefore think if we don't get rid of EntropyPred() as a function the problem will still be there, as long as we modify DtEntropy. This may affect other places where we use a variable that is being updated in the prediction.

The bluetides-ii and bluetides-i branch do not suffer from this as they were tagged before EntropyPred() is converted to a function.

hydra.c
```diff
@@ -138,13 +145,11 @@ hydro_copy(int place, TreeWalkQueryHydro * input, TreeWalk * tw)
...
-    input->EntVarPred = EntropyPred(place);
-    input->DhsmlDensityFactor = SPHP(place).DhsmlEgyDensityFactor;
-#else
-    input->DhsmlDensityFactor = SPHP(place).DhsmlDensityFactor;
+    input->EntVarPred = pow(SPHP(place).Entropy, 1/GAMMA);
 #endif
+    input->DhsmlEOMDensityFactor = SPHP(place).DhsmlEOMDensityFactor;
 
-    input->Pressure = PressurePred(place);
+    input->Pressure = Pressure(place);
```
rainwoodman: I feel we want to add a prefix to these function(Pressure / Entropy) -- at least add a verb,  GetPressure() to avoid potential conflicts with variable names.

sfr_eff.c
```diff
@@ -375,9 +375,10 @@ cooling_direct(int i) {
 
     double ne = SPHP(i).Ne;	/* electron abundance (gives ionization state and mean molecular weight) */
 
-    double unew = DMAX(All.MinEgySpec,
-            (SPHP(i).Entropy + SPHP(i).DtEntropy * dloga) /
-            GAMMA_MINUS1 * pow(SPHP(i).EOMDensity * All.cf.a3inv, GAMMA_MINUS1));
+    /*Entropy at kick time after this timestep: note this uses this timestep's DtEntropy!*/
+    const double KickEntropy = SPHP(i).Entropy + SPHP(i).DtEntropy * dloga/2;
```
rainwoodman: There wasn't a /2 on the left hand side.

sfr_eff.c
```diff
@@ -449,10 +450,12 @@ static int get_sfr_condition(int i) {
         flag = 1;		/* only normal cooling for particles in the wind */
 
     if(All.QuickLymanAlphaProbability > 0) {
-        double dloga = get_dloga_for_bin(P[i].TimeBin);
+        const double dloga = get_dloga_for_bin(P[i].TimeBin);
+        /*Entropy at kick time after this timestep: note this uses this timestep's DtEntropy!*/
+        const double KickEntropy = SPHP(i).Entropy + SPHP(i).DtEntropy * dloga/2;
```
rainwoodman: You mean predicting Entropy at kick time of the current drift time.

rainwoodman:
I tend to merge the entire PR to master because it does fix a bug in a clean way. I haven't tagged a 5.0.0 release yet anyways. I am thinking of going with periodical releases -- which makes no difference between stable and master.

sbird:
Ah yes. So this is a bug I introduced, which is reassuring.

sfr_eff.c
sbird: But that is because the Entropy was formerly at the Kick time, ie, t - dt/2 and is now at drift time, t.

sfr_eff.c
sbird: Yes, indeed.

hydra.c
sbird: Ok.

rainwoodman:
Yes. 7fdce really scared me since we are >> half way on BT-ii allocation and some of the troubles of Blackhole repositioning were pointing to too low of a sound speed than usually seen, and the soundspeed  comes from pressure / entropy... Glad to know it is not due to such obvious problem at least.

You'll do a rename and we can then merge.

sfr_eff.c
rainwoodman: Problem: I am still confused by the 1/2. 

After a drift the position of particles are t_d. the velocity are t_k = t_d - 1/2. now we compute the force at t_d, then we use force to kick velocity to t_k = t_d. then we recompute step size `alpha`, kick velocity to t_k + `alpha` * 0.5. next we will drift particles to t_k + `alpha`.

We assert Entropy is synced with position. Therefore, drift uses DtEntropy to update Entropy. force calculation requires the entropy variable predicted in the future. However I think we shall use t_d + 1, not t_d + 1/2. For two reasons:
1. notice a few lines later the 'modulated' DtEntropy is divided by dloga, not 0.5 dloga.
2. the BH feedback injected energy shall be deposited in the duration of a full time step, not a half time step (which would increase the BH heating rate, messing up the parameter calibration).

rainwoodman: In short I argue that the Entropy equation doesn't go with leap-frog. It is an implicit-euler, so there is no 0.5.

sbird: Yes, you are correct. Well noticed!

rainwoodman: Fix it and we merge the last three PRs to deliver a 'good' version for #171?

PR #187, opened 2018-01-05
add -DSTACKTRACE; print stack trace with an external tool.

rainwoodman:
When no tools exist it will fallback to glibc's backtrace.

The code is largely modelled from xorg-server's backtrace and osinit.

### DIFF
--- a/Options.mk.example
+++ b/Options.mk.example
@@ -18,7 +18,9 @@ GSL_LIBS = $(shell pkg-config --libs gsl)
 #--------------------------------------- Basic operation mode of code
 OPT += -DDENSITY_INDEPENDENT_SPH
 #OPT += -DLIGHTCONE                       # write a lightcone on the fly; in development
-#OPT += VALGRIND  # allow debugging with valgrind, disable the GADGET memory allocator.
+#OPT += VALGRIND     # allow debugging with valgrind, disable the GADGET memory allocator.
+#OPT += -DDEBUG      # print a lot of debugging messages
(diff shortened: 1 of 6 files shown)


rainwoodman:
Yes. I am supposed to be the expert on these things. Thanks for checking though.

rainwoodman:
But after merging it I felt it shall probably be converted to a configuration flag since there is no external dependency.

sbird:
Why don't we have @chrispederson test it, and if it works, just remove the config flag so this feature is always on? Having the ability to turn it off just increases the probability of a crash without a decent stack trace.

rainwoodman:
Sometimes the process is run inside a debugger or crash analyzing tool provided by the vendor that produces far more comprehensible stacktraces. But the signal handler will interfere with these tools -- a concern is launching a few hundred K ptrace processes at the same time at the point of crash may get stuck. So we definitely need a parameter to disable this at least.

sbird:
I don't think launching ptrace processes is a real problem - we are crashing anyway if we get here, so we can't make things much worse.

Also, is there not a way to detect at runtime whether a program is run inside a debugger? If not, it does need to be a runtime thing, so you can disable it without recompiling and potentially moving the bug...

rainwoodman:
It becomes a real problem if pstack deadlocks give a stack trace, and if rerunning is too expensive. It's better to observe the behavior and fix it if it becomes problematic. Did you try to crash the code (e.g. with a 1/0 error) in GDB with -DSTACKTRACE? If not I'll give it a go.

sbird:
I didn't (I didn't know gdb could do that), so go ahead.

PR #188, opened 2018-01-05
Introduce a Policy object for domain decomposition.

rainwoodman:
We will try a few policies before bailing out, with increasing
computational cost and decreasing accuracy.

This PR fixes a bug where the TopTree split may underflow the 'shift' variable, which may be causing #171 .

### DIFF
--- a/domain.c
+++ b/domain.c
@@ -177,7 +177,7 @@ void domain_decompose_full(void)
     }
 
     if(decompose_failed) {
-        endrun(0, "Not suitable domain decomposition policy worked for this particle distribution\n");
+        endrun(0, "No suitable domain decomposition policy worked for this particle distribution\n");
     }
 
     domain_balance();
--- a/domain.c
+++ b/domain.c
@@ -861,9 +861,15 @@ domain_toptree_garbage_collection(struct local_topnode_data * topTree, int start
     topTree[start].Daughter = newd;
 
     (*last_free) += 8;
+
+    /* copy first in case oldd + j is overwritten by the recursed gc
+     * if last_free is less than oldd */
     for(j = 0; j < 8; j ++) {
         topTree[newd + j] = topTree[oldd + j];
         topTree[newd + j].Parent = start;
+    }
+
+    for(j = 0; j < 8; j ++) {
         domain_toptree_garbage_collection(topTree, newd + j, last_free);
     }
 }
--- a/domain.c
+++ b/domain.c
@@ -122,7 +122,6 @@ static int domain_allocated_flag = 0;
 static int
 domain_policies_init(DomainDecompositionPolicy policies[],
         const int NincreaseAlloc,
-        const int NdecreaseAccuracy,
         const int SwitchToGlobal);
 
 /*! This is the main routine for the domain decomposition.  It acts as a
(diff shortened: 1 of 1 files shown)


sbird:
What you have done looks reasonable, but can you explain more which bit exactly fixes the underflow? Is it the check for Nsample == 0?

rainwoodman:
It is the check for shift.

sbird:
Forgive me if this is obvious, but why do we hit that condition? I thought Shift small meant a very overdense environment? Is it just because @chrispederson is running a small simulation on a lot of nodes, so there is likely to be a very sparsely populated toptree?

rainwoodman:
No This is not obvious at all. There Might be another bug that causes shift
of child and parent to be the same.

On Jan 5, 2018 8:38 AM, "Simeon Bird" <notifications@github.com> wrote:

Forgive me if this is obvious, but why do we hit that condition? I thought
Shift small meant a very overdense environment? Is it just because
@chrispederson <https://github.com/chrispederson> is running a small
simulation on a lot of nodes, so there is likely to be a very sparsely
populated toptree?

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/rainwoodman/MP-Gadget/pull/188#issuecomment-355601417>,
or mute the thread

domain.c
```diff
@@ -131,26 +148,37 @@ void domain_decompose_full(void)
...
+            LastSuccessfulPolicy = i;
+            break;
         }
-    } while(decompose_failed);
+        domain_free();
+
+    }
+
+    if(decompose_failed) {
+        endrun(0, "Not suitable domain decomposition policy worked for this particle distribution\n");
```
sbird: No suitable domain decomposition

domain.c
```diff
@@ -215,12 +243,46 @@ void domain_maintain(void)
     force_tree_rebuild();
 }
 
+/* this function generates several domain decomposition policies for attempting
+ * creating the domain. */
```
sbird: Longer comment needed here.

rainwoodman:
I think I found that bug in 96cc3af.

domain.c
```diff
@@ -215,12 +243,46 @@ void domain_maintain(void)
...
+        policies[i].UseGlobalSort = All.DomainUseGlobalSorting;
+        policies[i].PreSort = 0;
+        policies[i].SubSampleDistance = 16;
+    }
+
+    for(i = NincreaseAlloc; i < NincreaseAlloc + NdecreaseAccuracy; i ++) {
+        policies[i].TopNodeAllocFactor = policies[NincreaseAlloc].TopNodeAllocFactor;
+        policies[i].UseGlobalSort = All.DomainUseGlobalSorting;
+        policies[i].PreSort = 0;
+        policies[i].SubSampleDistance = 16 * (i - NincreaseAlloc);
```
sbird: I don't understand this line - why would increasing the sampling distance make a formerly failing domain decomposition succeed?

domain.c
```diff
@@ -215,12 +243,46 @@ void domain_maintain(void)
...
+    }
+
+    for(i = SwitchToGlobal; i < NincreaseAlloc + NdecreaseAccuracy; i ++) {
+        /* global sorting of particles is slow, so we add a presort to make the particle 
+         * distribution locally even to improve the balance */
+        policies[i].UseGlobalSort = 1;
+        policies[i].PreSort = 1;
```
sbird: This adds a presort and global sort at the same time as decreasing the accuracy. Is that what you want?

sbird:
Aha, very possible.

domain.c
rainwoodman: I don't think it will do, though I thought so. I'll change this back to just increasing TopNodeAllocFactor for now.

domain.c
rainwoodman: The comment above meant to say that since global sorting is slow anyway, a presort can always be added without making it much slower.

PR #189, opened 2018-01-10
Fix growth function in the presence of massive neutrinos

sbird:
Neutrinos are usually free-streaming at simulation start over the scales of interest
and so need to be omitted from the source term of the growth function.

We also warn in case the box is large enough that free-streaming is not
a good approximation, which is most of the code.

This fixes a porting bug from S-GenIC and warns in a case where our ICs may be inaccurate at the percent level.

### DIFF
--- a/begrun.c
+++ b/begrun.c
@@ -189,7 +189,7 @@ set_units(void)
     /* convert some physical input parameters to internal units */
 
     All.CP.Hubble = HUBBLE * All.UnitTime_in_s;
-    init_cosmology(&All.CP, All.TimeInit);
+    init_cosmology(&All.CP, All.TimeInit, All.BoxSize * All.UnitLength_in_cm);
     /*Initialise the hybrid neutrinos, after Omega_nu*/
     if(All.HybridNeutrinosOn)
         init_hybrid_nu(&All.CP.ONu.hybnu, All.CP.MNu, All.HybridVcrit, C/1e5, All.HybridNuPartTime, All.CP.ONu.kBtnu);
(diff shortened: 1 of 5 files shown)


andreufont:
Hi @sbird - @Chris-Pedersen and I are very interested in understanding these changes. Is this something that would affect our previous neutrino runs? 

I'm a bit confused, since once you have more than one fluid, the growht factor does not have a clear meaning, does it? 

I'm not sure either where the growth factor is used. To generate initial conditions we use the transfer function for each species, evaluated at the initial redshift, right? So no need for a growth factor there. And when running the simulation, I don't see where the growth factor would be used at all. 

Is this related to the non-neutrino simulatios, where you could define the initial power spectrum at z=0 and then rescale the power to an initial redshift, using the growth factor?

sbird:
Yes, this will affect existing neutrino runs using MP-GenIC (but only slightly: at the percent level or less). We use the derivative of the growth factor to relate the particle velocities to particle displacements in the initial conditions:  v = a H dln D1/dlna x For multiple fluids it is indeed not quite right - we need to change it to solving the perturbation theory equations directly at some point.

andreufont:
I see. Something important to keep in mind, @Chris-Pedersen.

@sbird - The effect should be proportional to the neutrino mass assumed, right? I mean, if we run a simulation with neutrino mass = 0.0001 eV, the effect would not a the percent level, right?

sbird:
yes, it is ~ 1% for M_nu ~ 0.5 eV. If you check that your simulations match camb on large scales you'll be fine.

PR #191, opened 2018-01-11
Restore EntVarPred from bluetides-ii

sbird:
This restores the EntVarPred variable from bluetides-ii and thus fixes the various entropy handling bugs found in the driftentropy branch, without breaking anything. Some of the cleanups from the driftentropy branch are also included, as is the neutrino bugfix. I checked it runs to z=2 this time.


cosmology.c
```diff
@@ -36,8 +36,19 @@ void init_cosmology(Cosmology * CP_in, const double TimeBegin)
...
         CP->OmegaCDM -= get_omega_nu(&CP->ONu, 1);
+        /* Warn if we have a large enough box that neutrinos are not free-streaming on the largest scales.
+         * Free-streaming scale is ~500 Mpc at z=99 for M_nu = 0.3 eV.
+         * FIXME: Implement scale dependent growth correction, by differentiating transfer functions.*/
+        /*Neutrino thermal velocity in cm/s*/
+        double v_th = BOLEVK*(TNUCMB*CP->CMBTemperature)/(CP->MNu[0] + CP->MNu[1] + CP->MNu[2])*3. / TimeBegin * C;
+        /*Neutrino free-streaming length from Lesgourgues & Pastor in (comoving) cm/h*/
+        double nufs = 2 * M_PI * sqrt(2/3.) * v_th / (hubble_function(TimeBegin)/CP->Hubble * HUBBLE)/TimeBegin;
+        /*Box should be in comoving cm/h: we ask for 1/4 of the box because sample variance will matter on larger scales.*/
+        if(nufs < box/4.)
```
rainwoodman: Shall we lift the logic out of cosmology by adding a function that gets the nu free streming scale of any redshift / a?

sbird: This is only temporary - we will implement scale-dependent growth functions (also for the baryons)

rainwoodman:
Are you planning to use the fitting forms or the solutions from a true
Boltzmann code, or are you planning to embed a 3fl solver?

sbird:
Haven't decided yet - one possibility is to embed CLASS, one possibility is to numerically differentiate transfer functions, one other is to solve the perturbation theory equations directly (the last would enable relative velocity support, ala O'Leary & McQuinn 2012, which I think CLASS does not do yet)

sbird:
I think the remaining "particle coupling" with different transfer functions is at least partly due to the velocity prefactors being the same for baryons and CDM: they should be a bit different.

rainwoodman:
Say we write a solver, then shall it be non-relativistic or relativistic? NR 3fl is easy but we cannot initialize it properly -- it won't match the linear scale growth unless we start with a velocity at a->0. Relativistic solver looked quite nasty -- at least the CLASS paper looked so.

For numerically differentiating transfer function, we shall provide a python script to generate the file from class(our classylss binding makes it easy). But then we need to interpolate in MP-Gadget.

@energy-berkeley played around with both a while ago. I wonder if he's back and still recall any wisdom to share the issue. 

I suspect it is not due to the velocity prefactors. don't we already shift them by half a mesh in the IC?

sbird:
The solver I was thinking of was along the lines of: http://faculty.washington.edu/mcquinn/Init_Cond_Code.html which uses CAMB at z=1100 for initialisation. Differentiating transfer functions is easier though and would only need one FFT per species (I think).

We shift the particles by half a mesh, but the growth function d ln D/dlna should be slightly different for baryons and CDM (as well as scale-dependent). Because baryons are needing to 'catch up' to the CDM they grow slightly faster, and thus should have slightly larger initial velocities.

PR #193, opened 2018-01-19
Small fixes

sbird:
The main thing here is to reintroduce the entropy limiter in predicted entropy from before (but only apply it when cooling). It does fix a real (but very rare) bug where the predicted entropy can become negative.

Also a couple of trivial cleanups.

### DIFF
--- a/drift.c
+++ b/drift.c
@@ -137,7 +137,11 @@ static void real_drift_particle(int i, inttime_t ti1)
         /* Evolve entropy at drift time: evolved dlog a.
          * Used to predict pressure and entropy for SPH*/
         double dloga = dloga_from_dti(P[i].Ti_drift - P[i].Ti_kick);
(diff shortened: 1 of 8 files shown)


drift.c
```diff
@@ -137,7 +137,11 @@ static void real_drift_particle(int i, inttime_t ti1)
         /* Evolve entropy at drift time: evolved dlog a.
          * Used to predict pressure and entropy for SPH*/
         double dloga = dloga_from_dti(P[i].Ti_drift - P[i].Ti_kick);
-        SPHP(i).EntVarPred = pow(SPHP(i).Entropy + SPHP(i).DtEntropy * dloga,1/GAMMA);
+        SPHP(i).EntVarPred = SPHP(i).Entropy + SPHP(i).DtEntropy * dloga;
+        /*Entropy limiter for the predicted entropy: makes sure entropy stays positive. */
```
rainwoodman: ... doesn't decrease too fast, and stays positive.

PR #194, opened 2018-01-20
Openmpmoments

sbird:
Thread the second phase of the tree build procedure, the moments calculation. This is done using a split/merge algorithm. It doesn't work particularly well with the treebuild test (because those are deliberately pathological setups) , but it scales roughly linearly with thread number up to at least 8 threads on a real simulation.

Testing done:
- dm-only example to z=3 with 8 threads and 4 mpis / 2 threads (on a hyperthreaded quad-core), and I checked the final output was the same for master and this branch in both cases. Note that the output did differ slightly between the 8 threads and 4 mpis/2 threads tests, which may need investigating.

With this the % time spent in tree build goes from 11% in the 8 thread example to 6%.

### DIFF
--- a/forcetree.c
+++ b/forcetree.c
@@ -707,8 +707,12 @@ force_get_sibling(int sib, int j, int * suns)
  *  and all its subnodes using a recursive computation.  The result is
  *  stored in tb.Nodes in the sequence of this tree-walk.
  *
- *  The function also computes the NextNode and sibling linked lists. The return value
- *  and argument tail is the current tail of the NextNode linked list.
+ *  The function also computes the NextNode and sibling linked lists.
+ *  The return value is the current tail of the NextNode linked list.
+ *
+ *  This function is called recursively using openmp tasks.
+ *  We spawn a new task for a fixed number of levels of the tree.
+ *
  */
 static int
 force_update_node_recursive(int no, int sib, int level, const struct TreeBuilder tb)
@@ -814,8 +818,8 @@ force_update_node_recursive(int no, int sib, int level, const struct TreeBuilder
     }
 
     /*This loop sets the next node value for the row we just computed.
-     Tails at this point contains: no, (<=8 return values of force_update_node_recursive).
-     The last entry needs to be the return value of this function.*/
+      Note that tails[i] is the next node for suns[i-1].
+      The the last tail needs to be the return value of this function.*/
     int tail = no;
     for(j = 0; j < 8; j++)
     {
@@ -832,14 +836,10 @@ force_update_node_recursive(int no, int sib, int level, const struct TreeBuilder
 }
 
 /*! This routine determines the multipole moments for a given internal node
- *  and all its subnodes in parallel, assigning the recursive algorithm to different threads.  The result is
- *  stored in tb.Nodes in the sequence of this tree-walk.
+ *  and all its subnodes in parallel, assigning the recursive algorithm to different threads using openmp's task api.
+ *  The result is stored in tb.Nodes in the sequence of this tree-walk.
  *
- *  The function also computes the NextNode and sibling linked lists. The return value
- *  and argument tail is the current tail of the NextNode linked list.
- *
- * The parallel algorithm is as follows:
- * - The tree is walked a number of levels (three by default) down from each local topleaf. Local topleaves are used
+ * - A new task is spawned from each  down from each local topleaf. Local topleaves are used
  * so that we do not waste time trying moment calculation with pseudoparticles.
  * - Each internal node found at that level is added to a list, together with its sibling.
  * - Each node in this list then has the recursive moment calculation called on it.
--- a/forcetree.c
+++ b/forcetree.c
@@ -205,6 +205,7 @@ static void init_internal_node(struct NODE *nfreep, struct NODE *parent, int sub
     nfreep->len = 0.5 * parent->len;
     nfreep->f.TopLevel = 0;
     nfreep->f.InternalTopLevel = 0;
+    nfreep->f.MomentsDone = 0;
 
     for(j = 0; j < 3; j++) {
         /* Detect which quadrant we are in by testing the bits of subnode:
@@ -352,6 +353,7 @@ int force_tree_create_nodes(const struct TreeBuilder tb, const int npart)
         nfreep->father = -1;
         nfreep->f.TopLevel = 1;
         nfreep->f.InternalTopLevel = 0;
+        nfreep->f.MomentsDone = 0;
         nnext++;
         /* create a set of empty nodes corresponding to the top-level domain
          * grid. We need to generate these nodes first to make sure that we have a
@@ -963,7 +965,7 @@ void force_exchange_pseudodata(void)
         }
     }
 
-    /* share the pseudo-particle data accross CPUs */
+    /* share the pseudo-particle data across CPUs */
 
     recvcounts = (int *) mymalloc("recvcounts", sizeof(int) * NTask);
     recvoffset = (int *) mymalloc("recvoffset", sizeof(int) * NTask);
--- a/forcetree.c
+++ b/forcetree.c
@@ -687,7 +687,7 @@ add_particle_moment_to_node(struct NODE * pnode, int i)
 
 /*Get the sibling of a node, using the suns array. Only to be used in the tree build, before update_node_recursive is called.*/
 static int
-force_get_sibling(int sib, int j, int * suns)
+force_get_sibling(const int sib, const int j, const int * suns)
 {
     /* check if we have a sibling on the same level */
     int jj;
@@ -752,7 +752,7 @@ force_update_node_recursive(int no, int sib, int level, const struct TreeBuilder
              *or if we are deep enough that we already spawned a lot.
              Note: final clause is much slower for some reason. */
             if(chldcnt > 1 && level < 513) {
-                #pragma omp task shared(tails)
+                #pragma omp task default(none) shared(tails, level, chldcnt) firstprivate(j, nextsib, p)
                 tails[j] = force_update_node_recursive(p, nextsib, level*chldcnt, tb);
             }
             else
--- a/forcetree.c
+++ b/forcetree.c
@@ -862,6 +862,10 @@ force_get_sibling(int curnode, const struct TreeBuilder *tb)
     return force_get_sibling(parent,tb);
 }
 
+/* This function adds an internal node to the list of nodes to have their moments calculated recursively
+ * in the parallel (first) step of the algorithm. When handed a node in 'no' it will recurse down levels until level == 0.
+ * It will then compute the sibling for each internal node attached to the current parent, skipping particles and pseudoparticles,
+ * and add the new node to the list in TaskNodes.*/
 static void
 force_queue_refinement(int no, int level, int * index, struct TaskNode *TaskNodes, const int nmax, const struct TreeBuilder *tb)
 {
@@ -890,6 +894,16 @@ force_queue_refinement(int no, int level, int * index, struct TaskNode *TaskNode
  *
  *  The function also computes the NextNode and sibling linked lists. The return value
  *  and argument tail is the current tail of the NextNode linked list.
+ *
+ * The parallel algorithm is as follows:
+ * - The tree is walked a number of levels (three by default) down from each local topleaf. Local topleaves are used
+ * so that we do not waste time trying moment calculation with pseudoparticles.
+ * - Each internal node found at that level is added to a list, together with its sibling.
+ * - Each node in this list then has the recursive moment calculation called on it.
+ * Note: If the tree is very unbalanced and one branch much deeper than the others, this will not be efficient.
+ * - Once each tree's recursive moment is generated in parallel, the tail value from each recursion is stored, and the node marked as done.
+ * - A final recursive moment calculation is run in serial for the top 3 levels of the tree. When it encounters one of the pre-computed nodes, it
+ * searches the list of pre-computed tail values to set the next node as if it had recursed and continues.
  */
 int
 force_update_node_parallel(const struct TreeBuilder tb)
@@ -927,6 +941,7 @@ force_update_node_parallel(const struct TreeBuilder tb)
 
     /* now compute the multipole moments recursively for each subtree*/
     if(nfound > 0) {
+        /* Dynamic scheduling is used as it is common for one or two tree branches to be deep.*/
         #pragma omp parallel for schedule(dynamic,1)
         for(i = nfound-1; i >= 0; i--)
         {
--- a/forcetree.c
+++ b/forcetree.c
@@ -716,21 +716,21 @@ compare_nodes(const void * a, const void * b)
  *  and argument tail is the current tail of the NextNode linked list.
  */
 static int
-force_update_node_recursive(int no, int sib, int tail, const struct TaskNode * PreComp, int ntask, const struct TreeBuilder tb)
+force_update_node_recursive(int no, int sib, int tail, const struct TaskNode * PreComp, int ntask, const struct TreeBuilder *tb)
 {
     /*Set NextNode for this node*/
-    if(tail < tb.firstnode && tail >= 0 && force_get_next_node(tail, tb) != -1) {
-        endrun(2,"Particle %d with tail %d already has tail set: %d\n",no, tail, force_get_next_node(tail, tb));
+    if(tail < tb->firstnode && tail >= 0 && force_get_next_node(tail, *tb) != -1) {
+        endrun(2,"Particle %d with tail %d already has tail set: %d\n",no, tail, force_get_next_node(tail, *tb));
     }
-    tail = force_set_next_node(tail, no, tb);
+    tail = force_set_next_node(tail, no, *tb);
 
     /* For particles and pseudo particles we have nothing to update; */
     /* But the new tail is the last particle in the linked list. */
-    if(no < tb.firstnode || no >= tb.lastnode) {
+    if(no < tb->firstnode || no >= tb->lastnode) {
         int next = no;
         while(next != -1) {
             no = next;
-            next = force_get_next_node(next, tb);
+            next = force_get_next_node(next, *tb);
         }
         return no;
     }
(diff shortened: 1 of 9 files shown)


forcetree.c
```diff
@@ -805,8 +832,116 @@ force_update_node_recursive(int no, int sib, int tail, const struct TreeBuilder
...
+        return sib;
+    /*We reached the root node and are the last entry on it.*/
+    if(parent == tb->firstnode)
+        return -1;
+    /*If no sibling here, check the level above*/
+    return force_get_sibling(parent,tb);
+}
 
+static void
+force_queue_refinement(int no, int level, int * index, struct TaskNode *TaskNodes, const int nmax, const struct TreeBuilder *tb)
```
rainwoodman: What does this function do?

rainwoodman:
The split/merge algorithm you mean (from a brief read of the code)
1. truncate the tree to a forest, 
2. do the recursion on each new tree, 
3. then run a shorter update on the tree of the forest
Or something more complicated?

forcetree.c
sbird: I will add a comment

sbird:
Nope, that's all I mean. I will add a comment on the algorithm.

rainwoodman:
Sure. Did you look into using omp's task construct? I suspect it may be cleaner than the explicit tasking loop you current use. The wikibook gives some good description on the topic

 https://en.wikibooks.org/wiki/OpenMP/Tasks

sbird:
I did look into it. Actually the first thing I did was just to call force_update_recursive with #pragma omp task. But that doesn't work easily, because each call uses the tail value from the previous one. It is possible that one could make it work somehow, but I didn't figure it out.

I also looked into moving the moments calculation into the node creation routine, but that was very slow because of the locking involved, and used a lot of memory because it needed a per-cpu copy of the node tree.

rainwoodman:
I think it is good strategy to keep the tree build as minimal as possible, as it requires locking.

The example on wikibook seems to be exactly doing the tail value problem. The caller has to host the tail value. Or was it more complicated?

The `final` clause appear to give a much simpler syntax (no need to duplicate a parallel and serial version of code:
```
double sum_p(double * s, size_t size)
{
    size_t half = size / 2;
    if(size == 0) {
        return 0;
    }
    if(size == 1) {
        return s[0];
    }
    double x, y;

    #pragma omp task shared(x) final(size <= 1024)
    x = sum_p(s, half);

    #pragma omp task shared(y) final(size <= 1024)
    y = sum_p(s + half, size - half);

    #pragma omp taskwait

    return x + y;
}

double sum(double * s, size_t size)
{
    double r;
    #pragma omp parallel 
    #pragma omp single nowait
    r = sum_p(s, size);

    return r;
}
```

sbird:
The difference is that the return value in that example is only used by the parent recursion, whereas for us the return value is used by siblings. So we need something like:

#pragma omp parallel
    {
        #pragma omp task
        tail = sum(tail);
        #pragma omp task
        tail = sum(tail);
    }

and that means you cannot run y until you have run x. It may be possible to solve this by moving the tail into the parent, so we have something like:

#pragma omp parallel
    {
        tails[8];
        #pragma omp task
         tail[1] = sum(tail);
        #pragma omp task
        tail[2] = sum(tail);
        set_next_node(tail, p)
    }

Then we also have to compute the sibling in the recursion loop, which might cause us to slow down.
I can give this a try when I get the chance - it might look cleaner and it would avoid the bsearch, which would be nice.

sbird:
So I was able to make this work at the cost of an extra 8-element stack array in each recursion. I had to move some things into the parent to make it fast, and there was a certain amount of effort around when to stop spawning tasks, but it is still cleaner now. I will run it through a realistic test.

sbird:
There are now more loops, but hopefully the compiler is smart enough to unroll them anyway.

forcetree.c
```diff
@@ -684,51 +687,88 @@ add_particle_moment_to_node(struct NODE * pnode, int i)
...
+            }
+            tails[j] = p;
+        }
+        else {
+            const int nextsib = force_get_sibling(sib, j, suns);
+            /*Don't spawn a new task if we only have one child,
+             *or if we are deep enough that we already spawned a lot.
+             Note: final clause is much slower for some reason. */
+            if(chldcnt > 1 && level >= 0 && level < 513) {
+                #pragma omp task shared(tails)
```
rainwoodman: Didn't the final clause work?

rainwoodman: I think it is because the compiler is still emitting the code to do the book keeping of tasks. I am curious if mergeable helps (some say yes); but that's not very important. The current one is already very clean.

rainwoodman:
Additional heuristics in splitting the task may hugely improve the balancing.

I suspect we tend to have very shallow subtrees if only chopped at level 3. It may be useful to add a few counters to tree build to gather the imbalance between threads for further diagnostics (e.g. max, min and mean of the work done per thread), if that's not too hard.

Do we have any proxy about the depth of the subtree from tree build? If so we can say the split happens only if the depth is sufficient; this will likely hugely improve the balance, if we are targeting a lot more than 8 threads. 

In tree build we do lock the parent -- but does it go all the way to the root? If so we can update the depth there. Otherwise we can have a new serial routine that first calculating the depth with a recursion.

forcetree.c
sbird: What you say about final sounds plausible, but note that it slows things down even if the conditional is unchanged (so the final clause is always false). This suggests some kind of locking to me. mergeable is also slower (I just tested it). I hadn't seen it before, but it seems to give the compiler the option to not generate a task?  

The other thing I noticed was that one cannot have an array element in the shared clause. So manually unrolling the loop might help a little with contention. But I did not do that, because it is ugly.

sbird:
We don't lock the parent during build (we don't even touch it always, actually), but we could a 'depth' variable to each Node during insertion and pick that up once the lock is taken, then set a maxdepth counter atomically. Bit messy though.

Heuristics probably require more tests on larger real simulations - I am only able to go to 8 threads on my workstation. We may need to get access to a KNL node at some point and see what happens on that.

sbird:
Or by 'depth' you mean from the node to the leaf of the tree, not from the node to the root of the tree? I don't know any good proxy for the former that we can get without actually walking the tree again, which will be just as slow as the serial recursive moments calculation.

sbird:
This version passes my dm-only real simulation test case.

4 mpi and 2 threads was about 10% slower tree build with the task-based parallelization than the explicit parallelization (probably because of the differing treatment of topnodes, which are done in serial in the new version, whereas before I fanned out all the topleaves explicitly). This is probably not so much of a big deal.

sbird:
The special-casing for the topnodes actually made us much slower. I have removed it, but this may mean we need a larger number of tasks for larger simulations somehow.

rainwoodman:
Yes. That's what I mean by depth. My hope was that a pure walk without computing the moments will be a lot faster than walking again. Or perhaps we can compute the depth without strict correctness by not using atomic operations (since we only need a proxy of the true depth for balancing). 

A fairly large test will be necessary -- the current heuristic is affected by the number of external nodes, so the number of ranks matters too.

sbird:
With openmp, if I do x=0; x+=1 on two separate threads, is the result guaranteed to be either 1 or 2? Or can the writes overlap and do weird things?

rainwoodman:
If x is the size of an atomic read/write, and aligned, then it is either 1 or 2. So for practical purposes I think it won't do any weird thing.

Are there any contention on the shared tails array. No two tasks write to the same part of the shared array as far as I can see. shared means the compiler won't create any proxies for the variable in the task context.

sbird:
My worry was that it is all the tails array which is shared, not just one element. So one thread can be returning and writing to tails[2] at the same time as another writes to tails[1]. I am not sure if the compiler is smart enough to avoid contention in this case, or if it creates a single shared lock? It is probably a small effect.

rainwoodman:
There is no lock. I believe shared means no lock.

If tails[*] is 8 bytes and aligned then there definitely no issue. (I think the compiler automatically align to the width)

Currently tails[*] is int. I recall seeing somewhere a bug that gcc generates a code sequence that overwrites the adjacent member to the old value, because it thinks manipulating the full 8 bytes is faster; but it was a bug and fixed long ago. But this is only on x86-64. On some platform it is problem not possible to directly access 4 bytes. 

I will checkout the branch and look at the assembly to be sure.

sbird:
Thanks - I am not at assembly skills level of compsci.

rainwoodman:
It's not that bad. This is the segment that writes to tails in the task.

Notice that all it does is to fetch the pointer to `tails` to `_13`, then write to it with a 4 byte write `movl`. So we are good here.

```
   call   force_update_node_recursive#
   movl   %eax, %ecx  #, _12
   movq   -56(%rbp), %rax # .omp_data_i, tmp110
   movq   8(%rax), %rax   # .omp_data_i_2(D)->tails, _13
   movl   -32(%rbp), %edx # j, tmp112
   movslq %edx, %rdx  # tmp112, tmp111
   movl   %ecx, (%rax,%rdx,4) # _12, *_13    (tails[j] = force_update_node_recursive(....))
```

I assume openmp will trigger a fence with a library function at the end of the task. And that's probably what is causing the slowness with final -- a fence is probably requires even for a final task.

sbird:
Ah, that makes sense. If you are moderately happy with this approach I can run a larger scaling test?

rainwoodman:
Yes I am. Do you want me to do some tests on cori?

sbird:
On the knl nodes? Sure. We should do some general scalability tests on those anyway. Shall I still do some as well? 

For comparison, the tree build was around 6-7% of runtime in my small dm-only test with 8 threads.

rainwoodman:
I am thinking of adding a semi automatic benchmark system for NERSC.

sbird:
Oops, my larger test bugged:
[ 000013.01 ] Task 11: LASTMESSAGE did not end with new line: Updating pseudos: 5592407 -> -1 which is not an internal node between 5592405 and 9507673.

I think I know why...

sbird:
This is my gadget paramfile:

the box size is 300 Mpc.

InitCondFile = ICS/300_512_99
OutputDir = output
SnapshotFileBase = snap
TimeLimitCPU = 31000
TimeMax = 1.0
Omega0 = 0.288
OmegaLambda = 0.712
OmegaBaryon = 0.0
HubbleParam = 0.7
RadiationOn = 1
HydroOn = 1
Nmesh = 1024
SnapshotWithFOF = 1
FOFHaloLinkingLength = 0.2
OutputList = "0.02,0.1,0.2,0.25,0.3333,0.5,0.66667,0.83333"
MassiveNuLinRespOn = 0
LinearTransferFunction = camb_linear/ics_transfer_99.dat
DensityKernelType = cubic
CoolingOn = 0
StarformationOn = 0
PartAllocFactor = 2
BlackHoleOn = 0
WindOn = 0

sbird:
The crash is because all the sibling entries are -1. The test just committed passes on my laptop and fails on the cluster.

sbird:
Interestingly, icc compiles the recursive test code to something which is much faster than gcc manages.

rainwoodman:
Here is a number on KNL from master. On 256x4 (1024 hyper threads), a 512**3 simulation of dm-only with 1024**3 mesh, it took half an hour to go 68 steps, from z=99 to z=46.

The biggest offender is Tree walk.
```
Timeline@timestep.c:0231          4.69  0.3%        0.09  1.4%        0.09       0.09
 Tree                           1531.77 89.5%        6.02 92.2%        6.02       6.02
  Build@forcetree.c:0112           22.16  1.3%        0.27  4.1%        0.27       0.27
  Misc@gravshort-tree.c:0096        7.88  0.5%        0.05  0.8%        0.04       0.06
  PostProcess@gravshort-tree.c:0088        3.51  0.2%        0.00  0.1%        0.00       0.04
  Recv@gravshort-tree.c:0090       24.93  1.5%        0.37  5.6%        0.01       0.42
  Send@gravshort-tree.c:0089        4.18  0.2%        0.03  0.5%        0.02       0.03
  Wait1@gravshort-tree.c:0091      279.91 16.3%        4.69 71.8%        0.00       5.24
  Wait2@gravshort-tree.c:0092        1.16  0.1%        0.02  0.2%        0.00       0.02
  Walk1@gravshort-tree.c:0086     1112.56 65.0%        0.54  8.3%        0.00       5.22
  Walk2@gravshort-tree.c:0087       75.47  4.4%        0.05  0.7%        0.00       0.41
```

I find this quite strange.....

sbird:
Is it better on master?

rainwoodman:
I was talking about master.

rainwoodman:
I'll see how it goes without KNL.

sbird:
That is very strange. The wait time is also remarkably long for this size, which suggests some kick of locking contention.

rainwoodman:
It appears to be pretty consistent over the board -- did I set up something wrong all the time?
Here is the 2048^3 run with 4096 ranks (1380 Mpc/h), at 47 steps
```
 Timeline@timestep.c:0230         10.26  0.4%        0.01  0.0%        0.01       0.01
 Tree                           1437.89 60.2%       34.38 94.2%       34.38      34.39
  Build@forcetree.c:0110           35.18  1.5%        0.51  1.4%        0.51       0.51
  Misc@gravshort-tree.c:0096       12.04  0.5%        0.33  0.9%        0.22       0.47
  PostProcess@gravshort-tree.c:0088        4.13  0.2%        0.12  0.3%        0.09       0.26
  Recv@gravshort-tree.c:0090       28.12  1.2%        0.63  1.7%        0.10       1.14
  Send@gravshort-tree.c:0089        7.72  0.3%        0.22  0.6%        0.18       0.24
  Wait1@gravshort-tree.c:0091      425.95 17.8%        7.22 19.8%        0.00      14.34
  Wait2@gravshort-tree.c:0092        1.58  0.1%        0.04  0.1%        0.00       0.07
  Walk1@gravshort-tree.c:0086      878.34 36.8%       24.09 66.0%       17.02      31.30
  Walk2@gravshort-tree.c:0087       44.84  1.9%        1.22  3.3%        0.72       1.75
Step 47, Time: 0.0253389, MPIs: 4096 Threads: 2 Elapsed: 2391.87
```

Here is the 50 Mpc/h 512^3 run with 256 ranks
```
   Short@timestep.c:0257             3.63  0.9%        0.06  3.4%        0.03       0.10
 Timeline@timestep.c:0231          2.35  0.6%        0.06  3.2%        0.06       0.06
 Tree                            296.04 73.1%        1.15 66.3%        1.15       1.15
  Build@forcetree.c:0112            8.27  2.0%        0.14  8.1%        0.14       0.14
  Misc@gravshort-tree.c:0096        4.18  1.0%        0.02  1.0%        0.01       0.03
  PostProcess@gravshort-tree.c:0088        1.66  0.4%        0.00  0.0%        0.00       0.01
  Recv@gravshort-tree.c:0090        5.88  1.5%        0.05  3.1%        0.00       0.06
  Send@gravshort-tree.c:0089        1.82  0.4%        0.00  0.3%        0.00       0.00
  Wait1@gravshort-tree.c:0091       68.12 16.8%        0.91 52.3%        0.00       0.93
  Wait2@gravshort-tree.c:0092        0.52  0.1%        0.00  0.2%        0.00       0.00
  Walk1@gravshort-tree.c:0086      190.39 47.0%        0.02  1.2%        0.00       0.93
  Walk2@gravshort-tree.c:0087       15.21  3.8%        0.00  0.1%        0.00       0.05
```

Did we mess up something major in the treewalk? I don't think it is supposed to be so bad comparing to mesh -- how was it on your runs?

rainwoodman:
Do you think this PR is ready to be merged? The TreeWalk slowness shall be reopened as another issue.

sbird:
I haven't found any problems with the PR, so I would be ok to merge it. For the TreeWalk slowness, did we accidentally make sigma8 very large for the dm-only test or something?

sbird:
My tests have a slow treewalk also, but I expected it, because they are neutrino particle simulations. Maybe check the last domain decomposition merge? The treewalk can be quite sensitive to particle ordering.

sbird:
This is a run with 512 CDM, 512 nu, on 96 ranks: It is about half the speed of an earlier 512 CDM, linear response neutrino simulation, which is what I would expect. I used S-GenIC, not MP-GenIC.
I have been seeing fairly large wait percentages for a while now.

 PMgrav                         1207.57  7.3%       17.04 23.9%       17.04      17.04
  Misc@petapm.c:0279                1.49  0.0%        0.02  0.0%        0.01       0.02
  Misc@petapm.c:0290                0.00  0.0%        0.00  0.0%        0.00       0.00
  Misc@petapm.c:0337                0.00  0.0%        0.00  0.0%        0.00       0.00
  Regions@gravpm.c:0187            43.53  0.3%        0.62  0.9%        0.52       0.70
  c2r@petapm.c:0328               353.42  2.1%        5.15  7.2%        4.87       5.61
  calc@petapm.c:0326              139.82  0.8%        1.94  2.7%        1.82       1.97
  cic@petapm.c:0281                74.22  0.4%        1.13  1.6%        1.05       1.23
  comm@petapm.c:0286               68.38  0.4%        0.93  1.3%        0.65       1.01
  comm@petapm.c:0331              219.81  1.3%        2.93  4.1%        2.22       3.54
  r2c@petapm.c:0313               164.04  1.0%        2.32  3.2%        2.05       2.66
  readout@petapm.c:0334           142.85  0.9%        2.00  2.8%        1.73       2.70
 SPH                               0.25  0.0%        0.00  0.0%        0.00       0.00
  Density                           0.25  0.0%        0.00  0.0%        0.00       0.00
   Comm@density.c:0223               0.00  0.0%        0.00  0.0%        0.00       0.00
   Compute@density.c:0221            0.00  0.0%        0.00  0.0%        0.00       0.00
   Init@density.c:0172               0.13  0.0%        0.00  0.0%        0.00       0.00
   Misc@density.c:0224               0.10  0.0%        0.00  0.0%        0.00       0.00
   Wait@density.c:0222               0.02  0.0%        0.00  0.0%        0.00       0.00
  Write@io.c:0062                  20.60  0.1%        0.00  0.0%        0.00       0.00
 Timeline                        129.11  0.8%        0.33  0.5%        0.27       0.38
  HalfKick                         82.29  0.5%        0.33  0.5%        0.27       0.38
   Long@timestep.c:0268             20.10  0.1%        0.15  0.2%        0.11       0.18
   Short@timestep.c:0257            62.18  0.4%        0.19  0.3%        0.16       0.20
 Timeline@timestep.c:0231         46.83  0.3%        0.00  0.0%        0.00       0.00
 Tree                          13634.12 82.5%       50.46 70.7%       50.46      50.46
  Build@forcetree.c:0112         1737.08 10.5%        1.93  2.7%        1.93       1.93
  Misc@gravshort-tree.c:0096       66.84  0.4%        0.36  0.5%        0.30       0.38
  PostProcess@gravshort-tree.c:0088       31.78  0.2%        0.19  0.3%        0.15       0.23
  Recv@gravshort-tree.c:0090      140.43  0.8%        0.64  0.9%        0.13       1.03
  Send@gravshort-tree.c:0089       15.32  0.1%        0.09  0.1%        0.09       0.10
  Wait1@gravshort-tree.c:0091     2564.83 15.5%       13.84 19.4%        4.52      20.92
  Wait2@gravshort-tree.c:0092        3.43  0.0%        0.02  0.0%        0.00       0.06
  Walk1@gravshort-tree.c:0086     8621.50 52.1%       31.64 44.3%       24.60      40.92
  Walk2@gravshort-tree.c:0087      452.91  2.7%        1.75  2.4%        1.35       2.25

sbird:
I will merge this and move discussion of the treewalk to another thread

PR #196, opened 2018-01-25
Fixes for particle neutrinos

sbird:
Two trivial fixes for pure particle neutrinos. These have no effect if you are using the linear response neutrinos, which @andreufont is.

### DIFF
--- a/genic/main.c
+++ b/genic/main.c
@@ -58,7 +58,19 @@ int main(int argc, char **argv)
   if(0 != big_file_mpi_create(&bf, buf, MPI_COMM_WORLD)) {
       endrun(0, "%s\n", big_file_get_error_message());
   }
-  saveheader(&bf, TotNumPart);
+  /*Massive neutrinos*/
+
+  const int64_t TotNu = (int64_t) NGridNu*NGridNu*NGridNu;
+  double total_nufrac = 0;
+  if(TotNu > 0) {
+    const double kBMNu = 3*CP.ONu.kBtnu / (CP.MNu[0]+CP.MNu[1]+CP.MNu[2]);
+    double v_th = NU_V0(InitTime, kBMNu, UnitVelocity_in_cm_per_s);
+    if(!UsePeculiarVelocity)
+        v_th /= sqrt(InitTime);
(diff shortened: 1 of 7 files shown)


genic/main.c
```diff
@@ -58,13 +58,25 @@ int main(int argc, char **argv)
...
+  /*Massive neutrinos*/
+
+  const int64_t TotNu = (int64_t) NGridNu*NGridNu*NGridNu;
+  double total_nufrac = 0;
+  if(TotNu > 0) {
+    const double kBMNu = 3*CP.ONu.kBtnu / (CP.MNu[0]+CP.MNu[1]+CP.MNu[2]);
+    double v_th = NU_V0(InitTime, kBMNu, UnitVelocity_in_cm_per_s);
+    if(!UsePeculiarVelocity)
+        v_th /= sqrt(InitTime);
+    total_nufrac = init_thermalvel(v_th, Max_nuvel/v_th, 0);
```
rainwoodman: Is this the total fraction of massive neutrino? Would it be useful to save this value in the header as a properly named attr?

genic/power.c
```diff
@@ -92,7 +92,8 @@ void parse_transfer(int i, double k, char * line, struct table *out_tab, int * I
      * Everything is a ratio against tot*/
     out_tab->logD[0][i] = pow(transfers[1]/transfers[5],2);
     out_tab->logD[1][i] = pow(transfers[0]/transfers[5],2);
-    out_tab->logD[2][i] = pow(transfers[6]/transfers[5],2);
+    out_tab->logD[2][i] = pow(transfers[4]/transfers[5],2);
+    out_tab->logD[3][i] = pow(transfers[6]/transfers[5],2);
```
rainwoodman: Could you add a comment which column is which species?

genic/main.c
sbird: Yes! This is the fraction of massive neutrino in particles.

genic/power.c
sbird: Done

sbird:
Will merge when travis goes green.

rainwoodman:
It is green.

PR #197, opened 2018-01-26
install class to depends directory.

rainwoodman:
How do we deal with the data files used by class?

And it will be pain if class tries to read so many files
from all ranks.

### DIFF
--- a/depends/Makefile
+++ b/depends/Makefile
@@ -4,6 +4,9 @@ include $(CONFIG)
 LIBRARIES=lib/libbigfile-mpi.a lib/libmpsort-mpi.a lib/libpfft_omp.a
 depends: $(LIBRARIES)
 
+lib/libclass.a: Makefile
+	bash class/install_class.sh $(PWD)/class 2.6.1 $(PWD)
+
 lib/libmpsort-mpi.a: mpsort/mpsort-mpi.c
 	mkdir -p lib; \
 	mkdir -p include; \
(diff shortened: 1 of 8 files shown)


sbird:
My memory of CAMB is that it doesn't need the datafiles for the transfer functions, only for the CMB C_ls, so we may be ok.

PS Why the extra patches in the repo? Are they classylss patches?

rainwoodman:
These are. I didn't bother to remove them.

Here is a list of files -- at minimal bbn and hyrec are probably still needed for the the spectra module, which computes the transfer functions. Does this dependency physically make sense?

```
.
├── bbn
│   └── sBBN.dat
├── benchmark.ini
├── cl_ref.pre
├── concise.ini
├── explanatory.ini
├── external_Pk
│   ├── Pk_example.dat
│   └── Pk_example_w_tensors.dat
├── hyrec
│   ├── Alpha_inf.dat
│   ├── input.dat
│   ├── R_inf.dat
│   └── two_photon_tables.dat
├── lcdm.ini
├── myevolution.dat
├── myselection.dat
├── output
│   ├── explanatory00_cl.dat
│   ├── explanatory00_cl_lensed.dat
│   └── explanatory00_parameters.ini
├── pk_ref.pre
├── psd_FD_single.dat
└── trg.ini
```

sbird:
Yes. Indeed we need the hyrec and bbn codes. The external P(k) stuff is useful to us too, although we don't need those data files.

rainwoodman:
I'll close this for now; since we'd try a more decoupled approach.

PR #199, opened 2018-01-30
Bring back WDM thermal velocities

sbird:
Once again allow adding thermal velocities corresponding to warm dark matter to the initial CDM distribution. Do not alter the power spectrum though - that is best done externally.

### DIFF
--- a/genic/main.c
+++ b/genic/main.c
@@ -49,7 +49,7 @@ int main(int argc, char **argv)
       shift_nu = -0.5 * (CP.Omega0 - OmegaNu) / CP.Omega0 * meanspacing;
       shift_dm = 0.5 * OmegaNu / CP.Omega0 * meanspacing;
   }
-  setup_grid(ProduceGas * shift_dm, 0);
+  setup_grid(ProduceGas * shift_dm, 0, Ngrid);
 
   /*Write the header*/
   char buf[4096];
@@ -101,7 +101,7 @@ int main(int argc, char **argv)
 
   /*Now make the gas if required*/
   if(ProduceGas) {
-    setup_grid(shift_gas, TotNumPart);
+    setup_grid(shift_gas, TotNumPart, Ngrid);
     displacement_fields(GasType);
     write_particle_data(0, &bf);
     free_ffts();
@@ -109,7 +109,7 @@ int main(int argc, char **argv)
   /*Now add random velocity neutrino particles*/
   if(NGridNu > 0) {
       int i;
-      setup_grid(shift_nu, 2*TotNumPart);
+      setup_grid(shift_nu, 2*TotNumPart, NGridNu);
       displacement_fields(NuType);
       for(i = 0; i < NumPart; i++)
           add_thermal_speeds(&nu_therm, P[i].ID, P[i].Vel);
--- a/genic/main.c
+++ b/genic/main.c
@@ -62,12 +62,13 @@ int main(int argc, char **argv)
 
   const int64_t TotNu = (int64_t) NGridNu*NGridNu*NGridNu;
   double total_nufrac = 0;
+  struct thermalvel nu_therm;
   if(TotNu > 0) {
     const double kBMNu = 3*CP.ONu.kBtnu / (CP.MNu[0]+CP.MNu[1]+CP.MNu[2]);
     double v_th = NU_V0(InitTime, kBMNu, UnitVelocity_in_cm_per_s);
     if(!UsePeculiarVelocity)
         v_th /= sqrt(InitTime);
-    total_nufrac = init_thermalvel(v_th, Max_nuvel/v_th, 0);
+    total_nufrac = init_thermalvel(&nu_therm, v_th, Max_nuvel/v_th, 0);
     message(0,"F-D velocity scale: %g. Max particle vel: %g. Fraction of mass in particles: %g\n",v_th*sqrt(InitTime), Max_nuvel*sqrt(InitTime), total_nufrac);
   }
   saveheader(&bf, TotNumPart, TotNu, total_nufrac);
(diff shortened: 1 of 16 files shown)


genic/main.c
```diff
@@ -99,7 +112,7 @@ int main(int argc, char **argv)
       setup_grid(shift_nu, TotNu);
       displacement_fields(NuType);
       for(i = 0; i < NumPart; i++)
-          add_thermal_speeds(P[i].Vel);
+          add_thermal_speeds(&nu_therm, P[i].Vel);
```
rainwoodman: I recall Arka was talking about a scheme to drastically shrink the shotnoise level of these thermal speeds by coherently sampling the directions or something alike. He'll be around this Friday.

genic/params.c
```diff
@@ -42,6 +42,7 @@ create_parameters()
     param_declare_double(ps, "MNue", OPTIONAL, 0, "First neutrino mass in eV.");
     param_declare_double(ps, "MNum", OPTIONAL, 0, "Second neutrino mass in eV.");
     param_declare_double(ps, "MNut", OPTIONAL, 0, "Third neutrino mass in eV.");
+    param_declare_double(ps, "MWDM_therm", OPTIONAL, 0, "Assign a thermal velocity to the DM. Specifies WDM particle mass in keV.");
```
rainwoodman: So, WDM replaces CDM?

genic/thermal.c
```diff
@@ -76,29 +74,29 @@ init_thermalvel(const double v_amp, double max_fd,const double min_fd)
...
     return total_frac;
 }
 
 /*Add a randomly generated thermal speed in v_amp*(min_fd, max_fd) to a 3-velocity*/
 void
-add_thermal_speeds(float Vel[])
+add_thermal_speeds(struct thermalvel * thermals, float Vel[])
 {
-    const double p = gsl_rng_uniform (g_rng);
+    const double p = gsl_rng_uniform (thermals->g_rng);
```
rainwoodman: Where did you set the seed?

genic/params.c
sbird: Yup.

genic/thermal.c
sbird: I didn't - I just used the default seed. Shall I set it?

genic/main.c
sbird: You mean like this paper? https://arxiv.org/abs/1801.03906 That'd be a cool feature.

genic/thermal.c
rainwoodman: It will be correlated on different ranks, which is a very bad thing. 
A quick fix is to seed it by the rank -- but you probably want something that doesn't depend on the number of ranks. Main Gadget use a table of random numbers for different particle ID (by the modulus). The draw back is that it will have some spatial correlation. We probably need better scheme, e.g. generate a 2d seed table, then in each pencil sample the random number from the 2d. Some refactoring of the current gaussian white noise code would do. But too hairy?

genic/main.c
rainwoodman: Yes.

genic/thermal.c
sbird: A good point. Also affect neutrino particle simulations with MP-GenIC. I will fix it.

sbird:
What do you think of my proposed solution? I also added a test case and improved MP-Gadget's random number seeding while I was about it.

timebinmgr.c
```diff
@@ -35,7 +35,7 @@ setup_sync_points(double TimeIC, double no_snapshot_until_time)
     SyncPoints[1].a = All.TimeMax;
     SyncPoints[1].loga = log(All.TimeMax);
     SyncPoints[1].write_snapshot = 1;
-    SyncPoints[1].write_fof = 0;
+    SyncPoints[1].write_fof = All.SnapshotWithFOF ? 1 : 0;
```
rainwoodman: Is there a problem with the old one? If TimeMax was not in output list it won't write FOF, but a snapshot is dumpped just in case.

rainwoodman:
I am a bit confused by RNDTable. Is this an accurate description of what you are doing?

You have divided particles into chunks by their x, y, z coordinates, which happens to be the same striping along processors. Each chunk is assigned a RNG. The RNG is initialized such that the first chunk is connected to the last chunk of the previous rank. 

I see you are using ID for this striping. But ID on a single rank is not contiguous. Would that break your seeding / fast-forwarding logic?

sbird:
Nuts, it would. I did not think of that.

timebinmgr.c
sbird: Oh - I didn't realise that. I generally don't put TimeMax in the output list because at one point there was a bug that it would write two snapshots. If that is the case I can drop this commit.

sbird:
I will switch it to using one of the position dimensions for striping.

sbird:
Or I could just use each particle ID to seed the RNG directly: no need for any state. What about this version?

rainwoodman:
I worry there is no guarantee that gsl_uniform returned from different seeds is uniform.

sbird:
hmm. I think it has to be, if the first number in the rng is really guaranteed to adhere to a random distribution. I think if code like:
gsl_rng * rng = gsl_rng_alloc();
gsl_rng_set(rng, gettimeofday());
double rand = gsl_uniform(rng);
is safe, then it should be ok to reseed it with every id, right? I will double-check by generating a few million random numbers and plotting the distribution.

rainwoodman:
If the RNG does any non-linear transformation on the seed then it is not -- hence usually we generate a lot of numbers with the same seed instead. -- What you have here is just the first transformation from seed to uniform number.

I guess the worst case is we fall back to one seed per rank for now, and work out something else later?

sbird:
Yes, you'e right. I see some libraries to achieve this, but I can't find one that doesn't use a linear congruential generator. I guess we have two choices:
1) Seed each rank independently
2) Do the rng in serial

Which do you prefer?

sbird:
Now similar to the gaussian seeding in GenIC. Re-seeds every time Pos[2] == 0. Tested that it gives the same velocities irrespective of number of processors.

genic/thermal.c
```diff
@@ -76,33 +74,50 @@ init_thermalvel(const double v_amp, double max_fd,const double min_fd)
...
+    thermals->fd_intp = gsl_interp_alloc(gsl_interp_cspline,LENGTH_FERMI_DIRAC_TABLE);
+    thermals->fd_intp_acc = gsl_interp_accel_alloc();
+    gsl_interp_init(thermals->fd_intp,thermals->fermi_dirac_cumprob, thermals->fermi_dirac_vel,LENGTH_FERMI_DIRAC_TABLE);
     return total_frac;
 }
 
-/*Add a randomly generated thermal speed in v_amp*(min_fd, max_fd) to a 3-velocity*/
+/*Generate a table of random seeds, one for each pencil.*/
+unsigned int *
+init_rng(int Seed, int Nmesh)
```
rainwoodman: The memory consumption goes like N**3 -- and will be too much for 4096. With the pfft `local_ni` and `local_i_start` We can probably work something out later.

rainwoodman:
Yes. This looks OK.

genic/thermal.c
sbird: It goes like N^2

sbird:
Thanks, merging!

PR #201, opened 2018-02-11
Refactor fof slightly to reduce peak memory usage

sbird:
Avoid allocating memory for qsort when base is still allocated.

I had a simulation recently where this was the point of peak memory usage: moving the allocation a little later helped, because there is no qsort allocation with both Base and Group allocated.

### DIFF
--- a/fof.c
+++ b/fof.c
@@ -77,6 +77,10 @@ static void
 fof_compile_base(struct BaseGroup * base);
 static void
 fof_compile_catalogue(struct Group * group);
+
+static struct Group *
+fof_alloc_group(const struct BaseGroup * base, const uint64_t NgroupsExt);
+
 static void fof_assign_grnr(struct BaseGroup * base);
 
 void fof_label_primary(void);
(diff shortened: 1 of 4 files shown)


fof.c
```diff
@@ -76,7 +76,7 @@ static void fof_finish_group_properties(struct Group * Group);
 static void
 fof_compile_base(struct BaseGroup * base);
 static void
-fof_compile_catalogue(struct Group * group, struct BaseGroup * base);
+fof_compile_catalogue(struct Group * group);
```
rainwoodman: Can you keep this as a function?

sbird:
Sure, how's that?

sbird:
The last commit fixes a somewhat unfortunate bug: the post-processing FOF generation did not actually do anything! I guess no-one had used it.

I noticed at the same time that this mode does not save the per-particle Group ID, which is normally done when writing the snapshot. Do we want to fix this? If so, how? The easiest (code-wise) way is just to re-write the entire snapshot to disc. This is expensive, but the other way is easier to write bugs in (because the on-disc particles may have a different order to the in-memory particles). So it may be a bad idea for a feature we apparently never use.

sbird:
I have thought this over and I believe the current behaviour is best: snapshots once written should be immutable. It is easy enough to restart MP-Gadget from a snapshot with an output at the same time and get a new FOF table.

rainwoodman:
Thanks. Merging!

PR #202, opened 2018-02-19
Scaledepgrowth

sbird:
This is an implementation of scale-dependent growth factors. I am posting it to get review on the idea, but I have not fully tested it. Currently it generates ICs for the lya and hydro examples, and passes tests, but I do not know if the ICs are reasonable.

The basic approach is to use the most minimal, loose coupling of CLASS possible. A python script is added which parses an MP-GenIC parameter file and uses classylss to generate transfer tables and matter power spectra. 

Note these tables are passed in (CAMB/CLASS) text format to MP-GenIC, not as bigfile. Although this seems a bit like more work, I think that the benefit of a familiar input format is large. There is a large community of theorists (who I sometimes work with) that have modified CLASS/CAMB to do novel things, and asking them to send me transfer functions is much easier than asking them to send me a bigfile. 

So I tried to avoid mandating the use of our python script, which means I could also not have to cover all possible uses for it. 

TODO:
- Add a unit test for the growth function.
- Consider using the CAMB/CLASS velocity transfer functions directly for the velocities. This will involve working out their units and may not work.


sbird:
I forgot to mention: for other things requiring a Boltzmann solver (like radiation perturbations) it was my intention to let MP-GenIC store the required data in the ICs.

rainwoodman:
Would it be better to have the python script generate/check the GenIC parameter file, rather than parsing the parameter file? The current work flow ( python script reads the parameter file to produce data referred in the parameter file) feels a bit reversed -- unless the longer term goal is to have MP-Gadget, GenIC and the boltzmann solver use the same parameter file, at which point the path to the linear power spectrum /transfer function becomes optional.

tools/make_class_power.py
```diff
@@ -0,0 +1,206 @@
...
+
+from __future__ import print_function
+import sys
+import math
+import os.path
+import numpy as np
+import classylss
+import classylss.binding as CLASS
+import configobj
+import validate
```
rainwoodman: What are these two packages?

Are they in general installed? On my vanilla conda (with ipython notebook / jupyter) they do not seem to be installed; I am a bit concerned on this dependency.

rainwoodman: BTW; why didn't we just use a yaml file for parameters (which embeds nicely in C and Python).

tools/make_class_power.py
```diff
@@ -0,0 +1,206 @@
...
+        gparams['Omega_Lambda'] = config['OmegaLambda']
+    #Set up massive neutrinos
+    if omeganu > 0:
+        gparams['m_ncdm'] = '%.2f,%.2f,%.2f' % config['MNue'], config['MNum'], config['MNut']
+        gparams['N_ncdm'] = 3
+    #Power spectrum amplitude
+    if config['Sigma8'] > 0:
+        gparams['sigma8'] = config['Sigma8']
+    else:
+        #Pivot scale is by default 0.05 1/Mpc! This number is NOT what is reported by Planck.
```
rainwoodman: There is this comment but no action -- did you want to set something here?

tools/make_class_power.py
```diff
@@ -0,0 +1,206 @@
...
+    #Get and save the matter power spectrum
+    mink = 2*math.pi/boxmpc/100
+    kk = np.logspace(np.log10(mink), np.log10(maxk), max(config['Ngrid']*2,300))
+    pk_lin = powspec.get_pklin(k=kk, z=redshift)
+    pkfile = os.path.join(sdir, config['FileWithInputSpectrum'])
+    if os.path.exists(pkfile):
+        raise IOError("Refusing to write to existing file: ",pkfile)
+    np.savetxt(pkfile, np.vstack([kk, pk_lin]).T)
+
+def save_transfer(transfer, transferfile, bg, redshift):
```
rainwoodman: We shall probably add this functionality into classylss.

tools/make_class_power.py
```diff
@@ -0,0 +1,206 @@
...
+    #We do not want this for the matter power: we want CDM + b + massive-neutrino.
+    ftrans[:,6] = -1*(omegacdm *transfer['d_cdm'] + omegab * transfer['d_b'] + omeganu * ftrans[:,5])/(omeganu + omegacdm + omegab)/kk**2
+    #The CDM+baryon weighted density.
+    ftrans[:,7] = -1*(omegacdm *transfer['d_cdm'] + omegab * transfer['d_b'])/(omegacdm + omegab)/kk**2
+    ftrans[:,8] = -1*transfer['d_tot']/kk**2
+
+    np.savetxt(transferfile, ftrans)
+
+if __name__ == "__main__":
+    extpk = None
```
rainwoodman: Could you add argparse and a main function -- it's actually no more code than this but easier to read.

rainwoodman:
I 've been planning to shuffle the source code files around for a while -- too many files in the code root directory! This PR is sufficiently small that I don't think it will be too hard to rebase.. I'll look at this now.

tools/make_class_power.py
sbird: Don't do that yet - I have a patch to class pending to change the transfer function output to better match CAMB/CMBFAST and if that gets merged this function will need to change.

tools/make_class_power.py
sbird: I was unsure - in many ways a pivot scale of 0.05 1/Mpc is sensible for large scale structure, but it does mean that the A_s values are different from what is in the Planck papers, where pivot is 2e-3. If we want to be able to use Planck parameters we should change it. I wasn't sure which is more obvious.

tools/make_class_power.py
sbird: ok!

sbird:
My intention was to avoid specifying the same parameter in two places, and to fit nicely into the existing examples (which are also our current usage documentation).

My other solution for this does generate the MP-GenIC file. See: https://github.com/sbird/SimulationRunner/blob/master/SimulationRunner/simulationics.py#L166 What do you think about doing something like that instead?

sbird:
The trouble is that because the script has to know all of the MP-GenIC options to generate the file, the matrix of options becomes large, but I guess we can just pass most of them down unchanged.

tools/make_class_power.py
rainwoodman: I wasn't aware if a different pivot scale is not sensible for large scale structure, is it? 

I pulled the number nick put in nbodykit for Planck15, and it is larger than 0.05. The number (0.07ish?) looked somewhat strange does it?

```
In [1]: from nbodykit.cosmology import Planck15

In [2]: Planck15.k_pivot
Out[2]: 0.07381163271331562

In [3]: Planck15.A_s
Out[3]: 2.1413038238853928e-09
```

tools/make_class_power.py
rainwoodman: Sure.

andreufont:
@rainwoodman - I haven't followed the rest of the discussion, but if you use the same value of A_s and n_s, but a different pivot point, you'll efectively have a different amplitude of the power at a given k. 

The differences should be very small because n_s is very close to 1, but still large enough to be an important issue!

Note that this would not affect any cosmology for which you have also specified sigma_8, since that would already renormalize the power, but if you do not specify sigma_8 and rely on your value of A_s, then you would have a problem (and this is our situation in the neutrino project with Chris).

rainwoodman:
Let's keep the paramfile arrangement your current way -- It is not too difficult to teach Python our file format; and parameter file is messy anyways.

I attempted libyaml and realized it doesn't provide a document object model. So it means a lot of extra work duplicating our current simple, non-nested DOM. I decide it is not worth it.

I thought python has a builtin configparser (https://docs.python.org/3/library/configparser.html) I wonder if it is good enough?

rainwoodman:
Thanks. So it is really the primordial power law power spectrum that is at
play here!

sbird:
I'll see what I can do with configparser

sbird:
configparser has some issues. It is different on python 2 and python 3, it doesn't accept files without section headers without a massive hack, and it removes all comments on write.

However, I start to think that your idea of writing a script to generate genic parameter files is the way to go. I will try it.

sbird:
hm. The parameter file for the script that generates genic parameter files ends up looking very much like a copy of the genic paramter file! I think this is not worth the complexity.

sbird:
Is there a way to get more k bins for the transfer functions from classylss?

rainwoodman:
I never did this, but I assume class take some magic parameter for this? You can pass in any class parameter to the dictionary -- it will be sent in automatically.

rainwoodman:
I'll merge the file shuffle PR. You may need to do some rebase here.

sbird:
I guess there is a magic parameter, I just don't know what it is!

sbird:
I found it: k_per_decade_for_{pk|bao}

sbird:
MP-GenIC's memory usage is quite high:
[ 000004.11 ] Task 0:  PMreal               | B | 0115605504 0115609600 | petapm.c:678
[ 000004.11 ] Task 0:  PMcomplex            | B | 0115605504 0115609600 | petapm.c:679
[ 000004.11 ] Task 0:  PMrho_k              | B | 0115605504 0115609600 | petapm.c:680
[ 000004.11 ] Task 0:  PMmesh               | B | 0117113920 0117121024 | petapm.c:689
[ 000004.11 ] Task 0:  PMlayout             | B | 0000001024 0000008192 | petapm.c:367
[ 000004.11 ] Task 0:  PencilSend           | B | 0000532336 0000536576 | petapm.c:390
[ 000004.11 ] Task 0:  PencilRecv           | B | 0000516096 0000520192 | petapm.c:459
[ 000004.11 ] Task 0:  PMBufSend            | B | 0113246208 0113250304 | petapm.c:561

I understand why it needs three copies of the mesh array, but why does PMmesh, PMBufSend and PMRecvBuf have to be the same size as the whole mesh? And why do they have to be distinct?

sbird:
I have answered my own question above...

sbird:
Here is the current status at z=9 and z=49: looks like the baryon power is a little low for some reason...

part_000
part_003

rainwoodman:
Does the baryon displacement come from gradient of potential of delta or from finite differentiation of delta?

sbird:
From finite differentiation of the transfer function

sbird:
This is the ratio to CAMB at z=9: looks like the CDM is correct but the baryons are 10% low.

part_003-full

rainwoodman:
Why shifting the phase?

sbird:
It is recommended by: https://arxiv.org/pdf/1610.08501.pdf
The older https://arxiv.org/pdf/astro-ph/0305210.pdf
suggests using two different glass files.

I'm not completely convinced though - treat this as work in progress for now.

sbird:
This PR has gotten out of hand, so I am splitting it into two pull requests. The first pull will only contain fixes and small cleanups, and the second one will be the change to the initial velocities, which I am not yet sure I got right.

sbird:
This is what it looks like currently if I start at z=19:

fish://cmu/home/spb/data/transfer_functions/test_scaledep_small/output/PART_000.png
fish://cmu/home/spb/data/transfer_functions/test_scaledep_small/output/PART_001.png

not bad, but something is missing...

sbird:
and this is the absolute ratios to CAMB: makes it look like the DM is getting a bit too much power.
part_001-full

andreufont:
Hi Simeon,

Are you using the option to fix the initial amplitudes to sqrt(power)? I
help it useful for this type of test, a part from using a smaller power
amplitude to reduce non-linearities.

Cheers,
Andreu

sbird:
Not yet - but I will switch my test to Chris' soon I think.

sbird:
part_001-full
part_001
part_002-full
part_002

sbird:
This is what I have with the latest HEAD of this branch. I switched to requiring CLASS formatted transfer functions, which avoids the need for numerical differentiation, and uses the same formula as fastpm. Now the total growth function matches very closely what F_Omega gave us before.

However, the per-species growth functions still are not right!

Could someone (Yu?) give the current code a look through to make sure I have not messed something up? I am a little puzzled.

Chris-Pedersen:
I've run the low amplitude tests for massless neutrino cosmology, using the class files output by tools/make_class_output.py

scaledep99
scaledep9
scaledep3

sbird:
@Chris-Pedersen Thanks, that's extremely helpful. It looks like the baryon velocity is off somehow. Would you mind updating your pull request with the new config files? (or creating a new one, if that's easier). We'll merge it to this branch and see what we can do.

andreufont:
Hi Simeon, Chris,

Now the power does not look fine at z=99, at the initial conditions. That
was not the case before, right? At least in the master branch we had the
right power in the IC, right?

Cheers,
Andreu

rainwoodman:
I felt we may need to do backscaling to get this right. Even if the
background contains radiation, the baryons in MP-Gadget do not experience
the radiation drag, and thus the linear behavior should be wrong if we
don't backscale with a radiation-free linear model.

However -- it sounds like the radiation drag should go the wrong direction
from here?

On Mon, May 7, 2018 at 7:43 AM, Andreu Font-Ribera <notifications@github.com
> wrote:

> Hi Simeon, Chris,
>
> Now the power does not look fine at z=99, at the initial conditions. That
> was not the case before, right? At least in the master branch we had the
> right power in the IC, right?
>
> Cheers,
> Andreu
>

sbird:
At these redshifts radiation drag really should be negligible. There ought to be some effect of radiation perturbations on scales of the horizon, but that's k < 0.01, which we are nowhere near here. This ought to work; perhaps there is a bug in the IC code?

rainwoodman:
You mean negligible even at z=99? I suppose, one can run a two fluid model
forward with the radative background expansion and the radiative
perturbation at z=0, and see how th e baryons evolve to confirm this?

I can probably dig out a few old notebooks -- or do you already have
something?

If it is a bug in IC generator it can't affect baryons only.

sbird:
I do not have something, your notebooks would be very helpful! I am concerned that I have made an implementation error setting something to the total transfer which should instead be the species-specific transfer. This would only show up in the baryons.

Also, if we have an independent theory estimate of the velocity power from somewhere we could compare the velocity power spectrum in the ICs; that would confirm whether it is the ICs or the evolution that is messed up.

rainwoodman:
Here is the notebook. The large cell shall be replaced with an import from nbodykit when it gets released.

[ThreeFluid.zip](https://github.com/MP-Gadget/MP-Gadget/files/1985890/ThreeFluid.zip)

sbird:
Ok! With the notebook, I found the units error in the transfer functions - the CDM transfer was right, but the baryon one was off (I didn't realise t_b had different units to h_prime). Fixed, and running tests now...note neutrino transfer function is not tested yet.

sbird:
So it's fine at z=49, but somewhere between z=49 and z=9 we get this: looks like both the baryon and DM fluids are off by equal amounts in different directions. I'm going to try Chris' test in case this is non-linear stuff, and also try turning the tree off. We are much closer. 

part_002-full
part_002

sbird:
So - using a modified version of Chris' linear growth test I can reproduce the baryon/CDM offset for z > 49, but baryons don't grow enough (by a few percent) between z=19 and z=49. I don't think this is the initial conditions: if I turn the tree off I get extremely good agreement!

sbird:
It seems that there is still some error in the tree. One can fix this partially by setting Asmth = 0.125, as in Valkenburg & Villaescusa-Navarro (or by turning the tree off, or with adaptive softenings). This pull is really for about making the velocity dependent ICs work, so I would like to ask for a final review before merging, and investigate tree improvements in a separate pull request.

sbird:
Yu: Please review for merging!

sbird:
This is what it looks like with the tree off (from the linear_growth test):

part_001
part_002
part_005

sbird:
and this is with the tree on:
part_001

Interestingly with PR #207 I get:
part_001
we should merge this and then work on that, and general force accuracy.

rainwoodman:
I am looking at this.
So the reason is that the tree force amplified the shot noise, wiping out some of the large scale growh?

examples/travis/check-results.py
```diff
@@ -1,87 +1,109 @@
...
     #Check types have the same power
-    fig = Figure(figsize=(5, 5), dpi=200)
+    fig = Figure()
     canvas = FigureCanvasAgg(fig)
-    ax = fig.add_subplot(121)
+    ax = fig.add_subplot(111)
 
-    pklin = LinearPower(WMAP9, redshift=z)
+    ctf = os.path.join(transfer, "class_tk_99.dat")
+    if z < 98.5:
```
rainwoodman: why 98.5? Is there way way to notify z=99 is special here?

libgenic/power.c
```diff
@@ -75,33 +97,49 @@ void parse_power(int i, double k, char * line, struct table *out_tab, int * Inpu
...
-    out_tab->logD[3][i] = transfers[6]/transfers[5];
+     * 0 is baryons, 1 is CDM, 2 is massive neutrinos (we use the most massive neutrino species).
+     * 3 is growth function for baryons, 4 is growth function for CDM, 5 is growth function for massive neutrinos.
+     * We use the formulae for velocities from fastpm in synchronous gauge:
+     * https://github.com/rainwoodman/fastpm-python/blob/master/fastpm/multi.py#L185
+     * CDM = - h_prime / 2 / d_cdm
+     * bar = - (h_prime / 2  + t_b) / d_b
+     * nu = - (h_prime / 2 + t_ncdm) / d_ncdm
+     * and there is a normalisation factor of (1+z)/ hubble applied later on.*/
+    out_tab->logD[0][i] = -1*transfers[1];
```
rainwoodman: Can we rename them to physical names rather than these magical numbers? Or an enum
```
{
POWER_TABLE_TYPE_CDM = 0, ...

}
``

File: libgenic/power.c [hunk 7]
```diff
@@ -75,33 +97,49 @@ void parse_power(int i, double k, char * line, struct table *out_tab, int * Inpu
...
-     * 3 is massless neutrinos, 4 is massive neutrinos, 5 is total,
-     * 6 is cdm + baryons.*/
-    out_tab->logD[0][i] = transfers[1]/transfers[5];
-    out_tab->logD[1][i] = transfers[0]/transfers[5];
-    out_tab->logD[2][i] = transfers[4]/transfers[5];
-    out_tab->logD[3][i] = transfers[6]/transfers[5];
+     * 0 is baryons, 1 is CDM, 2 is massive neutrinos (we use the most massive neutrino species).
+     * 3 is growth function for baryons, 4 is growth function for CDM, 5 is growth function for massive neutrinos.
+     * We use the formulae for velocities from fastpm in synchronous gauge:
+     * https://github.com/rainwoodman/fastpm-python/blob/master/fastpm/multi.py#L185
```
rainwoodman: The permlink won't change after master advances:

https://github.com/rainwoodman/fastpm-python/blob/02ce2ff87897f713c7b9204630f4e0257d703784/fastpm/multi.py#L185

libgenic/power.c
```diff
@@ -75,33 +97,49 @@ void parse_power(int i, double k, char * line, struct table *out_tab, int * Inpu
...
+     * nu = - (h_prime / 2 + t_ncdm) / d_ncdm
+     * and there is a normalisation factor of (1+z)/ hubble applied later on.*/
+    out_tab->logD[0][i] = -1*transfers[1];
+    out_tab->logD[1][i] = -1*transfers[2];
+    /*This is ur if neutrinos are massless*/
+    out_tab->logD[2][i] = -1*transfers[3+nnu];
+    /*h_prime is entry 8 + nnu. t_b is 12 + nnu, t_ncdm[2] is 13 + nnu * 2.*/
+    out_tab->logD[3][i] = transfers[12+nnu];
+    out_tab->logD[4][i] = transfers[8+nnu] * 0.5;
+    out_tab->logD[5][i] = transfers[13+nnu*2];
```
rainwoodman: The unit here is in CLASS  units. (time derivative is against CLASS unit of time). Could you add a xref to the function that does the unit conversion to GADGET unit?

libgenic/power.h
```diff
@@ -14,10 +16,31 @@ struct power_params
...
+ * See also (Valkenburg & Villaescusa-Navarro 2016) arxiv:1610.08501, which includes
+ * an N-body IC code.
+ *
+ * In this gauge, as long as there is zero anisotropic stress (true if omega_r = 0)
+ * the equations of motion are the Newtonian equations and the density is the
+ * comoving synchronous gauge density. Roughly speaking one is choosing a spatial gauge
+ * comoves with the large-scale radiation perturbation, making them zero.
+ *
+ * The velocities for this gauge can be obtained by differentiating the synchronous gauge
+ * velocity perturbations. Note that on super-horizon scales this is not the same as the
```
rainwoodman: You sure it is velocity perturbations? It appears we are diffing against density perturbations.

tools/make_class_power.py
```diff
@@ -127,6 +125,10 @@ def make_class_power(paramfile, external_pk = None, extraz=None):
         - Massive neutrinos.
         - Using Sigma8 to set the power spectrum scale.
         - Different transfer functions.
+
+    We use class velocity transfer functions to have accurate initial conditions
+    even on superhorizon scales, and to properly support multiple species.
+
     Not supported:
         - Warm dark matter power spectra.
         - Rescaling with different transfer functions."""
```
rainwoodman: Rescaling can be done by running the 3-fluid model backward from target redshift to the initial redshift. Caveat is we don't yet know effective sound-speed of baryons. It can be a cool entry level student project.

tools/make_class_power.py
```diff
@@ -184,55 +184,32 @@ def make_class_power(paramfile, external_pk = None, extraz=None):
...
 
-def save_transfer(transfer, transferfile, bg, redshift):
-    """Save a transfer function. Note we save the CAMB FORMATTED transfer functions.
-    These can be generated from CLASS by passing the 'format = camb' on the command line.
-    The transfer functions differ by:
+def save_transfer(transfer, transferfile):
+    """Save a transfer function. Note we save the CLASS FORMATTED transfer functions.
+    The transfer functions differ from CAMB by:
         T_CAMB(k) = -T_CLASS(k)/k^2 """
     if os.path.exists(transferfile):
```
rainwoodman: The check shall probably be done in the main function rather than here -- the current function really is maybe_save_transfer.

rainwoodman:
A few changes in the reference and the naming of columns in the power table.

examples/travis/check-results.py
sbird: Yes, that was hacky, sorry. I'll fix it.

tools/make_class_power.py
sbird: Sure, comment changed. I prefer not to rescale, precisely because it hides force errors like the ones we are encountering :)

tools/make_class_power.py
sbird: Fixed!

libgenic/power.c
sbird: Done.

sbird:
Yu: our travis integration is broken! Where did it go?
I have fixed the review comments. I also found a major bug which will have broken CDM+neutrino simulations. These I will test next.

PR #203, opened 2018-02-20
Split executables from the library.

rainwoodman:
This is a step towards better modularity. A few changes:

- libgadget-utils is split from libgadget; as a way to explore building sub-libraries.
- libgadget-ic is a subset of libgadget; needs to go away eventually; it is a work around for the poor binary integration of genic and libgadget;
- genic modules are not built as a library yet; it needs to be further merged with libgadget, and eventually go away;
- test cases now mostly links with the entire libgadget;
- header files are formatted as ```<libgadget/....h>``` to avoid name conflicts;
- supporting libraries are built and linked to executables, but not bundled into libgadget.
- a portion of kspace-neutrino is built directly into libgadget.
- offtree build always puts object files in ```.objs```. the executables are put into two directories, one for gadget, one for genic.
- dependency between binary and library is not coupled at object level; so one must rebuild the library before rebuilding the exectuable. makefile in the toplevel does this.


sbird:
This does not compile for me: 
cd libgadget; make
make[1]: Entering directory '/home/spb/codes/MP-Gadget/libgadget'
make[1]: *** No rule to make target '.objs/gdbtools.o', needed by 'libgadget.a'.  Stop.
make[1]: Leaving directory '/home/spb/codes/MP-Gadget/libgadget'
make: *** [Makefile:14: all] Error 2

sbird:
Ah - I see this is because I did not re-check out the submodules.

sbird:
I think that soonish I will move the remaining parts of kspace-neutrinos that we use directly into MP-Gadget and remove the submodule...

sbird:
Yes, there are unfortunately interesting compile time tricks. Apologies. 

However...we are not using very much of the library anymore, having replaced most of it with more powerful native routines. What remains could just be imported and fully integrated. I will add it to my todo, but will not get to it for a little while.

PR #204, opened 2018-02-24
Petamem

sbird:
This PR performs a bit of housekeeping in petapm.

It makes some variables that were previously global statics local. It then uses the double-ended malloc to reduce the peak memory usage in MP-GenIC by 2 * fftsize * sizeof(double).

### DIFF
--- a/libgadget/petapm.c
+++ b/libgadget/petapm.c
@@ -336,9 +336,8 @@ void petapm_force_c2r(
         pfft_execute_dft_c2r(plan_back, complx, real);
         walltime_measure("/PMgrav/c2r");
         myfree(complx);
-        /* read out the potential */
+        /* read out the potential: this will copy and free real.*/
         layout_build_and_exchange_cells_to_local(&layout, meshbuf, real);
-        myfree(real);
         walltime_measure("/PMgrav/comm");
         
         pm_iterate(readout, regions);
@@ -620,12 +619,17 @@ static void to_region(double * cell, double * region) {
 }
 
 static void layout_build_and_exchange_cells_to_local(struct Layout * L, double * meshbuf, double * real) {
-    L->BufSend = mymalloc("PMBufSend", L->NcExport * sizeof(double));
     L->BufRecv = mymalloc("PMBufRecv", L->NcImport * sizeof(double));
     int i;
     int offset;
 
+    /*layout_iterate_cells transfers real to L->BufRecv*/
     layout_iterate_cells(L, to_region, real);
+
+    /*Real is done now: reuse the memory for BufSend*/
+    myfree(real);
+    /*Now allocate BufSend, which is confusingly used to receive data*/
+    L->BufSend = mymalloc("PMBufSend", L->NcExport * sizeof(double));
 
     /* exchange cells */
     /* notice the order is reversed from to_pfft */
@@ -643,8 +647,8 @@ static void layout_build_and_exchange_cells_to_local(struct Layout * L, double *
                 sizeof(double) * p->len);
         offset += p->len;
     }
-    myfree(L->BufRecv);
     myfree(L->BufSend);
+    myfree(L->BufRecv);
 }
 
 /* iterate over the pairs of real field cells and RecvBuf cells 
--- a/libgadget/petapm.c
+++ b/libgadget/petapm.c
@@ -76,7 +76,6 @@ static int NTask;
 static int ThisTask;
 
 /* these variables are allocated every force calculation */
-static double * real;
 static double * meshbuf;
 static size_t meshbufsize;
 static pfft_complex * rho_k;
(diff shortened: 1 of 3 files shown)


rainwoodman:
Could you point to me the location of the saving?

libgadget/petapm.c
```diff
@@ -310,57 +318,61 @@ void petapm_force_r2c(
...
+        /* apply the greens function turn rho_k into potential in fourier space */
+        pm_apply_transfer_function(&fourier_space_region, rho_k, complx, transfer);
         walltime_measure("/PMgrav/calc");
+
+        double * real = (double * ) mymalloc2("PMreal", fftsize * sizeof(double));
         pfft_execute_dft_c2r(plan_back, complx, real);
         walltime_measure("/PMgrav/c2r");
-        /* read out the potential */
-        layout_build_and_exchange_cells_to_local(&layout);
+        myfree(complx);
```
sbird: There is one saving here, where complx is freed.

libgadget/petapm.c
```diff
@@ -607,13 +619,18 @@ static void to_region(double * cell, double * region) {
     *region = *cell;
 }
 
-static void layout_build_and_exchange_cells_to_local(struct Layout * L) {
-    L->BufSend = mymalloc("PMBufSend", L->NcExport * sizeof(double));
+static void layout_build_and_exchange_cells_to_local(struct Layout * L, double * meshbuf, double * real) {
     L->BufRecv = mymalloc("PMBufRecv", L->NcImport * sizeof(double));
```
sbird: This array now goes into the memory previously allocated for complx

libgadget/petapm.c
```diff
@@ -607,13 +619,18 @@ static void to_region(double * cell, double * region) {
...
     int i;
     int offset;
 
-    layout_iterate_cells(L, to_region);
+    /*layout_iterate_cells transfers real to L->BufRecv*/
+    layout_iterate_cells(L, to_region, real);
+
+    /*Real is done now: reuse the memory for BufSend*/
+    myfree(real);
```
sbird: Then we free real before allocating PMBufSend, so that PMBufSend can go into the memory previously allocated for it.

sbird:
Does that help? The peak memory usage is lower because we do not allocate real and complx at the same time as the Recv and Send buffers. 

I thought that it may be possible to reduce it a bit more by avoiding allocating meshbuf until after the AlltoAllv for MP-GenIC, but since rebuilding the regions may be a bit slow, and we would have to do it for every axis I decided not to do that yet. It also seemed a bit messy to make the same code work for MP-GenIC and petapm.c.

sbird:
By the way, shall we create a -stable branch before merging this?

rainwoodman:
My experience with 'stable' branches has been negative. It behaves more or
less like a tag by definition. But we can certainly try..

Yu

sbird:
Even if it is basically a tag, it would at least mean that we can point people at something we know to be reasonably good. Until we find bugs that we can't backport...

By the way, I spent yesterday chasing a NaN coming from the hydro code that disappeared when I put a if(isnan()) message(); in. Must be a compiler bug.

rainwoodman:
I feel the real issue is that currently feature branches are not sufficiently tested before merged to master; adding a stable branch is just deferring the time problems occur to the point when master is merged to stable. 

The other posibility is it came from multi-threading. Disable it?

sbird:
Alas, this was with OMP_NUM_THREADS=1 in the first place. I just deleted the debugging statements and it *still* didn't crash, so I'm going to say this was at worst a build system bug.

sbird:
We should also test the feature branches more. You are right. So don't merge this just yet - I will test it more :) But even so if we miss something it is valuable to have an easy way to work out which feature branches introduced the crash, which we can do with a series of stable branches or tags.

rainwoodman:
Let's experiment by starting a stable branch -- could you make it?

sbird:
Sure.

sbird:
I have tested this more thoroughly with a large simulation to low redshift and it works. Let us defer the creation of a stable branch until after this, the scale dependent growth and the intel fixes go in, since andreu seems to want the scale dependent growth and the intel fixes are bugfixes.

andreufont:
@sbird - This issue with different IC velocities, is it common in other codes? Does Gadget-2 or Gadget-3 have the same problem?

sbird:
@andreufont : Yes, it is ubiquitous. It is an issue with the initial conditions solver. The differences are smaller at low redshift and the total power is followed correctly. 

Almost every Lyman alpha forest simulation constraint runs with both the baryons and CDM using the total matter power, which gives extremely similar results for the flux power when normalised to the same mean flux to the currently implemented code. It is likely that the correct velocities will also give very similar answers for the Lyman alpha forest. But it is not certain! If we check and it does not, I will write a letter about it.

PR #206, opened 2018-03-13
Intelfix

sbird:
This makes a whole bunch of shifts unsigned. Apologies for the whitespace changes.

### DIFF
--- a/libgadget/forcetree.c
+++ b/libgadget/forcetree.c
@@ -515,48 +515,49 @@ force_tree_build_single(const struct TreeBuilder tb, const int npart)
  */
 void force_create_node_for_topnode(int no, int topnode, int bits, int x, int y, int z, int *nextfree, const int lastnode)
 {
-    int i, j, k, n, sub, count;
-    MyFloat lenhalf;
+    int i, j, k;
 
-    if(TopNodes[topnode].Daughter >= 0)
-    {
-        for(i = 0; i < 2; i++)
-            for(j = 0; j < 2; j++)
-                for(k = 0; k < 2; k++)
-                {
-                    sub = 7 & peano_hilbert_key((x << 1) + i, (y << 1) + j, (z << 1) + k, bits);
+    /*We reached the leaf of the toptree*/
+    if(TopNodes[topnode].Daughter < 0)
+        return;
 
-                    count = i + 2 * j + 4 * k;
+    for(i = 0; i < 2; i++)
+        for(j = 0; j < 2; j++)
+            for(k = 0; k < 2; k++)
+            {
+                int sub = 7 & peano_hilbert_key((x << 1) + i, (y << 1) + j, (z << 1) + k, bits);
 
-                    Nodes[no].u.suns[count] = *nextfree;
-                    /*We are an internal top level node as we now have a child top level.*/
-                    Nodes[no].f.InternalTopLevel = 1;
+                int count = i + 2 * j + 4 * k;
 
-                    lenhalf = 0.25 * Nodes[no].len;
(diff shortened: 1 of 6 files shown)


libgadget/domain.c
```diff
@@ -317,8 +317,8 @@ static int64_t
 domain_particle_costfactor(int i)
 {
     /* We round off GravCost to integer*/
-    if(P[i].TimeBin)
-        return (1 + P[i].GravCost) * (TIMEBASE / (1 << P[i].TimeBin));
+    if(P[i].TimeBin > 0)
+        return (1 + P[i].GravCost) * (TIMEBASE / (1u << P[i].TimeBin));
```
rainwoodman: TimeBin also need to be cast unsigned, unfortunately...

rainwoodman:
Perhaps define a SHL and SHR macro that always cast both operators?

libgadget/forcetree.c
```diff
@@ -515,7 +515,7 @@ force_tree_build_single(const struct TreeBuilder tb, const int npart)
  */
 void force_create_node_for_topnode(int no, int topnode, unsigned int bits, unsigned int x, unsigned int y, unsigned int z, int *nextfree, const int lastnode)
 {
-    unsigned int i, j, k;
+    int i, j, k;
```
rainwoodman: Why is it important to keep these signed?

libgadget/domain.c
sbird: TimeBin has been made unsigned in the definition in partmanager.h

libgadget/forcetree.c
sbird: Look at line 537: 2*i - 1, which is -1 if i == 0. if i is unsigned that comes out as 2**32 - 1. I am worried about this pull for that reason: there is the potential to introduce subtle bugs worse than the compiler bug we are working around. Please review carefully.

libgadget/timebinmgr.h
```diff
@@ -10,7 +10,7 @@
  *   Note that because each snapshot uses TIMEBASE on the integer timeline, the conversion
  *   factor between loga and ti is not constant across snapshots.
  */
-#define TIMEBINS 20
+#define TIMEBINS 20u
 #define TIMEBASE (1u<<TIMEBINS)
 #define MAXSNAPSHOTS (1u<<(30-TIMEBINS))
```
rainwoodman: 30u?

sbird: Yes! Thank you.

sbird:
I am wary of the macro because I fear that it could create subtle bugs. I am already a little unsettled by how large this one is, for an easily worked around bug.

sbird:
I would appreciate your opinion on whether we should commit a more minimal version of this, actually. The bug seems quite specific and definitely a compiler bug.

rainwoodman:
I definitely felt uncomfortable about the wide spread of 'u' everywhere. (hence the suggestion of using a macro).

Perhaps we can limit this to only the line where the bug is triggered, and add a reference to the ticket on github in the source code?

sbird:
That sounds like a good idea. I will strip this one down.

sbird:
Do you have a link for the upstream icc bug report?

rainwoodman:
No. I don't think we'll have anything, but we can have an issue here, and probably update it when @marcel-sl hear anything back from the site admin.

sbird:
I like that a lot more. Few opportunities for bugs. Let us wait until travis goes green and then commit

rainwoodman:
Yes. It looks good.

PR #207, opened 2018-03-20
Introduce exact tables for short range forces and make them default.

rainwoodman:
[ ] need to commit the routine to make these tables somewhere.

[ ] need to test the effects with a run.


rainwoodman:
The gadget-2 short-range kernel doesn't match the long range kernel when compared with brute force.

It has a systematic offset near the split point. see the attached image around 5.0 mesh units, the erf version has a 3 percent amplification of the short range force compared to the exact value that would have compensated the short range force.

image

image

For these tests I used periodic summation of a plummer kernel (not spline) as the direct force and also as the template of the shortrange force. a single charge is put in the center of box. The cic window is removed. 

I think it has something to do with the finiteness of the mesh, which wasn't explored quite carefully in the gadget paper. 

The potential agrees with direct calculation a lot better except near the center:

image

Notice that potential gradient can be way off even if the potential looked similar.

sbird:
What is the difference between poly and spl? I understand that erf is the kernel used in Gadget-2.

This looks quite big. How much is the short range force contributing at 5 mesh radii?

rainwoodman:
poly and spl are two possible fits of the data. I recall Salman said poly is faster than table look up. But I don't think we want poly because near the cap it is not flat -- would be strange distortions in the tree dominating regime.

The ratio is against the total force. At 5 mesh radii the short range force contribution (after windowing) should be very small ~ a few percent -- reading from the orange line.

sbird:
I see. Small but still not negligible. This is neat. It would be nice to be able to claim better force accuracy than Gadget.

rainwoodman:
Fixing only the aligned directions is not going to be a big deal -- I was
reminded that the non-align directions are probably still wrong..

sbird:
@rainwoodman : It looks in some small tests with different transfer functions that this really does help, so maybe we should merge it? If you can commit or attach the scripts for generating the tables (maybe in tools?) I can take care of writing a force test and checking that everything still works.

sbird:
Hooray!

rainwoodman:
OK. The script is in. I added calculating the anisotropy of the potential and force. This could be useful in deciding Split=1.0 or Split=1.25. It shows that Rcut=4.5 (times Split to get mesh units) is not that great,and erf is not very good neither -- as Split decreases it gets worse.

rainwoodman:
diagonstics-1 50
diagonstics-1 25
diagonstics-1 00
diagonstics-0 75

rainwoodman:
Gadget-paper considered it a zero crossing when erf error is 1 percent;
That's where 1.25 * 4.5 roughly is.

A few more items that I think shall be added to the benchmark:

- Turn off cic deconvolution in the PM solution. Does the anisotropy
improve? The cic decovolution amplifies the grid signal at small scale, and
may increase anisotropy.

- What if the source charge is not sitting right on a grid point? I think
we may need to add that into the scattering, in addition to the anisotropy
for a full evaluation.

- Also currently I am only checking the radial component of the force. In
direct summation and short-range it is aligned to the radial direction, but
the PM force will have some parpendicular component. I think we need to
look at this as well to be extra safe.

I am not so confident about the renormalization of the direct potential.
The direct summation of potential diverges (1 / n) and a constant infinity
needs to be removed.
In 1d, a simple normalization, using 1 / (x + nL) - 1 / nL (when n != 0),
makes it 1 / (n**2)  at large n; I checked this matches the PM solution of
the potential.
However, I do not find a good generalization of this in three-d. If I use 1
/ |nL| like 1d, then there is an offset. Therefore I am currently insisting
the furthest test charge PM and direct sum shall agree. I think the current
renormalizatoin shall be standard, but when I tried to look this up from
the ewald summation paper(s), I couldn't parse them.

rainwoodman:
- No the infinite comes from ewald summation. even if x is not zero at large n the series is asymptotically harmonic. 

I added average over source positions. The mean actually gets closer to erf, though it is still 1 % off at about 2 to 3 mesh cells. 

shifted-diagonstics-1 25
shifted-diagonstics-1 00
shifted-diagonstics-1 50

The current Gadget choice seems to be tolerating maximum 1-sigma anisotropy and truncation error at the same level of about 1%.

rainwoodman:
The Poisson part I have no idea yet, but the differentiation part seems like we can benefit by using the compact finite differentiation: 

https://link.springer.com/article/10.1186/s13662-014-0343-0

See figure 3. Our current method is like the left panel, and you see the anisotropy goes down as the number of mesh points increases, but doesn't converge till about 8 points. This is faster in the right panel, which appears to have converged with 5 mesh points.

Equation 2 is the scheme, but I am not sure if we can write it in fourier space easily. It is also possible equation 4 is not exact. I haven't worked it all out yet. What do you see?

sbird:
Does the exact summation depend strongly on where the charge is in the grid cell? I ask because it looks like the erf accuracy does - I worry that this will be difficult (although for cosmology homogeneity presumably means that averaging is the right thing to do anyway).

For finite differencing: what does Nyx do? I seem to remember that they use a higher order difference kernel in real space. This might be because Zarija tried to put eq. 2 into k space and found it impossible. Or I might be misremembering.

tools/generate-force-kernels.py
```diff
@@ -0,0 +1,383 @@
...
+    pot[inner] = (a**-1 * (
+            -2.8 + u * u * (16./3 + u * u * (6.4 * u - 0.6))
+        ))[inner]
+    pot[mid]  = (a**-1 * (
+            -3.2 + 0.2 / 3 / u + u * u * (32. / 3 + u * (-16.0 + u *(9.6 - 6.4 / 3 * u)))
+        ))[mid]
+
+    return dist * fac[..., None], -pot
+
+def gravity_plummer(dist, a):
```
sbird: I understand where gravity_plummer comes from, but it's not totally obvious from the name. gravity_spline makes sense if you happen to be familiar enough with the Gadget source to know recognise the softening kernel is, but could become confusing otherwise (especially if we change the softening kernel). Maybe you could add some comments?

tools/generate-force-kernels.py
```diff
@@ -0,0 +1,383 @@
...
+
+
+    table = numpy.array([rx, rp_1d, rf_1d, rp_erf, rf_erf]).T
+
+    numpy.savetxt('shortrange-force-kernels.txt',
+        table,
+        header='x(in mesh units) w_pot_1d(x) w_force_1d(x) [erfc + other terms] w_pot_erf(x) w_force_erf(x) split=%.2f' % Split
+    )
+
+    # simply multiple these numbers to the unwindowed short-range force.
```
sbird: Could you add more of a comment here, or a script that turns these output tables into the actual code that sits in gravity.c? Would help to ensure no confusion later.

tools/generate-force-kernels.py
```diff
@@ -0,0 +1,383 @@
...
+
+    def dochunk(x1):
+        dist = []
+
+        for ix, iy, iz in product(*([range(-Nimg, Nimg+1)] * 3)):
+            nL = pm.BoxSize * [ix, iy, iz]
+
+            dist1 = (y + nL)[:, None, :] - x1[None, :, :]
+
+            dist.append(dist1)
```
sbird: I don't understand how this comes from the ewald summation...could you also add a comment here?

sbird:
I went through and read the script and added various comments and cleanups as I understood it. I still don't quite understand how the Nimg argument to force_direct works though...

sbird:
As far as I can see what you have done is fine and a definite improvement, and I have nothing intelligent to say about the questions that are still troubling you. I'm running some longer tests on this.

sbird:
Thanks - I see I was over-thinking the periodic image stuff quite a bit!

sbird:
I've made a lot of runs with this. It definitely works and seems to improve agreement with linear theory/convergence. I still haven't written the force test, but I have been so slow I think we should stop waiting for me. Shall we merge?

rainwoodman:
OK. Let's merge it then. If problem rises we can fix them down the road.

rainwoodman:
[ForceKernels.ipynb.gz](https://github.com/MP-Gadget/MP-Gadget/files/3364957/ForceKernels.ipynb.gz)

Uploading the notebook such that @cheng-zhao can take a look.

PR #209, opened 2018-04-08
Bhdetails

rainwoodman:
This PR dumps more information about the blackholes and do so at every time step. The output is one binary file per rank.

### DIFF
--- a/allvars.h
+++ b/allvars.h
@@ -619,6 +619,7 @@ struct bh_particle_data {
     MyFloat MinPot;
 
     short int TimeBinLimit;
+    MyDouble a; /* scaling factor / time of last update */
 } * BhP;
 
 /*! This structure holds all the information that is
--- a/allvars.h
+++ b/allvars.h
@@ -598,6 +598,7 @@ extern size_t TotalNodeDrifts;
 struct bh_particle_data {
     int ReverseLink; /* used at GC for reverse link to P */
     MyIDType ID; /* for data consistency check, same as particle ID */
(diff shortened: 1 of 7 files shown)


rainwoodman:
Yes it will go to master as well. 
I indeed have recently added a `grow` interface to bigfile, with supporting this use case in mind.
The reason for not using bigfile is because this happens every time step, and I worry bigfile is too slow for this -- opening and closing many disk files frequently.

sbird:
You should probably check that bigfile is too slow...it is a serious usability problem to have multiple output formats, and if we can avoid it we should. If this is intended for debugging only then we should be ok as well.

rainwoodman:
Creating a file can take the order of 1 second on a busy lustre. Then each
data block is at least 3 files.

I think the proper way to write to any storage model that is not one file
per rank is going to be
an infrastructure for collecting / buffering the output, and
notifying the caller for writing collectively when any of the local buffer
is going to be full.

This shall probably go into bigfile, as an extension to the BigArray
interface.

Will there be other use cases than BH details?

sbird:
I think a separate on-disc format is fine if needed as long as the bigfile python routine can read it.

Stars as well as BH would be useful. I was also considering extending the code to allow outputting struct state_of_system on every timestep in an easier format. Plus at some point I wanted to dump selected particles every timestep, which this could be used for.

sbird:
Since this is going to bluetides-ii and we clearly didn't get the time to implement the more general code, do you want to just merge this?

sbird:
Dropping because Yueying cleaned it up and merged something similar.

PR #210, opened 2018-04-10
Cleanups and fixes

sbird:
Some small cleanups from me. The most important one is that the tests should now correctly recompile when doing incremental edits, which makes bisecting much easier.

I also implemented some small micro-ops, which were previously only done at -O3 (and make no difference) and I allocated the kick tables on the heap and made the drifting exact. This marginally changes the drifting, but the previous version was pretty accurate, so it is not a big change.

Tested to make no significant difference to final positions or runtime in a small dm-only run.


libgadget/forcetree.c
```diff
@@ -163,33 +163,25 @@ int force_tree_build(int npart)
...
 {
-    int k;
-    for(k=0; k<3; k++) {
-        double pdiff = 2*(P[p_i].Pos[k] - node->center[k]);
-        pdiff = pdiff > 0 ? pdiff : -1*pdiff;
-        if(pdiff > node->len)
-            return 0;
-    }
-    return 1;
+    int inside =
```
rainwoodman: not a loop because?

rainwoodman: The unrolling produces a faster code? I wonder why the compiler doesn't unroll this if you use a loop.

sbird: The compiler unrolls it at -O3: this just makes the -O3 output work at -O2 and lower.

sbird: Also the use of the ternary operator was a bit messy. It is not a big or important change.

rainwoodman: Could you add a comment about the unrolling such that in the future we don't roll it back?

sbird: Sure.

sbird:
Shall I merge it? Or do you want to look more?

rainwoodman:
Go ahead. Looks good.

sbird:
cheers

PR #212, opened 2018-04-11
Classintegration

sbird:
This integrates class into MP-Gadget, by adding a script to generate transfer functions. There are also the usual subset of small bugfixes. This is the ready-to-merge subset of scaledepgrowth. Yu, please can you review it for merging?


examples/hydro/paramfile.genic
```diff
@@ -30,9 +30,9 @@ FileWithTransferFunction = ../camb_transfer_99.dat  # filename of transfer funct
 InputSpectrum_UnitLength_in_cm = 3.085678e24 # defines length unit of tabulated
                                                                    # input spectrum in cm/h. 
                                     # Note: This can be chosen different from UnitLength_in_cm
-    
 
-PrimordialIndex  = 1. # may be used to tilt the primordial index. Only used if WhichSpectum != 2
+PrimordialAmp = 2.215e-9
+PrimordialIndex  = 0.971 # may be used to tilt the primordial index.
```
rainwoodman: I don't think this is always used? Why is the Only used clause removed? We may want to rename the parameter to reflect that it doesn't always apply!

sbird: Oops, yes, will fix.

gadget/params.c
```diff
@@ -140,10 +140,11 @@ create_gadget_parameter_set()
     param_declare_double(ps, "Omega_fld", OPTIONAL, 0, "Energy density of dark energy fluid.");
     param_declare_double(ps, "w0_fld", OPTIONAL, -1., "Dark energy equation of state.");
     param_declare_double(ps, "wa_fld", OPTIONAL, 0, "Dark energy evolution parameter.");
+    param_declare_double(ps, "Omega_ur", OPTIONAL, 0, "Extra radiation density, eg, a sterile neutrino");
     param_declare_double(ps, "HubbleParam", REQUIRED, 0.697, "");
 
     param_declare_int(ps,    "OutputPotential", OPTIONAL, 1, "Save the potential in snapshots.");
-    param_declare_int(ps,    "MaxMemSizePerNode", OPTIONAL, 0.6 * get_physmem_bytes() / (1024 * 1024), "Preallocate this much memory MB per computing node/ host. Default is 80\% of total physical mem per node. ");
+    param_declare_double(ps,    "MaxMemSizePerNode", OPTIONAL, 0.6, "Preallocate this much memory per computing node/ host. If < 1, in fraction of total physical mem (default 0.6), otherwise in MB. ");
```
rainwoodman: rename it to fraction. I still think using bytes / MB is more reasonable because if one want to specify a number than the total available memory is already known. And usually it is Total - ReservedForOS. And also consider the case where we wrongly detected the total mem. A fraction we are totally stuck without workaround.

gadget/params.c
```diff
@@ -199,11 +200,11 @@ create_gadget_parameter_set()
     param_declare_double(ps, "PartAllocFactor", REQUIRED, 0, "");
     param_declare_double(ps, "TopNodeAllocFactor", OPTIONAL, 0.5, "");
 
-    param_declare_double(ps, "InitGasTemp", OPTIONAL, 300, "Initial gas temperature");
+    param_declare_double(ps, "InitGasTemp", OPTIONAL, -1, "Initial gas temperature. By default set to CMB temperature.");
```
rainwoodman: CMB temp at the redshift?

libgadget/Makefile
```diff
@@ -139,10 +138,16 @@ libgadget-utils.a: $(GADGET_UTILS_OBJS)
 libgadget.a: $(GADGET_OBJS)
 	$(AR) rv $@ $(GADGET_OBJS)
 
+.objs/petaio.o: petaio.c config.h $(INCL)
```
rainwoodman: Looks like a workaround. Why?

libgadget/tests/test_hci.c
```diff
@@ -156,16 +156,16 @@ test_hci_terminate(void ** state)
...
     message(0, "UsingPrefix : '%s'\n", prefix);
-
+    return !ret;
 }
 
-static void teardown(void ** state)
+static int teardown(void ** state)
 {
-    remove(prefix);
+    return remove(prefix);
```
rainwoodman: likely non-resursive and will fail.

libgenic/save.c
```diff
@@ -12,8 +12,9 @@
 #include <libgadget/utils.h>
 #include <libgadget/cosmology.h>
 #include <libgadget/walltime.h>
+#include <libgadget/utils/mymalloc.h>
 
-static void saveblock(BigFile * bf, void * baseptr, int ptype, char * bname, char * dtype, int items_per_particle, int64_t TotNumPart) {
+static void saveblock(BigFile * bf, void * baseptr, int ptype, char * bname, char * dtype, int items_per_particle, ptrdiff_t structsz, int64_t TotNumPart) {
```
rainwoodman: I usually call this term elsize (element size). I forgot where it came from, but it looks better than structsz.

tools/make_class_power.py
```diff
@@ -0,0 +1,243 @@
...
+            tfile = os.path.join(sdir, config['FileWithTransferFunction']+"-"+str(red))
+            save_transfer(trans, tfile, bg, red)
+            #Get and save the matter power spectrum
+            pk_lin = powspec.get_pklin(k=trans['k'], z=red)
+            pkfile = os.path.join(sdir, config['FileWithInputSpectrum']+"-"+str(red))
+            if os.path.exists(pkfile):
+                raise IOError("Refusing to write to existing file: ",pkfile)
+            np.savetxt(pkfile, np.vstack([trans['k'], pk_lin]).T)
+
+def save_transfer(transfer, transferfile, bg, redshift):
```
rainwoodman: Better to save the velocity here as well. I find the interpolation code in C a bit hard to understand. Perhaps a next PR.

rainwoodman:
OK. Mostly trivial suggestions here.

tools/make_class_power.py
sbird: Yes, that is next PR. This extra transfer function is not used here yet anyway.

gadget/params.c
sbird: This is not always a fraction: it is a fraction only if less than one. I will adjust the docstring.

gadget/params.c
sbird: Fixed, thanks.

libgadget/Makefile
sbird: Ah - it is to ensure that petaio.o and only petaio gets rebuilt when config.h changes, because petaio.c is the file that includes config.h . It is just to declare the extra dependency. I will add a comment.

libgadget/tests/test_hci.c
sbird: You mean that the remove call will fail if the directory is non-empty? I think that's ok - the subfiles should be removed by the HCI manager, shouldn't they? This checks that they are.

sbird:
Ok - I will merge it when travis goes green, and continue working with a rebased scaledepgrowth branch.

sbird:
(missed one)

libgadget/Makefile
rainwoodman: But all .o files already depend on $(INCL),  which contains config.h see Makefile.rules.

libgadget/tests/test_hci.c
rainwoodman: Sure. But that means it is a test. I think it is non-critical to remove or not. Sometimes people do `system("rm -rf")` for the clean up.

sbird: Fair enough, I can change it to remove(prefix); return 0;

libgadget/Makefile
sbird: Not any more - I removed config.h from $(INCL) so that one can meaningfully do incremental builds. Otherwise every time you do a commit it will force config.h to be regenerated, which will force a full rebuild.

PR #213, opened 2018-04-12
Moregenic

sbird:
A few more MP-GenIC fixes split out from scaledepgrowth.

f16eaac we decided to change after discussion with Andreu. 
d18b81b is a simple threading on MP-GenIC, and 6258764 adjusts some of the memory management in MP-GenIC to make it a bit clearer.

### DIFF
--- a/libgenic/power.c
+++ b/libgenic/power.c
@@ -10,8 +10,8 @@
 
 #include "power.h"
 
-static double PowerSpec_EH(double k);
-static double PowerSpec_Tabulated(double k, int Type);
+static double Delta_EH(double k);
+static double Delta_Tabulated(double k, int Type);
 static double sigma2_int(double k, void * params);
 static double TopHatSigma2(double R);
 static double tk_eh(double k);
@@ -36,14 +36,14 @@ struct table
 static struct table power_table;
 static struct table transfer_table;
 
-double PowerSpec(double k, int Type)
+double DeltaSpec(double k, int Type)
 {
   double power;
 
   if(WhichSpectrum == 2)
-      power = PowerSpec_Tabulated(k, Type);
+      power = Delta_Tabulated(k, Type);
   else
-      power = PowerSpec_EH(k);
+      power = Delta_EH(k);
 
   /*Normalise the power spectrum*/
   power *= Norm;
@@ -71,7 +71,8 @@ void parse_power(int i, double k, char * line, struct table *out_tab, int * Inpu
     if ((*InputInLog10) == 0)
         p = log10(p);
     p += 3 * log10(scale);	/* convert to Kpc/h  */
-    out_tab->logD[0][i] = p;
+    /*Store delta, square root of power*/
+    out_tab->logD[0][i] = p/2;
 }
 
 void parse_transfer(int i, double k, char * line, struct table *out_tab, int * InputInLog10, double scale)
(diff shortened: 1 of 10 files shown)


sbird:
Yu, can you do me a quick review? Thanks.

libgenic/zeldovich.c
```diff
@@ -158,15 +159,20 @@ void displacement_fields(int Type) {
...
     petapm_force_c2r(rho_k, regions, functions);
+    myfree(rho_k);
+    myfree(regions);
     petapm_force_finish();
     double maxdisp = 0;
     int i;
-    for(i = 0; i < NumPart; i ++) {
+    #pragma omp parallel for reduction(max:maxdisp)
+    for(i = 0; i < NumPart; i++)
+    {
```
rainwoodman: Could you by the way make the `MPI_Reduce` a few lines lower in place? these one time variables are really eyesores.

libgenic/zeldovich.c
```diff
@@ -301,7 +307,6 @@ typedef struct {
         ptrdiff_t total;
     } ORegion;
     int Nmesh[3];
-    size_t allocsize;
```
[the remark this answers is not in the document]
sbird: Notice that fftsize is only used in petapm.c - this was previously only used for the memset. Once the memset is moved into petapm.c there is no need to store the total size at all

libgenic/zeldovich.c
sbird: I'll pull forward another part of the cleanup from the scaledepgrowth branch.

sbird: How do you like it now?

libgenic/zeldovich.c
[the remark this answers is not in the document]
sbird: I too would prefer to store allocsize in a local struct, but there isn't one handy for the gravpm user of petapm, unfortunately.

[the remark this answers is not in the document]
rainwoodman: OK then.

libgenic/power.c
```diff
@@ -204,17 +205,18 @@ double PowerSpec_Tabulated(double k, int Type)
...
   double trans = 1;
   /*Transfer table stores (T_type(k) / T_tot(k))^2*/
-  if(Type >= 0 && Type < MAXCOLS && transfer_table.Nentry > 0)
+  if(Type >= 0 && Type < MAXCOLS && transfer_table.Nentry > 0) {
       trans = gsl_interp_eval(transfer_table.mat_intp[Type], transfer_table.logk, transfer_table.logD[Type], logk, transfer_table.mat_intp_acc[Type]);
+  }
 
   double power = pow(10.0, logD) * trans;
 
   return power;
```
rainwoodman: power or delta?

libgenic/zeldovich.c
```diff
@@ -153,54 +154,52 @@ void displacement_fields(int Type) {
...
-    {
-        int k;
-        for(k = 0; k < 3; k++)
-        {
             ICP[i].Pos[k] += ICP[i].Vel[k];
             ICP[i].Vel[k] *= vel_prefac;
             ICP[i].Pos[k] = periodic_wrap(ICP[i].Pos[k]);
         }
     }
+    MPI_Reduce(&maxdisp, &maxdispall, 1, MPI_DOUBLE, MPI_MAX, 0, MPI_COMM_WORLD);
```
rainwoodman: I meant MPI_Allreduce(MPI_IN_PLACE, &maxdisp, 1, .....)

sbird: Oh, sorry. Fixed.

libgenic/power.c
sbird: Fixed

sbird:
You notice I also fixed the build system petaio rule from before: you were right that it was a hack. You can still do incremental builds because it doesn't check the git revision except when the build files changes.

sbird:
Ok to merge? Thanks for the quick reviews, by the way.

PR #214, opened 2018-04-16
Added linear growth test in example sims

Chris-Pedersen:
I have included the CAMB data for the given parameters at each snapshot redshift


examples/linear_growth/linear_growth_params.ini
```diff
@@ -0,0 +1,82 @@
+output_root = linear_growth
```
sbird: You don't need this file - it's just a copy of the input, which you also have below.

sbird:
The pull request is fine, but I am about to break some things, so hold off on it for now.

sbird:
Can you please re-run on the latest scaledepgrowth branch? Note that the code there has changed to require class-formatted transfer functions, so you will have to switch your ICs to class. You must also use "extra metric transfer functions = y". If you are confused, use the make_class_power.py script in tools/

sbird:
Actually, to avoid cluttering the repo with tons of matter power spectra files, could you rework the test to run tools/make_class_power.py in run.sh? You can use the --extraz parameter to generate transfer functions for lower redshifts, and base it on the version in scaledepgrowth. Thanks.

sbird:
You need to pip install classylss, which is a LBNL class wrapper. Yu, is there a reason we use it instead of class directly?

rainwoodman:
The class binding from class used to be very fragile.

conda install -c bccp classylss will give you classylss if you use conda.
otherwise pip install it.

Chris-Pedersen:
Ok its now set up to generate power spectra and transfer functions using[ make_class_power.py from the scaledepgrowth branch](https://github.com/sbird/MP-Gadget3/blob/scaledepgrowth/tools/make_class_power.py), so no 'InputFutureTransferFunction'. Let me know if this is alright!

sbird:
Merged, thanks!

PR #215, opened 2018-05-05
Small fixes and a small script

sbird:
This fixes a type definition in gravshort, which could have led the force softening to be truncated if we every used mixed softenings (but we don't).

It also removes the drift_particle from gravshort, which did nothing but was surprisingly hot in a profile.

Finally it adds a short script which uses nbodykit to plot the structure in a box. Just to have it handy.


sbird:
Merge this?

PR #219, opened 2018-06-06
Spinlockfix

### LINKED ISSUE #217: Compilation without OPENMP_USE_SPINLOCK
I noticed that if you do not define OPENMP_USE_SPINLOCK there is a compilation error:

/home/spb/codes/MP-Gadget/libgadget/fof.c:446: undefined reference to `lock_particle'
/home/spb/codes/MP-Gadget/libgadget/fof.c:453: undefined reference to `unlock_particle'

either we should provide workable definitions for lock_particle and unlock_particle in drift.c even without spinlocks , or we should force OPENMP_USE_SPINLOCK to be always on and remove the define.

Relatedly, we found that mac does not define pthread_spinlock_t which makes OPENMP_USE_SPINLOCK fail. If we decide to make this compiel flag always on, we should provide an alternative implementation, perhaps typedeffing pthread_spinlock_t to pthread_mutex_t , or providing a stub implementation, or using omp_set_lock instead.

Yu, what do you think?

sbird:
This commit fixes #217 and should fix compilation on mac. OPENMP_USE_SPINLOCK is replaced by NO_OPENMP_SPINLOCK, which compiles and enforces that the code runs in single-threaded mode. Most of the diff is actually cleanups.

### DIFF
--- a/Options.mk.example
+++ b/Options.mk.example
@@ -17,8 +17,9 @@ OPT += -DDENSITY_INDEPENDENT_SPH
 #OPT += -DLIGHTCONE                       # write a lightcone on the fly; in development
 #OPT += VALGRIND     # allow debugging with valgrind, disable the GADGET memory allocator.
 #OPT += -DDEBUG      # print a lot of debugging messages
-
 OPT += -DSPH_GRAD_RHO  # calculate grad of rho in SPH, required for Krumholtz & Gnedin H2 SFR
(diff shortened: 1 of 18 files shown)


rainwoodman:
i think this is a reasonable approach. the current threading requires
spinlocks.

Yu

sbird:
Thanks, merging.

PR #220, opened 2018-06-07
Fix the linear response neutrino code for class formatted transfer functions

sbird:
This PR fixes the linear response neutrino code, which was broken when we moved to class formatted transfer functions. It also improves the accuracy of the make_class_power.py script for neutrinos (which was the hardest part!).

I took the opportunity to fix a few niggles with the linear response code - in particular the neutrino power spectrum now has modes-per-bin associated with it. I also store the initial transfer functions and matter power spectrum in the IC bigfile, so we don't need to reload them from a text file on disc when starting a simulation. I saved all the transfer functions in case we want to use them for something else later. Note that the IC transfer and matter power are only stored in the initial conditions, not the later snapshots. Let me know if you would like me to change this.

Since it turned out that after this we were only using a small core of kspace-neutrinos, I moved the remaining code directly into MP-Gadget, with tests, and removed the submodule. This should make it easier to work with in future.

The LinearTransferFunction parameter now has no effect: I left the parameter in there so it was easy to switch between branches while developing, but I will remove it after this is merged.

There are also some cleanups: the speed of light is now LIGHTCGS, not 'C', which was impossible to grep for.

@andreufont : This is a good base to run neutrino simulations on dirac from. I have tested that the kspace integrator works the same as it did before.

### DIFF
--- a/gadget/params.c
+++ b/gadget/params.c
@@ -490,7 +490,6 @@ void read_parameter_file(char *fname)
         /*Massive neutrino parameters*/
         All.MassiveNuLinRespOn = param_get_int(ps, "MassiveNuLinRespOn");
         All.HybridNeutrinosOn = param_get_int(ps, "HybridNeutrinosOn");
-        All.CAMBInputSpectrum_UnitLength_in_cm = param_get_double(ps, "InputSpectrum_UnitLength_in_cm");
         All.CP.MNu[0] = param_get_double(ps, "MNue");
         All.CP.MNu[1] = param_get_double(ps, "MNum");
         All.CP.MNu[2] = param_get_double(ps, "MNut");
--- a/gadget/params.c
+++ b/gadget/params.c
@@ -300,7 +300,7 @@ create_gadget_parameter_set()
     param_declare_int(ps, "MassiveNuLinRespOn", REQUIRED, 0, "Enables linear response massive neutrinos of 1209.0461. Make sure you enable radiation too.");
     param_declare_int(ps, "HybridNeutrinosOn", OPTIONAL, 0, "Enables hybrid massive neutrinos, where some density is followed analytically, and some with particles. Requires MassivenuLinRespOn");
     param_declare_string(ps, "LinearTransferFunction", OPTIONAL, "camb_transfer_99.dat", "File containing linear transfer function in CAMB format. Used for massive neutrinos.");
-    param_declare_double(ps, "InputSpectrum_UnitLength_in_cm", OPTIONAL, 3.085678e24, "Units of the CAMB transfer function in cm. By default Mpc.");
+    param_declare_double(ps, "InputSpectrum_UnitLength_in_cm", OPTIONAL, CM_PER_MPC, "Units of the CAMB transfer function in cm. By default Mpc.");
     param_declare_double(ps, "MNue", OPTIONAL, 0, "First neutrino mass in eV.");
     param_declare_double(ps, "MNum", OPTIONAL, 0, "Second neutrino mass in eV.");
     param_declare_double(ps, "MNut", OPTIONAL, 0, "Third neutrino mass in eV.");
--- a/gadget/params.c
+++ b/gadget/params.c
@@ -300,7 +300,7 @@ create_gadget_parameter_set()
     param_declare_int(ps, "MassiveNuLinRespOn", REQUIRED, 0, "Enables linear response massive neutrinos of 1209.0461. Make sure you enable radiation too.");
     param_declare_int(ps, "HybridNeutrinosOn", OPTIONAL, 0, "Enables hybrid massive neutrinos, where some density is followed analytically, and some with particles. Requires MassivenuLinRespOn");
     param_declare_string(ps, "LinearTransferFunction", OPTIONAL, "camb_transfer_99.dat", "File containing linear transfer function in CAMB format. Used for massive neutrinos.");
-    param_declare_double(ps, "InputSpectrum_UnitLength_in_cm", OPTIONAL, CM_PER_MPC, "Units of the CAMB transfer function in cm. By default Mpc.");
+    param_declare_double(ps, "InputSpectrum_UnitLength_in_cm", OPTIONAL, CM_PER_MPC, "Units of the CAMB transfer function in cm/h. By default Mpc/h.");
     param_declare_double(ps, "MNue", OPTIONAL, 0, "First neutrino mass in eV.");
     param_declare_double(ps, "MNum", OPTIONAL, 0, "Second neutrino mass in eV.");
     param_declare_double(ps, "MNut", OPTIONAL, 0, "Third neutrino mass in eV.");
--- a/gadget/params.c
+++ b/gadget/params.c
@@ -490,7 +490,6 @@ void read_parameter_file(char *fname)
         /*Massive neutrino parameters*/
         All.MassiveNuLinRespOn = param_get_int(ps, "MassiveNuLinRespOn");
         All.HybridNeutrinosOn = param_get_int(ps, "HybridNeutrinosOn");
-        param_get_string2(ps, "LinearTransferFunction", All.CAMBTransferFunction);
         All.CAMBInputSpectrum_UnitLength_in_cm = param_get_double(ps, "InputSpectrum_UnitLength_in_cm");
         All.CP.MNu[0] = param_get_double(ps, "MNue");
         All.CP.MNu[1] = param_get_double(ps, "MNum");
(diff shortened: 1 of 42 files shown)


rainwoodman:
Is the bigfile update intentional?

gadget/params.c
```diff
@@ -300,7 +300,7 @@ create_gadget_parameter_set()
     param_declare_int(ps, "MassiveNuLinRespOn", REQUIRED, 0, "Enables linear response massive neutrinos of 1209.0461. Make sure you enable radiation too.");
     param_declare_int(ps, "HybridNeutrinosOn", OPTIONAL, 0, "Enables hybrid massive neutrinos, where some density is followed analytically, and some with particles. Requires MassivenuLinRespOn");
     param_declare_string(ps, "LinearTransferFunction", OPTIONAL, "camb_transfer_99.dat", "File containing linear transfer function in CAMB format. Used for massive neutrinos.");
-    param_declare_double(ps, "InputSpectrum_UnitLength_in_cm", OPTIONAL, 3.085678e24, "Units of the CAMB transfer function in cm. By default Mpc.");
+    param_declare_double(ps, "InputSpectrum_UnitLength_in_cm", OPTIONAL, CM_PER_MPC, "Units of the CAMB transfer function in cm. By default Mpc.");
```
rainwoodman: Mpc/h or Mpc?

rainwoodman: Units of the CAMB transfer function in cm/h, by default Mpc/h?

libgenic/power.c
```diff
@@ -149,15 +206,25 @@ void parse_transfer(int i, double k, char * line, struct table *out_tab, int * I
...
+    /*Should be weighted bu omega_nu*/
+    out_tab->logD[DELTA_NU][i] /= onu;
     /*h_prime is entry 8 + nnu. t_b is 12 + nnu, t_ncdm[2] is 13 + nnu * 2.*/
     out_tab->logD[VEL_BAR][i] = transfers[12+nnu];
     out_tab->logD[VEL_CDM][i] = transfers[8+nnu] * 0.5;
-    out_tab->logD[VEL_NU][i] = transfers[13+nnu*2];
+    out_tab->logD[VEL_NU][i] = 0;
+    for(j=0; j < nnu; j++)
+        out_tab->logD[VEL_NU][i] = transfers[13 + nnu + j] * omega_nu_single(Onu, InitTime, j);
+    /*Should be weighted bu omega_nu*/
```
rainwoodman: by

libgenic/power.c
```diff
@@ -149,15 +206,25 @@ void parse_transfer(int i, double k, char * line, struct table *out_tab, int * I
...
+    out_tab->logD[VEL_NU][i] /= onu;
 }
 
-void read_power_table(int ThisTask, const char * inputfile, const int ncols, struct table * out_tab, double scale, const int nnu, void (*parse_line)(int i, double k, char * line, struct table *, int *InputInLog10, const int nnu, double scale))
+void read_power_table(int ThisTask, const char * inputfile, const int ncols, struct table * out_tab, const double InitTime, void (*parse_line)(int i, double k, char * line, struct table *, int *InputInLog10, const double InitTime))
```
rainwoodman: What about taking out (*parse_line) with a typedef? This prototype is getting too long.

libgenic/power.c
```diff
@@ -322,12 +392,13 @@ int init_powerspectrum(int ThisTask, double InitTime, double UnitLength_in_cm_in
 
 double Delta_Tabulated(double k, int Type)
```
rainwoodman: Type as an enum here?

tools/make_class_power.py
```diff
@@ -108,6 +109,11 @@ def _build_cosmology_params(config):
         gparams['tol_ncdm_synchronous'] = 1e-5
         gparams['tol_ncdm_bg'] = 1e-10
         gparams['l_max_ncdm'] = 50
+        #This disables the fluid approximations, which make P_nu not match camb on small scales.
+        #We need accurate P_nu to initialise our neutrino code.
```
rainwoodman: You mentioned fluid approximation is faster? Why is it 'needed' to have accurate P_nu at small scales for the initialization?

tools/make_class_power.py
```diff
@@ -157,6 +163,10 @@ def make_class_power(paramfile, external_pk = None, extraz=None):
     powerparams = {'output': 'dTk vTk mPk', 'P_k_max_h/Mpc' : maxk, "z_max_pk" : 1+np.max(outputs),'z_pk': outputs, 'extra metric transfer functions': 'y'}
     pre_params.update(powerparams)
 
+    if verbose:
+        verb_params = {'input_verbose': 1, 'background_verbose': 1, 'thermodynamics_verbose': 1, 'perturbations_verbose': 1, 'transfer_verbose': 1, 'primordial_verbose': 1, 'spectra_verbose': 1, 'nonlinear_verbose': 1, 'lensing_verbose': 1, 'output_verbose': 1}
+        pre_params.update(verb_params)
+
     #Specify an external primordial power spectrum
     if external_pk is not None:
```
rainwoodman: This is fine, but I think it would be nicer to bypass the pk code and directly multiply transfer to the provided pkprim instead? It feels cleaner than doing the command; I can keep it on my todo list.

sbird: That's a good idea, would simplify the transfer function code a bit. We would need to set whichever class setting it is to output the primordial power.

sbird:
The bigfile update was an accident and is actually a downgrade - do you want me to rebase it out, or add an extra commit updating to master?

gadget/params.c
sbird: Fixed, thanks.

libgenic/power.c
sbird: oops, fixed.

libgenic/power.c
sbird: Done

tools/make_class_power.py
```diff
@@ -99,6 +99,7 @@ def _build_cosmology_params(config):
     if omeganu > 0:
         gparams['m_ncdm'] = '%.2f,%.2f,%.2f' % (config['MNue'], config['MNum'], config['MNut'])
         gparams['N_ncdm'] = 3
+        gparams['N_ur'] = 0.00641
```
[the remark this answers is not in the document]
sbird: That is the minimal neutrino mass, I think. Two massless species, one massive species. It's a very specific case: the known neutrino hierarchy (mass species split by 0.05 and 0.01) means that most interesting cases have three massive neutrinos.

libgenic/power.c
sbird: Very good idea, done!

libgenic/save.c
```diff
@@ -14,19 +14,16 @@
 #include <libgadget/walltime.h>
 #include <libgadget/utils/mymalloc.h>
 
-static void saveblock(BigFile * bf, void * baseptr, int ptype, char * bname, char * dtype, int items_per_particle, ptrdiff_t elsize, int64_t TotNumPart) {
+void saveblock_name(BigFile * bf, void * baseptr, char * name, char * dtype, size_t dims[], ptrdiff_t elsize, int64_t TotNumPart)
```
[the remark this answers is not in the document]
sbird: I don't understand what you mean about deriving TotNumPart, but I have done the rename.

sbird:
Thanks for the review! Rebased to remove the bigfile change and I think I addressed all the comments. Shall I merge when travis goes green?

rainwoodman:
That's what I suspected. rebase would work. Maybe we shall detach bigfile
as a submodule too, since relevant updates is rare.

rainwoodman:
Not worth it. But print a message when the script starts, and say it may
take about 8 minutes to obtain an accurate neutrino solution? (so people
don't think it has hang).

libgenic/save.c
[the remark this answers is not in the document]
rainwoodman: would
```
MPI_Allreduce(&dims[0], &TotNumPart, 1, MPI_LONG, MPI_SUM, comm)
```
produce the number?

(I never recall if we had defined MPI_PTRDIFF and MPI_INT64, btw.)

tools/make_class_power.py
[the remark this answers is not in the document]
rainwoodman: OK. We need to keep this in mind.

sbird:
Sounds good, I'll add that to the next batch of updates.

libgenic/save.c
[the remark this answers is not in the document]
sbird: Oh, I see. Yes, it would, and I'll add that to the next pull as well.

andreufont:
Hi @sbird - I just wanted to let you know that Chris is actively looking at this, and we'll keep you posted with the results from his tests.

PR #222, opened 2018-06-13
Cleanups

sbird:
Cleanups, mostly suggested by @rainwoodman  in the last PR, and fix compilation on mac.

### DIFF
--- a/genic/params.c
+++ b/genic/params.c
@@ -131,7 +131,6 @@ void read_parameterfile(char *fname)
     All2.PowerP.DifferentTransferFunctions = param_get_int(ps, "DifferentTransferFunctions");
     All2.PowerP.ScaleDepVelocity = param_get_int(ps, "ScaleDepVelocity");
     All2.PowerP.WhichSpectrum = param_get_int(ps, "WhichSpectrum");
-    All2.PowerP.SpectrumLengthScale = CM_PER_MPC / All.UnitLength_in_cm;
     All2.PowerP.PrimordialIndex = param_get_double(ps, "PrimordialIndex");
 
     /*Simulation parameters*/
--- a/libgenic/power.c
+++ b/libgenic/power.c
@@ -14,7 +14,7 @@
 #include "power.h"
 
 /*Defined in save.c*/
-void _bigfile_utils_create_block_from_c_array(BigFile * bf, void * baseptr, char * name, char * dtype, size_t dims[], ptrdiff_t elsize, int64_t TotNumPart, MPI_Comm comm);
+void _bigfile_utils_create_block_from_c_array(BigFile * bf, void * baseptr, char * name, char * dtype, size_t dims[], ptrdiff_t elsize, MPI_Comm comm);
 
 static double Delta_EH(double k);
 static double Delta_Tabulated(double k, enum TransferType Type);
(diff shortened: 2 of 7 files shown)


Makefile.rules
```diff
@@ -11,6 +11,11 @@ CFLAGS += -I../depends/include
 CFLAGS += -I../
 CFLAGS += "-DLOW_PRECISION=$(LOW_PRECISION)"
 CFLAGS += "-DHIGH_PRECISION=$(HIGH_PRECISION)"
+#This fixes linker errors on Mac.
+#Without it uninitialised variables are not included in static libraries.
+ifeq ($(shell uname -s),Darwin)
+    CFLAGS += "-fno-common"
```
rainwoodman: I find this a bit dangerous. It seems to suggest we have redudant definitions of global variables. See here:

http://www.keil.com/support/man/docs/armclang_ref/armclang_ref_chr1422532346348.htm

rainwoodman: Which variable were giving you trouble?

sbird: It is because common block variables (by default non-initialised globals, in particular the All structure) are not exported by default in Mac's static library format (which is not ELF), so when it comes to linking the static library into the executable you get an unresolved error. This is only on Mac - it does not apply on linux at all.

The error was:
mpicc -fopenmp -O3 -g -Wall -ffast-math -march=native .objs/main.o .objs/params.o ../libgadget/libgadget.a ../libgadget/libgadget-utils.a -lm -L/usr/local/lib -lgsl -lgslcblas -lm -L../depends/lib -lmpsort-mpi -lradixsort -lbigfile-mpi -lbigfile -lpfft_omp -lfftw3_mpi -lfftw3_omp -lfftw3 -o MP-Gadget
Undefined symbols for architecture x86_64:
  "_All", referenced from:
      _main in main.o
      _OutputListAction in params.o
      _read_parameter_file in params.o
      _write_cpu_log in libgadget.a(run.o)
      _begrun in libgadget.a(run.o)
      _run in libgadget.a(run.o)
      _density_kernel_init in libgadget.a(densitykernel.o)
      ...
     (maybe you meant: _MPI_Alltoallv_smart, _MPI_Alltoallv_sparse )

Variables in the common block were:
All
Fd{CPU,BlackHole, etc}
NTask
ThisTask
ie, everything in allvars.c

Anything initialised to zero is exported, so one fix is to just change allvars.c to initialise everything to zero, as is done in, eg, partmanager.c Another fix would be to tell mac to explicitly export the common variables, but I wasn't sure how to do that without labelling all the variables, which is certain to break at some point.

What this does is to place the uninitialised variables in a block which is exported. Kind of a hack, but the least-invasive way to support Mac I could think of.

rainwoodman: At least this is not due to redundant definitions.

either adding -c to ranlib or -fno-common will work.

https://public.kitware.com/Bug/view.php?id=10342

rainwoodman: I think it is cleaner to have these in the local Makefile for OSX, rather than having too many if conditions in the main Makefile. We shall simply ship an example file that compiles cleanly on OSX.

sbird: Ok, I'll commit that and merge this. Thanks.

sbird:
Travis has broken for I think unrelated reasons - a dask update seems to have broken nbodykit in some way.

rainwoodman:
Yes. It is this.
https://github.com/dask/dask/issues/3631

PR #223, opened 2018-06-14
Treewalk opt

sbird:
Treewalk optimisations for high thread count, especially knights landing

I ran the code on Stampede's knight's landing nodes. Our performance kind of sucks (perhaps half of a normal intel skylake node). Part of this is because I was using a very small problem size, but part of it is that our code needs to be threaded better. I was able to get an extra 10% of the total runtime by careful massaging of the tree: in particular I made sure that the WorkSet queue preserves the order of particles in memory.

Making the ActiveParticle list was also a hot-spot at high thread count: I parallelized it and was able to skip it entirely during a PM step. Skipping it entirely during a PM step saved memory but didn't noticeably decrease CPU time, probably because a PM step is relatively rare.

Tested that this gives almost identical output to master on both dm-only and hydro examples.

### DIFF
--- a/libgadget/init.c
+++ b/libgadget/init.c
@@ -185,6 +185,7 @@ setup_smoothinglengths(int RestartSnapNum)
 {
     int i;
 
+    force_tree_rebuild();
     if(RestartSnapNum == -1)
     {
 #pragma omp parallel for
@@ -317,4 +318,6 @@ setup_smoothinglengths(int RestartSnapNum)
 #ifdef DENSITY_INDEPENDENT_SPH
     density_update();
 #endif //DENSITY_INDEPENDENT_SPH
+
+    if(force_tree_allocated()) force_tree_free();
 }
--- a/libgadget/init.c
+++ b/libgadget/init.c
@@ -126,7 +126,9 @@ void init(int RestartSnapNum)
 
     domain_decompose_full();	/* do initial domain decomposition (gives equal numbers of particles) */
 
-    rebuild_activelist(0);
+    /*At the first time step all particles should be active*/
+    ActiveParticle = NULL;
+    NumActiveParticle = PartManager->NumPart;
 
     setup_smoothinglengths(RestartSnapNum);
 }
--- a/libgadget/run.c
+++ b/libgadget/run.c
@@ -253,11 +253,12 @@ void run(void)
         /* Update velocity to the new step, with the newly computed step size */
         apply_half_kick();
 
-        free_activelist();
-
         if(is_PM) {
             apply_PM_half_kick();
         }
+
+        /* We can now free the active list: the new step have new active particles*/
+        free_activelist();
     }
 
     close_outputfiles();
--- a/libgadget/run.c
+++ b/libgadget/run.c
@@ -249,6 +249,8 @@ void run(void)
         /* Update velocity to the new step, with the newly computed step size */
         apply_half_kick();
 
+        free_activelist();
+
         if(is_PM) {
             apply_PM_half_kick();
         }
--- a/libgadget/run.c
+++ b/libgadget/run.c
@@ -181,9 +181,7 @@ void run(void)
             domain_maintain();
         }
 
-        rebuild_activelist(All.Ti_Current);
-
-        print_timebin_statistics(NumCurrentTiStep);
+        rebuild_activelist(All.Ti_Current, NumCurrentTiStep);
 
         set_random_numbers(All.RandomSeed + All.Ti_Current);
 
--- a/libgadget/run.c
+++ b/libgadget/run.c
@@ -219,12 +219,12 @@ void run(void)
 
         if(WriteSnapshot) {
             /* The accel may have created garbage -- collect them before writing a snapshot.
-             * If we do collect, rebuild tree and active list.*/
+             * If we do collect, rebuild tree and reset active list size.*/
             int compact[6] = {0};
 
             if(slots_gc(compact)) {
                 force_tree_rebuild();
-                rebuild_activelist(All.Ti_Current);
+                NumActiveParticle = PartManager->NumPart;
             }
         }
 
--- a/libgadget/run.c
+++ b/libgadget/run.c
@@ -185,6 +185,9 @@ void run(void)
 
         set_random_numbers(All.RandomSeed + All.Ti_Current);
 
+        /* Need to rebuild the force tree because all TopLeaves are out of date.*/
+        force_tree_rebuild();
+
         /* update force to Ti_Current */
         compute_accelerations(is_PM, NumCurrentTiStep == 0, GasEnabled);
 
@@ -233,6 +236,9 @@ void run(void)
         NumCurrentTiStep++;
 
         report_memory_usage("RUN");
+
+        /*Note FoF will free the tree too*/
+        if(force_tree_allocated()) force_tree_free();
 
         if(!next_sync || stop) {
             /* out of sync points, or a requested stop, the run has finally finished! Yay.*/
--- a/libgadget/slotsmanager.c
+++ b/libgadget/slotsmanager.c
@@ -225,9 +225,16 @@ static int slot_cmp_reverse_link(const void * b1in, const void * b2in) {
 static int
 slots_gc_mark()
 {
-    int i;
+    int i, ptype;
+    int gc_slots = 0;
+    for(ptype = 0; ptype < 6; ptype ++)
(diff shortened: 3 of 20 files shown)


sbird:
Rebased to remove accidental bigfile update. I have another branch pending with some threading for the exchange code.

libgadget/init.c
```diff
@@ -315,4 +318,6 @@ setup_smoothinglengths(int RestartSnapNum)
 #ifdef DENSITY_INDEPENDENT_SPH
     density_update();
 #endif //DENSITY_INDEPENDENT_SPH
+
+    if(force_tree_allocated()) force_tree_free();
```
rainwoodman: You wanted to limit the life cycle of tree inside this function?
force_tree_allocated should be true here.

libgadget/timestep.c
```diff
@@ -686,36 +686,85 @@ inttime_t find_next_kick(inttime_t Ti_Current, int minTimeBin)
     return Ti_Current + dti_from_timebin(minTimeBin);
 }
 
+static void print_timebin_statistics(int NumCurrentTiStep, int * TimeBinCountType);
+
 /* mark the bins that will be active before the next kick*/
-int rebuild_activelist(inttime_t Ti_Current)
+int rebuild_activelist(inttime_t Ti_Current, int NumCurrentTiStep)
```
rainwoodman: NumCurrentTiStep is only used in the print. Is there a way to reformat the print to get rid of it, or moving it to All (since it is globally synced anyways?)

libgadget/run.c
```diff
@@ -249,6 +253,8 @@ void run(void)
         /* Update velocity to the new step, with the newly computed step size */
         apply_half_kick();
 
+        free_activelist();
```
rainwoodman: Why is activelist no longer needed after this?

Perhaps shall pass-in ActiveParticles as an argument to these functions above to signify if they need it.

libgadget/slotsmanager.c
```diff
@@ -379,8 +379,15 @@ slots_gc_sorted()
         PartManager->NumPart--;
     }
 
-    /*Set up ReverseLink*/
-    slots_gc_mark();
+    int gc_slots = 0;
+    for(ptype = 0; ptype < 6; ptype ++)
+        if(SLOTS_ENABLED(ptype))
+            gc_slots = 1;
```
rainwoodman: Can we move this inside slots_gc_mark?

libgadget/slotsmanager.c
```diff
@@ -379,8 +379,15 @@ slots_gc_sorted()
...
+
+    if(gc_slots) {
+        /*Set up ReverseLink*/
+        slots_gc_mark();
+    }
 
     for(ptype = 0; ptype < 6; ptype++) {
         if(!SLOTS_ENABLED(ptype))
```
rainwoodman: Also feels like the sorting and reduce can be moved into gc_slots_collect.

libgadget/treewalk.c
```diff
@@ -281,47 +275,74 @@ cmpint(const void *a, const void *b)
...
+    if(!tw->haswork && !may_have_garbage)
+    {
+        tw->WorkSetSize = size;
+        tw->WorkSet = active_set;
+        tw->work_set_allocated = 0;
+        return;
+    }
+
+    tw->WorkSet = mymalloc("ActiveQueue", size * sizeof(int));
+    tw->work_set_allocated = 1;
```
rainwoodman: Why not checking for NULL instead of a new variable?

rainwoodman: I see. It can be a stolen array.

rainwoodman: Use 'workset_stolen_from_active_set' and reverse the logic?

libgadget/treewalk.c
```diff
@@ -281,47 +275,74 @@ cmpint(const void *a, const void *b)
...
+        /* Skip the garbage particles */
+        if(P[p_i].IsGarbage) continue;
+
+        if(tw->haswork && !tw->haswork(p_i, tw))
+            continue;
+        thrqueue[size * tid + nqthr[tid]] = p_i;
+        nqthr[tid]++;
+    }
+    /*Merge step for the queue.*/
+    nqueue = 0;
```
rainwoodman: This showed up twice. Would a function like this be useful?

```
gadget_compact_arrays (void * dest, void * srcs[], size_t sizes[], size_t elsize, int narrays)
```

libgadget/treewalk.c
```diff
@@ -281,47 +275,74 @@ cmpint(const void *a, const void *b)
...
 #ifdef DEBUG
-        /* check the uniqueness of ActiveParticle list. */
-        /* FIXME: the sort may affect performance of treewalk */
-        qsort_openmp(queue, k, sizeof(int), cmpint);
-        for(i = 0; i < k - 1; i ++) {
-            if(queue[i] == queue[i+1]) {
-                endrun(8829, "A few particles are twicely active.");
-            }
+    /* check the uniqueness of the active_set list. This is very slow. */
+    qsort_openmp(queue, nqueue, sizeof(int), cmpint);
```
rainwoodman: We can replace this with a hash table of length nqueue and type char. Then it is blazingly fast.

libgadget/treewalk.c
```diff
@@ -342,10 +363,16 @@ static int ev_primary(TreeWalk * tw)
...
 
     ev_alloc_threadlocals();
-#pragma omp parallel
+
+    int nint = tw->Ninteractions;
+    int nnodes = tw->Nnodesinlist;
+#pragma omp parallel reduction(+: nint) reduction(+: nnodes)
     {
-        real_ev(tw);
+        real_ev(tw, &nint, &nnodes);
```
rainwoodman: The reason real_ev is moved into a function was to help GDB identifying the local variables in a omp block.  I think newer versions of gdb doesn't have this issue. So free free to merge real_ev back here.

libgadget/treewalk.c
```diff
@@ -379,6 +406,10 @@ static int ev_primary(TreeWalk * tw)
         tw->Nexport --;
     }
 
+    if(tw->BufferFullFlag) {
+        message(1, "Tree export buffer full with %d particles. Increase BufferSize if possible.\n", tw->Nexport);
```
rainwoodman: "This is not fatal but makes treewalk slower."

libgadget/init.c
sbird: It should, I was just being defensive for future double-frees. I can make that line force_tree_free() if you like. 

The tree needs to be rebuilt after the first domain exchange in run.c anyway (although the first domain will exchange zero particles, so this isn't strictly necessary). Limiting the tree lifetime to this function seemed to minimize special-casing.

libgadget/treewalk.c
sbird: Sure - Do we have a hash table structure around? This is within a DEBUG ifdef, so the speed is not vital.

libgadget/timestep.c
sbird: We could; it isn't really used anywhere except printing the logs though. I think I prefer to keep it local, it isn't too much hassle to pass it into the two functions that use it.

libgadget/slotsmanager.c
sbird: Done

libgadget/slotsmanager.c
sbird: No, because it isn't done in slots_gc_slots (the non-sorting gc). In that case the slots array is already sorted in the same order as the particles.

libgadget/treewalk.c
sbird: Done

libgadget/treewalk.c
sbird: Done.

sbird: Ideally we would increase the BufferSize dynamically at this point. But as I never hit this condition in practice I am unsure how exactly we would resize it.

libgadget/treewalk.c
sbird: Both real_ev and ev_primary are quite long functions. I'm happy to keep them separate.

libgadget/treewalk.c
sbird: Done - although I made it for integers only because I was scared of casting int ** to void ** 
It doesn't save any lines of code but I think it is clearer.

sbird:
Shall I merge (after a rebase to get the new commits from master)?

rainwoodman:
Sure!

Issue #224, opened 2018-06-20
Crash with optimisation -march=native when compiling with icc


Chris-Pedersen:
When compiling with icc and the OPTIMIZE flag -march=native, simulations crash early on with:
```
[ 000011.83 ] Created 64 segments in 1 rounds. Max leaf cost: 0.132111
[ 000011.83 ] Largest load: work=1.09922 particle=1.09922
[ 000011.85 ] Using 7302798848 bytes for exchange.
[ 000011.88 ] iter=0 exchange of 0000032911847 particles
[ 000011.96 ] GC : Reducing Particle slots from 33554432 to 642585
[ 000011.97 ] GC: Reducing number of slots for 0 from 16777216 to 318389
[ 000012.14 ] Task 0: Exchange: P[123562].ID = 16850946 != SLOT ID = 20987906
[ 000012.14 ] Task 32: Exchange: P[0].ID = 16794018 != SLOT ID = 16859522
...
[ 000012.14 ] Task 31: Exchange: P[0].ID = 16785762 != SLOT ID = 29458690
[ 000012.14 ] Task 63: Exchange: P[0].ID = 16843009 != SLOT ID = 16843138
Thread 2 (Thread 0x2b0bf3ab1700 (LWP 432509)):
#0  0x00002b0bda200e9d in poll () from /lib64/libc.so.6
#1  0x00002b0bef10f300 in ips_ptl_pollintr () from /lib64/libpsm2.so.2
#2  0x00002b0bd993edd5 in start_thread () from /lib64/libpthread.so.0
#3  0x00002b0bda20bb3d in clone () from /lib64/libc.so.6
Thread 1 (Thread 0x2b0bd78fab40 (LWP 432074)):
#0  0x00002b0bd99457ad in read () from /lib64/libpthread.so.0
#1  0x000000000044dcfa in show_backtrace () at utils/endrun.c:77
#2  messagev (crash=<optimized out>, where=<optimized out>, fmt=<optimized out>, va=<optimized out>) at utils/endrun.c:198
#3  endrun (where=18, fmt=0x7fff48268b40 "X\213&H\377\177") at utils/endrun.c:216
#4  0x0000000000412e48 in domain_exchange_once (layoutfunc=0x12, plan=0x7fff48268b40) at exchange.c:272
#5  0x0000000000411370 in domain_exchange (layoutfunc=0x44b3e0 <domain_layoutfunc>) at exchange.c:102
[...]

PR #225, opened 2018-06-22
Fix for icc bug when using -march=native optimisation

Chris-Pedersen:
Fix for #224


libgadget/exchange.c
```diff
@@ -362,8 +362,8 @@ domain_build_plan(int (*layoutfunc)(int p), ExchangePlan * plan)
 
     int rank;
     for(rank = 1; rank < NTask; rank ++) {
-        plan->toGoOffset[rank] = plan->toGoSum;
-        plan->toGetOffset[rank] = plan->toGetSum;
+        memcpy(&plan->toGoOffset[rank], &plan->toGoSum, sizeof(plan->toGoSum));
```
rainwoodman: Could you add a comment here to say that direct assignment breaks compilers like icc?

PR #226, opened 2018-07-02
Warnings

sbird:
This branch fixes some warnings in the gadget source tree. Nothing major. The first few commits fix potential buffer over-runs caught by gcc 8 warnings. The next few make MP-Gadget compile cleanly with -Wsign-compare, which mostly involves making loop variables 64 bit instead of 32 bit. This would I guess be important if we ever ran with more than 2^31 halos. Finally I swapped some VLAs out for heap-allocated memory, which might stop our stack growing unreasonably in very large simulations. There are a few remaining alloca and VLA calls, but all of them are in libgadget-utils and have hard compile-time bounds. With these commits the code is clean with:

-Wall -Wextra -Wno-unused-parameter 

and everything except walltime.c is clean with -Wvla


sbird:
Yu, any comments? Once you sign off I'll do a test run in case I missed a variable somewhere before I merge.

libgadget/exchange.c
```diff
@@ -317,7 +317,11 @@ domain_build_plan(int (*layoutfunc)(int p), ExchangePlan * plan)
     int n;
     size_t package;
     int ptype;
-    ptrdiff_t nlimit = FreeBytes - NTask * 2 * sizeof(MPI_Request);
+    size_t nlimit = FreeBytes;
```
rainwoodman: What if FreeBytes went negative?

rainwoodman: Not it won't happen here.

libgadget/fof.c
```diff
@@ -801,8 +802,9 @@ static void fof_reduce_groups(
     /* count how many we have of each task */
     memset(Send_count, 0, sizeof(int) * NTask);
 
-    for(i = 0; i < nmemb; i++) {
-        struct BaseGroup * gi = (struct BaseGroup *) (((char*) groups) + i * elsize);
+    size_t n;
```
rainwoodman: It feels a bit strange to iterate with an unsigned variable.

libgadget/slotsmanager.c
```diff
@@ -532,7 +527,7 @@ slots_check_id_consistency()
     for(i = 0; i < PartManager->NumPart; i++) {
         if(!SLOTS_ENABLED(P[i].Type)) continue;
 
-        if(P[i].PI >= SlotsManager->info[P[i].Type].size) {
+        if(P[i].PI >= (unsigned int) SlotsManager->info[P[i].Type].size) {
```
rainwoodman: Why didn't we make .size unsigned?

libgadget/timestep.c
```diff
@@ -473,7 +473,7 @@ get_timestep_ti(const int p, const inttime_t dti_max)
     /*
     sqrt(2 * All.ErrTolIntAccuracy * All.cf.a * All.SofteningTable[P[p].Type] / ac) * All.cf.hubble,
     */
-    if(dti <= 1 || dti > TIMEBASE)
+    if(dti <= 1 || (unsigned int) dti > TIMEBASE)
```
rainwoodman: Cleaner to make TIMEBASE signed.

libgadget/utils/event.c
```diff
@@ -51,7 +51,7 @@ event_emit(EventSpec * eh, EIBase * event)
     return 0;
 }
 
-EventSpec EventSlotsFork = {"SlotsFork", 0};
-EventSpec EventSlotsAfterGC = {"SlotsAfterGC", 0};
+EventSpec EventSlotsFork = {"SlotsFork", 0, {{NULL}, {NULL}}};
```
rainwoodman: Is this necessary? I thought C compiler is supposed to zero pad the struct when the initializer is incomplete.

sbird: {"SlotsFork", 0, {{0}}} works. Would you prefer that?

rainwoodman: The issue I'd like to avoid is when the struct grows, the initializer here shall not start to give warnings that we have to fix down the road. So `{{0}}`, as ugly as it is it solves this.

libgadget/slotsmanager.c
rainwoodman: Maybe PI shall be signed? I think we do set it to (-1) in some occasions.

libgadget/fof.c
rainwoodman: I recall reading otherwise. This is an interesting read:

https://software.intel.com/en-us/forums/intel-c-compiler/topic/534769

C++ does use unsigned indices in the standard library, but scroll to the 'we were young' part. 

I think size_t is for specifying sizes, and ptrdiff_t is for offsets (indexing, which can be signed). Some even argue size_t shall be signed as well. But anyhow, I can deal with it if indices are unsigned unless necessary.

libgadget/slotsmanager.c
sbird: NumPart is signed too, so making PI signed would work. I'll try that.

libgadget/fof.c
sbird: The STL does it was indeed my reasoning - I think the important thing is that the loop index not be a smaller type than the limit variable. A cleaner fix might just be to make Ngroups etc signed.

sbird: Can we rely on Ngroups being < 2^31? I think we probably can.

sbird:
Ready to merge when travis goes green - tested (and a bug that only showed up with threads fixed)

PR #229, opened 2018-07-26
Exchange thread scaling

sbird:
A few changes to make the exchange faster and more thread friendly. After this all the expensive functions during exchange should be parallelized and, at the cost of some memory, some loops over all particles are avoided. The export flags are removed and replaced with an ExchangeList array (not too much of a memory cost because we no longer have an ActiveParticle list allocated during exchange).

On a KNL node with 16 threads per MPI this reduces the total exchange time by a factor of two and the non-garbage exchange time by a factor of three.

### DIFF
--- a/libgadget/exchange.c
+++ b/libgadget/exchange.c
@@ -30,7 +30,7 @@ typedef struct {
  * exchange particles according to layoutfunc.
  * layoutfunc gives the target task of particle p.
 */
-static int domain_exchange_once(int (*layoutfunc)(int p), ExchangePlan * plan);
+static int domain_exchange_once(int (*layoutfunc)(int p), ExchangePlan * plan, int do_gc);
 static int domain_build_plan(int (*layoutfunc)(int p), ExchangePlan * plan);
 
 /* This function builts the count/displ arrays from
@@ -52,7 +52,7 @@ _transpose_plan_entries(ExchangePlanEntry * entries, int * count, int ptype)
         }
     }
 }
-int domain_exchange(int (*layoutfunc)(int p)) {
+int domain_exchange(int (*layoutfunc)(int p), int do_gc) {
     int i;
     int64_t sumtogo;
     int failure = 0;
@@ -99,7 +99,7 @@ int domain_exchange(int (*layoutfunc)(int p)) {
 
         message(0, "iter=%d exchange of %013ld particles\n", iter, sumtogo);
 
-        failure = domain_exchange_once(layoutfunc, &plan);
+        failure = domain_exchange_once(layoutfunc, &plan, do_gc || (need_more > 0));
 
         if(failure)
             break;
@@ -133,7 +133,7 @@ shall_we_compact_slots(int * compact, ExchangePlan * plan)
     MPI_Allreduce(MPI_IN_PLACE, compact, 6, MPI_INT, MPI_LOR, MPI_COMM_WORLD);
 }
 
-static int domain_exchange_once(int (*layoutfunc)(int p), ExchangePlan * plan)
+static int domain_exchange_once(int (*layoutfunc)(int p), ExchangePlan * plan, int do_gc)
 {
     int i, target, ptype;
     struct particle_data *partBuf;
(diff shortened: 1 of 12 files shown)


sbird:
Tested that it gives bit-identical output to master.

sbird:
One extra thing that I didn't do was to try skipping gc when the number of exchanged particles is small. That is for the future.

libgadget/exchange.c
```diff
@@ -163,40 +201,45 @@ static int domain_exchange_once(int (*layoutfunc)(int p), ExchangePlan * plan)
...
     ta_free(toGoPtr);
     walltime_measure("/Domain/exchange/makebuf");
 
-    /*Find which slots to gc*/
-    int compact[6] = {0};
-    shall_we_compact_slots(compact, plan);
-    slots_gc(compact);
+    /* Do a gc if we were asked to, or if we need one
+     * to have enough space for the incoming material*/
+    int shall_we_gc = do_gc || (PartManager->NumPart + plan->toGetSum.base > PartManager->MaxPart);
```
rainwoodman: do_gc doesn't always do_gc. Perhaps we shall call it 'skip_gc'?

libgadget/exchange.c
```diff
@@ -17,21 +17,36 @@ typedef struct {
...
     ExchangePlanEntry * toGoOffset;
     ExchangePlanEntry * toGet;
     ExchangePlanEntry * toGetOffset;
     ExchangePlanEntry toGoSum;
     ExchangePlanEntry toGetSum;
+    /*List of particles to exchange*/
+    int * ExchangeList;
+    /*Total number of exchanged particles*/
+    int nexchange;
+    /*First and last particles in this exchange*/
```
rainwoodman: struct {
  int first;  /* relative to P[0] */
  int last;
} current_batch;

libgadget/exchange.c
```diff
@@ -250,7 +293,7 @@ static int domain_exchange_once(int (*layoutfunc)(int p), ExchangePlan * plan)
     int src;
     for(src = 0; src < NTask; src++) {
         /* unpack each source rank */
-        int newPI[6];
+        int newPI[6], i;
```
rainwoodman: one line per variable.

libgadget/exchange.c
```diff
@@ -52,7 +67,7 @@ _transpose_plan_entries(ExchangePlanEntry * entries, int * count, int ptype)
         }
     }
 }
-int domain_exchange(int (*layoutfunc)(int p)) {
+int domain_exchange(int (*layoutfunc)(int p), int do_gc) {
```
rainwoodman: We prefer running gc inside exchange because it reduces the maximum memory footprint.
`int do_gc` really is a `incremental_hint` -- a flag to notify the routine not to worry about reducing the peak memory footprint.

libgadget/exchange.c
```diff
@@ -63,54 +78,77 @@ int domain_exchange(int (*layoutfunc)(int p)) {
...
-    plan.toGo = (ExchangePlanEntry *) mymalloc2("toGo", sizeof(plan.toGo[0]) * NTask);
-    plan.toGoOffset = (ExchangePlanEntry *) mymalloc2("toGo", sizeof(plan.toGo[0]) * NTask);
-    plan.toGet = (ExchangePlanEntry *) mymalloc2("toGet", sizeof(plan.toGo[0]) * NTask);
-    plan.toGetOffset = (ExchangePlanEntry *) mymalloc2("toGet", sizeof(plan.toGo[0]) * NTask);
+    /*Build a list of particles that will be exchanged*/
+    plan.first = 0;
+    plan.last = 0;
+    plan.ExchangeList = mymalloc2("exchangelist", sizeof(int) * PartManager->NumPart * omp_get_max_threads());
+    memset(plan.ExchangeList, 0, sizeof(int) * PartManager->NumPart * omp_get_max_threads());
+    size_t *nexthr = ta_malloc("nexthr", size_t, omp_get_max_threads());
```
rainwoodman: feels like this can be made into a function in the same spirit as gadget_compact_thread_arrays.

libgadget/exchange.c
sbird: Right. At the moment do_gc is true unless we know for sure that a gc will be run just after the exchange by the full domain decomposition. So all it does right now is avoid doing a gc twice.

But we can later use it as a hint to skip reducing the the peak memory when exchange is small.

libgadget/exchange.c
sbird: Would you be ok with just a comment? Creating a struct for two ints feels like overkill.

libgadget/exchange.c
sbird: void gadget_setup_thread_arrays(int * dest, int * srcs[], size_t sizes[], size_t total_size, int narrays)
{
    memset(dest, 0, sizeof(int) * total_size * narrays);
    srcs[0] = dest;
    for(i=0; i < narrays; i++) {
        srcs[i] = dest + i * total_size;
        sizes[i] = 0;
    }
}

Like that?

libgadget/exchange.c
rainwoodman: Yes. Overkill.

libgadget/exchange.c
rainwoodman: Yes. But it is surprising when `do_gc` is True, gc is not always done.

libgadget/exchange.c
rainwoodman: Yes. -- Don't we also use this elsewhere?

libgadget/exchange.c
sbird: I think you are misreading: shall_we_gc is an or, not an and. When do_gc is True, gc is always done. When do_gc is false, gc may or may not be done.

rainwoodman: Ah. Indeed! I misread.

sbird:
Thanks for the review! I will merge when travis goes green.

PR #230, opened 2018-07-30
Remove Barriers

sbird:
While working on something else I noticed that we were spending a lot of time syncing at MPI_Barrier calls after message(0, ). So I removed the MPI_Barrier - I think it is mostly around so that we have well-measured timestamps in the log, but walltime_measure does that for us better. I get a speed-up of about 1.5% on a small 2 node, 48 MPI run, which is not large but not that small either for the size of the change.
I also removed some MPI_Barrier calls in debug functions, and stray calls in a couple of other places.

There is the possibility that removing what is effectively a sync after every major function will reveal some latent race conditions somewhere, and initially I thought this was happening, but it went away after a recompile (or maybe required a specific timing condition only found on some specific cluster nodes).

In case it resurfaces, the symptom is a crash in find_timesteps because GravPM[1] ~ 1e180 for some SPH particles.

I then noticed an interesting bug - the peak memory usage was actually reporting the peak *free* memory! I fixed this and then made small changes to petapm and treewalk to reduce the peak a bit.

Feel free to object to this - I might have missed good reasons for these Barriers.

### DIFF
--- a/genic/main.c
+++ b/genic/main.c
@@ -155,7 +155,6 @@ int main(int argc, char **argv)
   message(0, "IC's generated.\n");
   message(0, "Initial scale factor = %g\n", All.TimeIC);
 
-  MPI_Barrier(MPI_COMM_WORLD);
   print_spec();
 
   MPI_Finalize();		/* clean up & finalize MPI */
--- a/libgadget/forcetree.c
+++ b/libgadget/forcetree.c
@@ -156,6 +156,8 @@ int force_tree_build(int npart)
 
     force_treeupdate_pseudos(PartManager->MaxPart, tb);
 
+    myrealloc(Nodes_base, (Numnodestree +1) * sizeof(struct NODE));
+
     event_listen(&EventSlotsFork, force_tree_eh_slots_fork, NULL);
     return Numnodestree;
 }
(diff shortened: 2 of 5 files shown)


sbird:
The long test showed that there were surprisingly large shifts in the particle positions from this. Please do not merge yet.

sbird:
It emerges that some machines (probably different openmpi versions) respond to a large desync by becoming really slow, and some respond by giving occasionally wildly incorrect answers (!). So I'm dropping these barrier removals.

I'm still a little alarmed by this because it means debug messages can have potentially large effects in behaviour - maybe we actually need more MPI_Barriers, particularly around the tree - but it seems less effort to use the barrier in message(0, than to figure out exactly where we need to put explicit MPI_Barrier() calls to avoid potential crazy bugs.

rainwoodman:
I suspect adding a barrier at the top and end of each major function, and at each  major iteration step would be good enough. 

Here are a few examples where barriers affect the result:

http://www.mcs.anl.gov/~thakur/papers/irrelevant-barriers.pdf

Looks like the only dangerous place is where we manually send / recv. I only find these calls in system.c, which may need a barrier before alltoallv_sparse, and
domain.c, which may need a barrier, near starting of domain_nonrecursively_combine_topTree. 

Could you give this a go?

sbird:
I'll try that, but I might not get to it soon. Let me merge the memory fix first though... I think we may also need explicit barriers before mpsort_mpi. My guess is this is a fast path being hit.

sbird:
mpsort_mpi has a Barrier near the start anyway.

PR #232, opened 2018-09-12
Script to convert bigfile to Gadget format 3

sbird:
This adds a python script to convert a bigfile snapshot to a gadget format 3 HDF5 snapshot. The use-case is people with existing analysis pipelines, so they can try out MP-Gadget without needing to switch snapshot format immediately.

At the moment it is pretty bare-bones: it should do Gadget format 2 as well, and support the reverse: conversion of Gadget formats 2,3 to bigfile.

Tested that it works and completes conversion: the results look roughly correct. I have also sent this to someone with an existing pipeline and they have not complained.


rainwoodman:
What about adding a round trip test in the travis script?

sbird:
Good plan, but I would have to implement reverse conversion first

rainwoodman:
Ah. I misread the 'should do' part. `convert-to-gadget-format3.py` then? Even if not a round trip, a test in travis can at least assert the script runs.

sbird:
Sure, that sounds a good idea. How about I make it convert-to-gadget-format.py , to give myself more motivation to finish writing the format 2 support?

I know Jia's weak lensing pipeline (for example) takes format 2, and we can cannibalize some of the code in pynbody (which I cannibalised from GadgetReader).

sbird:
Any more comments or shall I merge it?

rainwoodman:
go ahead!

sbird:
thanks!

PR #233, opened 2018-09-28
Fix multi-iteration exchanges

sbird:
When an exchange has multipe iterations, it performs a garbage
collection in between them. This moves the particles, and so the
ExchangeList no longer points in the right place. Fix this by
recomputing ExchangeList inside the iteration loop.

### DIFF
--- a/libgadget/exchange.c
+++ b/libgadget/exchange.c
@@ -47,6 +47,7 @@ typedef struct {
 static int domain_exchange_once(int (*layoutfunc)(int p), ExchangePlan * plan, int do_gc);
 static void domain_build_plan(int (*layoutfunc)(int p), ExchangePlan * plan);
 static int domain_find_iter_space(ExchangePlan * plan);
+static void domain_build_exchange_list(int (*layoutfunc)(int p), ExchangePlan * plan);
 
 /* This function builts the count/displ arrays from
  * the rows stored in the entry struct of the plan.
@@ -67,8 +68,9 @@ _transpose_plan_entries(ExchangePlanEntry * entries, int * count, int ptype)
         }
     }
 }
+
+/*Plan and execute a domain exchange, also performing a garbage collection if requested*/
 int domain_exchange(int (*layoutfunc)(int p), int do_gc) {
-    int i;
     int64_t sumtogo;
     int failure = 0;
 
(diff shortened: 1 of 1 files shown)


rainwoodman:
The first time step would always be a multi-iteration exchange. Did you observe a crash due to this bug?

What if we move GC outside of the iterations?

sbird:
The first timestep isn't always a multi-iteration exchange -  because Tree and ActiveParticle aren't allocated then anymore, it generally isn't. I did see a crash due to this.

We can't move gc outside the iterations because we need it to get the memory for the next iteration.

libgadget/exchange.c
```diff
@@ -82,34 +78,10 @@ int domain_exchange(int (*layoutfunc)(int p), int do_gc) {
...
-            nexthr[tid]++;
-        }
-    }
-    /*Merge step for the queue.*/
-    plan.nexchange = gadget_compact_thread_arrays(plan.ExchangeList, threx, nexthr, omp_get_max_threads());
-    ta_free(threx);
-    ta_free(nexthr);
-    /*Shrink memory*/
-    plan.ExchangeList = myrealloc(plan.ExchangeList, sizeof(int) * plan.nexchange);
+    plan.nexchange = PartManager->NumPart;
```
rainwoodman: just any large number?

libgadget/exchange.c
```diff
@@ -120,13 +92,46 @@ int domain_exchange(int (*layoutfunc)(int p), int do_gc) {
...
     plan.toGetOffset = (ExchangePlanEntry *) mymalloc2("toGet", sizeof(plan.toGo[0]) * NTask);
 
-
     walltime_measure("/Domain/exchange/init");
 
     int iter = 0;
 
     while(MPIU_Any(plan.last < plan.nexchange, MPI_COMM_WORLD))
     {
+        plan.ExchangeList = mymalloc2("exchangelist", sizeof(int) * plan.nexchange * omp_get_max_threads());
```
rainwoodman: Let's move this chunk of code into a function, e.g. merging the logic into domain_find_iter_space;

Further, ExchangeList, nexchange and last are different from other elements of the plan. We can lift ExchangeList and nexchange (whether a struct or not I don't know yet), which becomes the input argument of domain_build_plan and domain_find_iter_space. `domain_find_iter_space` could have truncated the ExchangeList and flagged that the loop shall not terminate.

rainwoodman:
I see our hands are tied.

I think this patch signifies that ExchangeList and Plan are two different objects. Plan depends on ExchangeList. Only if the ExchangeList is incomplete or Plan cannot be fully executed, the loop can continue.

libgadget/exchange.c
sbird: I want there to be enough memory in the first iteration of ExchangeList, which gets allocated with nexchange. 

plan.ExchangeList = mymalloc2("exchangelist", sizeof(int) * plan.nexchange * omp_get_max_threads());

libgadget/exchange.c
sbird: I'll move it into domain_find_iter_space, but we don't get anything from lifting ExchangeList, nexchange and last into a separate struct because ExchangeList is also used in domain_exchange_once to do the actual copying. So although the constructors are split into dfferent functions, the lifetime of the data is the same.

sbird:
> I recall that in my runs it were (similar load to MB and BT). Perhaps the memory was tighter. I suspect this can be reproduced at the start up by using a rather small PartAllocFactor (triggers gc) and rather tight memory limit (multiple exchanges)

This definitely used to be the case (and indeed can be reproduced as you suggest) - it stopped being so when you rewrote domain to use less memory and I changed the memory allocation patterns to allocate ActiveParticle later.  That means that the exchange is often no longer the peak memory usage, which is instead during treewalk.

PR #234, opened 2018-10-01
Improve threading scalability

sbird:
I ran some benchmarks with 48 threads. This addresses places where it did not scale as well as distributed memory, and does a few cleanups. 

- The GC was taking a lot of time. I first tried making it parallel and failed (because it is limited by memory bandwidth, the parallel version was not faster). I then changed the code to not do it on a short timestep (unless there is a big exchange). There is no perceptible performance change due to having garbage left floating around.
- I changed SFR and BH to just convert particles where possible, instead of creating garbage.
- I changed the treewalk buffer to just use available memory, like exchange does. This was motivated by threaded runs generally needing much larger buffers. BufferSize is no longer a parameter, but I added something to leave extra space for imported particles, which we can't strictly bound (although I have never observed the edge case the parameter guards against).
- SFR had an omp critical which serialised the whole loop! Removed it, but performance is still bad. Perhaps the atomic capture in slots_fork()?

With this the threading scalability is very close to linear, as long as there is no star formation. The treewalk in particular is faster at high thread count than with equivalent processors on MPI ranks.

Tested with a Lyman alpha run forming a lot of stars.

Other things that don't thread perfectly:
- The density and hydro calculations are both a little slower at high thread count. Not much, but I guess there is some part of the loop that isn't threading very well.
- The qsort_openmp at the end of a full domain_decompose, because the final copy in the indirect sort is serial. Making the sort direct is so much slower it isn't worth it.
- The pfft r2c transform (nothing I can do about that).

### DIFF
--- a/libgadget/blackhole.c
+++ b/libgadget/blackhole.c
@@ -705,20 +705,23 @@ void blackhole_make_one(int index) {
     if(P[index].Type != 0)
         endrun(7772, "Only Gas turns into blackholes, what's wrong?");
 
-    int child = slots_fork(index, 5);
+    int child;
 
-    BHP(child).FormationTime = All.Time;
-    /*Ensure that mass is conserved*/
-    double BHmass = All.SeedBlackHoleMass;
+    /*If the particle mass is less than that needed for a black hole, convert.*/
     if(P[index].Mass <= All.SeedBlackHoleMass) {
-        slots_mark_garbage(index);
-        BHmass = P[index].Mass;
+        child = slots_convert(index, 5);
     }
-    P[child].Mass = BHmass;
-    P[index].Mass -= BHmass;
+    /*Otherwise create a new particle*/
+    else {
+        child = slots_fork(index, 5);
+        /*Ensure that mass is conserved*/
+        P[child].Mass = All.SeedBlackHoleMass;
+        P[index].Mass -= All.SeedBlackHoleMass;
+    }
     BHP(child).base.ID = P[child].ID;
-    BHP(child).Mass = BHmass;
+    BHP(child).Mass = P[child].Mass;
     BHP(child).Mdot = 0;
+    BHP(child).FormationTime = All.Time;
 
     /* It is important to initialize MinPotPos to the current position of
      * a BH to avoid drifting to unknown locations (0,0,0) immediately
--- a/libgadget/blackhole.c
+++ b/libgadget/blackhole.c
@@ -719,6 +719,7 @@ void blackhole_make_one(int index) {
         P[index].Mass -= All.SeedBlackHoleMass;
     }
     BHP(child).base.ID = P[child].ID;
+    /*This is min(P[index].Mass, SeedBlackHoleMass */
     BHP(child).Mass = P[child].Mass;
     BHP(child).Mdot = 0;
     BHP(child).FormationTime = All.Time;
--- a/libgadget/blackhole.c
+++ b/libgadget/blackhole.c
@@ -719,8 +719,9 @@ void blackhole_make_one(int index) {
         P[index].Mass -= All.SeedBlackHoleMass;
     }
     BHP(child).base.ID = P[child].ID;
-    /*This is min(P[index].Mass, SeedBlackHoleMass */
-    BHP(child).Mass = P[child].Mass;
+    /* The accretion mass should always be the seed black hole mass,
+     * irrespective of the gravitational mass of the particle.*/
+    BHP(child).Mass = All.SeedBlackHoleMass;
     BHP(child).Mdot = 0;
     BHP(child).FormationTime = All.Time;
 
--- a/libgadget/sfr_eff.c
+++ b/libgadget/sfr_eff.c
@@ -50,11 +50,11 @@ static double u_to_temp_fac; /* assuming very hot !*/
 /* these guys really shall be local to cooling_and_starformation, but
  * I am too lazy to pass them around to subroutines.
  */
-static int stars_converted;
-static int stars_spawned;
-static double sum_sm;
-static double sum_mass_stars;
-static double localsfr;
+static int * stars_converted;
+static int * stars_spawned;
+static double * sum_mass_stars;
+static double * sum_sm;
+static double * localsfr;
 
 static void cooling_relaxed(int i, double egyeff, double dtime, double trelax);
 
@@ -233,8 +233,11 @@ void cooling_and_starformation(void)
 
     walltime_measure("/Misc");
 
-    stars_spawned = stars_converted = 0;
-    sum_sm = sum_mass_stars = localsfr = 0;
+    stars_spawned = ta_malloc("stars_spawned", int, All.NumThreads);
+    stars_converted = ta_malloc("stars_converted", int, All.NumThreads);
+    sum_sm = ta_malloc("sum_sm", double, All.NumThreads);
+    sum_mass_stars = ta_malloc("sum_mass_stars", double, All.NumThreads);
+    localsfr = ta_malloc("localsfr", double, All.NumThreads);
 
     TreeWalk tw[1] = {0};
 
@@ -245,9 +248,19 @@ void cooling_and_starformation(void)
 
     treewalk_run(tw, ActiveParticle, NumActiveParticle);
 
+    int i;
+    for(i = 1; i < All.NumThreads; i++)
+    {
+        sum_mass_stars[0] += sum_mass_stars[i];
+        localsfr[0] += localsfr[i];
+        sum_sm[0] += sum_sm[i];
+        stars_spawned[0] += stars_spawned[i];
+        stars_converted[0] += stars_converted[i];
+    }
+
     int tot_spawned, tot_converted;
-    MPI_Allreduce(&stars_spawned, &tot_spawned, 1, MPI_INT, MPI_SUM, MPI_COMM_WORLD);
-    MPI_Allreduce(&stars_converted, &tot_converted, 1, MPI_INT, MPI_SUM, MPI_COMM_WORLD);
+    MPI_Allreduce(&stars_spawned[0], &tot_spawned, 1, MPI_INT, MPI_SUM, MPI_COMM_WORLD);
+    MPI_Allreduce(&stars_converted[0], &tot_converted, 1, MPI_INT, MPI_SUM, MPI_COMM_WORLD);
 
     if(tot_spawned > 0 || tot_converted > 0)
     {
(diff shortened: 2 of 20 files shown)


sbird:
There is also a omp critical in fof.c : that code explicitly locks particles, so I am not sure the critical is needed. Yu, can you confirm?

rainwoodman:
It was commented out for some time:

https://github.com/MP-Gadget/MP-Gadget/commit/c2528a75092533a9565b4ce599b7aea8c8d28e87

I may have reenabled it out of caution. It is very likely it works just fine without.

rainwoodman:
Probably want to label whether a function is reentry-able, e.g. those new gc functions.

I will look closely tomorrow.

sbird:
From the git history it looks actually like uncommenting it was just a merge error

sbird:
I will comment it again! And we shall see if anything breaks.

sbird:
Thanks for the fast review

libgenic/Makefile
```diff
@@ -26,7 +26,7 @@ all : libgenic.a
 .objs/test_%: tests/test_%.c .objs/%.o ../tests/stub.c ../tests/cmocka.c ../libgadget/libgadget-utils.a ../libgadget/libgadget.a
 	$(MPICC) $(CFLAGS) -I../tests/ $^ $(LIBS) -o $@
 
-build-tests: $(TESTBIN)
+build-tests: $(TESTBIN) libgenic.a
```
rainwoodman: I don't see any tests linking against libgenic.a. If they do then the dependency shall be added to `.objs/tests_%` If they don't then we are not even testing it?

libgadget/timestep.c
```diff
@@ -731,11 +731,10 @@ int rebuild_activelist(inttime_t Ti_Current, int NumCurrentTiStep)
     {
         const int bin = P[i].TimeBin;
         const int tid = omp_get_thread_num();
+        if(P[i].IsGarbage)
```
rainwoodman: Was this crashing the code due to the condition is no longer strictly met? It would be nice to describe how a garbage could still be seen here. (due to a reduced frequency of GC).

libgadget/sfr_eff.c
```diff
@@ -677,26 +677,27 @@ static int make_particle_wind(MyIDType ID, int i, double v, double vmean[3]) {
...
     double mass_of_star = find_star_mass(i);
+    int child;
     if(P[i].Type != 0)
         endrun(7772, "Only gas forms stars, what's wrong?");
 
-    /* if we get all mass or a fraction */
-    int child = slots_fork(i, 4);
-
+    /*Store the index of the SPH particle properties as overwritten in slots_convert*/
+    int oldslot = P[i].PI;
```
rainwoodman: I feel it is better to save the properties rather than a fragile index here. It avoids direct access to the SphP array.

libgadget/slotsmanager.c
```diff
@@ -22,39 +22,86 @@ slots_gc_base();
 static int
 slots_gc_slots(int * compact_slots);
 
-/*Initialise a new slot for the particle at index i.*/
+/* Initialise a new slot with type at index pi
+ * for the particle at index i.*/
```
rainwoodman: /* This will modify both P[i] and the slot at pi for the ptype. */

libgadget/slotsmanager.c
```diff
@@ -22,39 +22,86 @@ slots_gc_base();
...
-     *
-     * Its mass and ptype can be then adjusted. (watchout detached BH /SPH
-     * data!)
-     * PI will point to a new slot for this type.
-     * if the slots runs out, this will trigger a slots growth
-     * */
+    P[parent].Generation++;
+    uint64_t g = P[parent].Generation;
+    /* change the child ID according to the generation. */
+    P[parent].ID = (P[parent].ID & 0x00ffffffffffffffL) + (g << 56L);
```
rainwoodman: We may want to add a crash here if we run out of generations and start producing duplicated IDs. e.g. a FIXME note

libgadget/slotsmanager.c
```diff
@@ -22,39 +22,86 @@ slots_gc_base();
...
+        slots_connect_new_slot(parent, PI, ptype);
+    }
+    /*Type changed after slot updated*/
+    P[parent].Type = ptype;
+    return parent;
+}
 
+/* this will fork a zero mass particle at the given location of parent of the given type.
+ *
+ * Assumes the particle is protected by locks in threaded env.
```
rainwoodman: `the parent particle`?

libgadget/blackhole.c
```diff
@@ -705,20 +705,23 @@ void blackhole_make_one(int index) {
...
+        child = slots_fork(index, 5);
+        /*Ensure that mass is conserved*/
+        P[child].Mass = All.SeedBlackHoleMass;
+        P[index].Mass -= All.SeedBlackHoleMass;
     }
-    P[child].Mass = BHmass;
-    P[index].Mass -= BHmass;
     BHP(child).base.ID = P[child].ID;
-    BHP(child).Mass = BHmass;
+    BHP(child).Mass = P[child].Mass;
```
rainwoodman: It is important to set the Mass to SeedBlackHoleMass.

BHP().Mass does not agree with the gravity mass in P.

rainwoodman: line 712 would have set it to P[].Mass

libgadget/sfr_eff.c
```diff
@@ -233,8 +233,11 @@ void cooling_and_starformation(void)
 
     walltime_measure("/Misc");
 
-    stars_spawned = stars_converted = 0;
-    sum_sm = sum_mass_stars = localsfr = 0;
+    stars_spawned = ta_malloc("stars_spawned", int, All.NumThreads);
```
rainwoodman: Add a FIXME comment these arrays will not be manually declared when we switch to OpenMP 4.7

rainwoodman: which supports per element array reduction

rainwoodman:
OK I left some comments per commit. Looks clean to me overall.

libgenic/Makefile
sbird: It's not to link against libgenic.a: it's just to ensure that the build process rebuilds all the object files before running the tests. Despite the % rule it wasn't doing this properly for me.

libgadget/timestep.c
sbird: Exactly, comment added.

libgadget/slotsmanager.c
```diff
@@ -364,6 +364,28 @@ order_by_type_and_key(const void *a, const void *b)
     return 0;
 }
 
+/*Returns the number of non-Garbage particles in an array with garbage sorted to the end.
+ * Used to trim recently sorted arrays. If ptype < 0, P array is trimmed.*/
```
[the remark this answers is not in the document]
sbird: Yes, that comment doesn't make sense!

libgadget/slotsmanager.c
sbird: Done

libgadget/slotsmanager.c
sbird: For convert we don't need to (because only one is ever spawned). I'll add a note to fork.

libgadget/slotsmanager.c
sbird: Done

libgadget/blackhole.c
sbird: Added comment

libgadget/sfr_eff.c
sbird: There are several places where that happens - it would tidy the code quite a bit. Unfortunately not yet, gcc openmp 4.5 support is still pretty new. Does openmp provide a way to allocate the temporary arrays on the heap?

libgenic/Makefile
rainwoodman: Could it be because libgenic.a wasn't declared in the `%` rule?

libgadget/blackhole.c
sbird: Are you sure? That's not what it does in the master branch. In both branches the logic should be the same: It is min(P[parent].Mass, SeedBlackHoleMass), which happens to also be the same as P[child].Mass.

libgenic/Makefile
sbird: Probably - do you think that is a better fix? If I did that it would link libgenic.a, but that probably doesn't matter

rainwoodman: Move libgenic.a to the % rule above?

libgadget/blackhole.c
rainwoodman: I am sure. The master has a bug then. The older version was correct.

https://github.com/MP-Gadget/MP-Gadget/blame/f5a6feaade9a53269ce073aaaefee655993c0f51/blackhole.c

SeedBlackholeMass is the accretion model which has almost nothing to do with the gravity mass of the object. lower seed mass produces later Eddington-limited accretion / bright quasar phases. If we give the blackholes different seed masses it would mean we are by chance using different accretion models!

libgadget/sfr_eff.c
rainwoodman: You can alloc with malloc inside a parallel session before entering for loop. Then do a reduction in a critical section after the for loop, but still in the parallel session.

libgenic/Makefile
sbird: hmm, that doesn't actually work (I think I do not understand make):

[spb43]codes/MP-Gadget/libgenic (gcthread)% make clean
rm -rf .objs libgenic.a
[spb43]codes/MP-Gadget/libgenic (gcthread)% make test
make: *** No rule to make target '.objs/test_power', needed by 'build-tests'.  Stop.
[spb43]codes/MP-Gadget/libgenic (gcthread)%

libgadget/sfr_eff.c
rainwoodman: I don't think that would be much cleaner than now. We shall catch these cases as they are touched. So that OpenMP is ready, all of the possible conversions are already marked.

libgenic/Makefile
sbird: But now the old fix doesn't work either...

libgadget/blackhole.c
sbird: Ah, indeed. This was introduced by  b5d1d95f775b7e7d9be248328d0da565f94d5393
which limited the BlackHole to only being as massive as the seed particle, but it should apply only to the gravitational mass, not the accretion mass.

libgadget/slotsmanager.c
rainwoodman: If for some reason we convert a particle many times between 0 and 4, it will still overflow in convert. the current sfr module doesn't trigger this I agree.

sbird: I added an error case to convert - no reason not to really.

PR #235, opened 2018-10-06
Warningsclang7

sbird:
Fix some warnings revealed by the new clang v7. Most are unimportant. 

There is one valid bug: the UV fluctuation model was using an uninitialized variable to allocate memory to copy into. The fix for this involves all the cooling data memory being allocated with mymalloc instead of malloc. I also turned on metal cooling in the travis tests to catch future bugs in that code.

### DIFF
--- a/libgadget/cooling.c
+++ b/libgadget/cooling.c
@@ -1055,7 +1055,7 @@ static void InitUVF(void) {
     int size;
     UVF.Table = h5readdouble(All.UVFluctuationFile, "Zreion_Table", &size);
     if(UVF.Table[0] < 0.01 || UVF.Table[0] > 100.0) {
-        endrun(123, "UV Flucutaiton doesn't seem right\n");
+        endrun(123, "UV Fluctuation out of range: %g\n", UVF.Table[0]);
     }
 }
 
--- a/libgadget/cooling.c
+++ b/libgadget/cooling.c
@@ -76,7 +76,6 @@ struct {
 
 static void find_abundances_and_rates(double logT, double nHcgs, struct UVBG * uvbg, struct abundance * y, struct rates * r);
 static double solve_equilibrium_temp(double u, double nHcgs, struct UVBG * uvbg, struct abundance * y);
-static double * h5readdouble(char * filename, char * dataset, int * Nread);
 
 double PrimordialCoolingRate(double logT, double nHcgs, struct UVBG * uvbg, double *nelec);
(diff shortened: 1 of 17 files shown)


libgadget/cooling.c
```diff
@@ -790,40 +789,6 @@ void MakeCoolingTable(void)
...
-        if(0 != big_file_open_block(bf, bb, dataset)) {
-            endrun(-1, "Cannot open %s %s\n", filename, dataset);
-        }
-
-        N = bb->size;
-        MPI_Bcast(&N, 1, MPI_INT, 0, MPI_COMM_WORLD);
-
-        BigArray array[1];
-
-        big_block_read_simple(bb, 0, N, array, "f8");
```
rainwoodman: I thought read_simple would have allocated the buffer automaticially.

rainwoodman: https://github.com/rainwoodman/bigfile/blob/master/src/bigfile.c#L774

sbird: It does - but it allocates it with malloc, not mymalloc. All I'm doing here is avoiding the copy from the malloc heap to the mymalloc heap.

sbird: Basically I have inlined read_simple and changed malloc to mymalloc

rainwoodman: Yes the copy was less than ideal -- but what was buggy?

libgadget/cooling.c
```diff
@@ -1034,31 +1042,21 @@ static void InitUVF(void) {
         interp_init_dim(&UVF.Finterp, 0, UVF.Zbins[0], UVF.Zbins[UVF.N_Zbins - 1]);
     }
 
-    UVF.Nside = size;
-
-    int Nside = UVF.Nside;
-    /* This is kinda big, so we move it to mymalloc (leaving more free space for
-     * system /MPI */
-    UVF.Table = mymalloc("Zreion", (sizeof(double) * Nside) * (Nside * Nside));
```
sbird: Size is not initialized here! And is copied into Nside. We could also fix this in fewer changed lines by just moving the mymalloc after h5readdouble

libgadget/cooling.c
sbird: See below - lines 1037 - 1044 used an uninitialized variable to allocate memory.

libgadget/cooling.c
rainwoodman: I see. Basically the entire Zerion table can be corrupt depending on the random number stored in size.
This explained why I was seeing ridiculously large memory usage from Zerion in BlueTides.

PR #236, opened 2018-10-19
Optimizations

sbird:
This series corrects three pessimizations I introduced earlier: 

9ac4abe removes a needless memset, which was taking a lot of time (20% of runtime!) since I made the buffer in question use all available memory.
9e5d93f avoids cooling dense particles that will become stars when using the quick lyman alpha star formation. This can be expensive since DoCooling is slow for dense particles.
1864109 replaces the function calls to sph_VelPred scattered around the hydro code with a variable in SPHP. I made it a function to save memory but it was very expensive! This saves about 20% of total runtime, and almost 50% of the SPH time!

While I was going I removed a few atomic variables, fixed a race condition that would never matter in practice and fixed a fixme by making the factor to expand slot memory configurable.

### DIFF
--- a/gadget/params.c
+++ b/gadget/params.c
@@ -206,8 +206,9 @@ create_gadget_parameter_set()
     param_declare_double(ps, "GravitySofteningGas", OPTIONAL, 1./30., "Softening for collisional particles (Gas); units of mean separation of DM; 0 to use Hsml of last step. ");
 
     param_declare_int(ps, "ImportBufferBoost", OPTIONAL, 6, "Memory factor to allow for there being more particles imported during treewlk than exported. Increase this if code crashes during treewalk with out of memory.");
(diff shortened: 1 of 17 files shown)


sbird:
Note the runtime numbers are at 48 threads; probably less emphatic at lower thread counts

sbird:
@rainwoodman The test failure is nbodykit saying:
   File "/home/travis/miniconda/envs/test/lib/python3.6/site-packages/nbodykit/source/catalog/species.py", line 201, in to_mesh
    raise RuntimeError("use resampler instead")
RuntimeError: use resampler instead

Do you know how to fix that?

rainwoodman:
Yes. Got a fix in #237. 'window' may stand for the sky mask; `resampler` has to such ambiguity.

sbird:
Ok, I'll address your review comments and rebase. Thanks for the quick reply.

PR #238, opened 2018-10-22
Sfrcleanup

sbird:
This is the first part of a series on removing thread contention in the star formation code. I split this one out to make the whole thing easier to review - this part is just cleanups. The main nice thing about it is that it removes the global thread-local variables in sfr_eff.c, instead just using locals with omp reduction clauses. Hooray! I also split winds into a separate file.

One thing I did was to make -DSFR just guard the wind model, so I could unify the routines for cooling and star formation and just cooling. This should prevent us from breaking compilation of the !SFR case. I also considered removing -DSFR entirely - the main runtime penalty is an extra double in SPHP which isn't really a big deal. What do you think?


sbird:
Tested that it gives identical results for the hydro example

rainwoodman:
Yes on removing the SFR conditional compilation flag.

libgadget/sfr_eff.c
```diff
@@ -23,373 +23,118 @@
...
-    tw->postprocess = (TreeWalkProcessFunction) cool_postprocess;
-
-    treewalk_run(tw, ActiveParticle, NumActiveParticle);
-
-    walltime_measure("/Cooling/StarFormation");
-}
-
+#ifdef SFR
+    /* Now apply the wind model: has to use the new NumActiveParticle.*/
+    winds_and_feedback(ActiveParticle, NumActiveParticle);
```
rainwoodman: But NumActiveParticle is also a global variable?

libgadget/sfr_eff.c
```diff
@@ -447,298 +192,68 @@ cooling_direct(int i) {
...
-        if(P[target].IsNewParticle && !P[target].DensityIterationDone) {
-             return 1;
-        }
-    }
-    return 0;
-}
-
-static int
-sfr_wind_feedback_haswork(int target, TreeWalk * tw)
+static int make_particle_star(int i, double mass_of_star)
```
rainwoodman: Please describe the meaning of retflag.

sbird: ok, will do. If it helps retflag disappears in the later (feature) commits

libgadget/sfr_eff.c
sbird: Comment is badly worded: I mean that NumActiveParticle changes within the loop when new particles are spawned, so one cannot use nactive.

sbird:
Ok! I'll take out -DSFR

PR #239, opened 2018-10-24
Improvements to star formation

sbird:
This is the second part of the pull request for improving thread scalability of the star formation code. This is the "feature" part. The original purpose was to avoid contention on the slot atomic in quick lyman alpha simulations. To do this I split slots_fork into two sections, one that splits the base particle and one that fills in a pre-allocated stellar slot, locklessly. There is still locking in the effective equation of state star formation, but this is much less severe because fewer stars are formed.

Now that the slot filling in is split into a separate loop, I can make sure enough memory is allocated to receive the new slots. Because I'm still not at the top of the heap doing this is quite involved and extremely messy. Suggestions as to how to do it better would be appreciated.

Finally, I realised I could avoid a lot of work in the winds model because I now have a list of the new stars, which is the winds active set. (so I don't need a haswork). This speeds up even the unthreaded, effective equation of state, star formation significantly

Tested identical results with 1 thread on both hydro and lya examples. Tested similar results with multiple threads as well.


rainwoodman:
I will take a look at this (including travis tomorrow). A quick way to check is to use a passed build and rerun it, see if the travis test fails? Then it is something external, e.g. nbodykit, or even numpy. Though I don't think anything changed in this week.

sbird:
Found it - a merge error had put an extra ifdef SFR in sfr_eff.c I had -DSFR left over in my makefile and so it was passing for me anyway. I'll fix it.

libgadget/partmanager.h
```diff
@@ -29,13 +29,13 @@ struct particle_data
         unsigned int IsGarbage            :1; /* True for a garbage particle. readonly: Use slots_mark_garbage to mark this.*/
         unsigned int Evaluated            :1; /* True if already query already ran in treewalk */
         unsigned int DensityIterationDone :1; /* True if the density-like iterations already finished; */
-        unsigned int IsNewParticle        :1; /* True if the particle is created this step; used in SFR */
```
rainwoodman: Changing this to something like spare_8 in place may be an easier changeset

libgadget/sfr_eff.c
```diff
@@ -25,40 +25,61 @@
...
+
+    /*Need to capture this so that when NumActiveParticle increases during the loop
+     * we don't add extra loop iterations on particles with invalid slots.*/
+    const int nactive = NumActiveParticle;
+
+    if(All.StarformationOn) {
+        NewStars = mymalloc("NewStars", nactive * sizeof(int) * All.NumThreads);
+        gadget_setup_thread_arrays(NewStars, thrqueuesfr, nqthrsfr, nactive, All.NumThreads);
+        NewParents = mymalloc2("NewParents", nactive * sizeof(int) * All.NumThreads);
+        gadget_setup_thread_arrays(NewParents, thrqueueparent, nqthrsfr, nactive, All.NumThreads);
```
rainwoodman: Could be easier if 
1. gadget_setup_thread_arrays allocates thrqueueparent with malloc and returns the per thread array directly. Then the frees later can be regular free(no ordering requirements), and you can hide a few ta_mallocs.

2. gadget_setup_thread_arrays also has almost enough information to do the mymalloc stuff, but it appears you need to use two different allocators because they have to be freed in different orders?

libgadget/partmanager.h
sbird: Since there are now 8 spares, I thought of just shrinking struct particle_data by a word, what do you think?

libgadget/sfr_eff.c
rainwoodman: I see the memory usage can be NumPart * NThreads, which can be huge. There are two ways of dealing with this:
- If we use a static scheduling policy, then the total size of NewStars can be nactive, rather than each has nactive. But the thread loads can be very imbalanced. We can try to schedule by interval rather than by chunk (probably need to manually write a for loop inside omp parallel.

- Use a single queue. What if you use an atomic capture to advance a single queue pointer? Would it be a lot slower than having one queue per thread?

libgadget/sfr_eff.c
```diff
@@ -78,28 +99,75 @@ void cooling_and_starformation(void)
...
     }
 
+    /*Merge step for the queue.*/
+    if(NewStars) {
+        NumNewStar = gadget_compact_thread_arrays(NewStars, thrqueuesfr, nqthrsfr, All.NumThreads);
+        int NumNewParent = gadget_compact_thread_arrays(NewParents, thrqueueparent, nqthrsfr, All.NumThreads);
+        if(NumNewStar != NumNewParent)
+            endrun(3,"%d new stars, but %d new parents!\n",NumNewStar, NumNewParent);
+        /*Shrink star memory as we keep it for the wind model*/
+        NewStars = myrealloc(NewStars, sizeof(int) * NumNewStar);
```
rainwoodman: If we limit the queue length to NumPart, then there is no need to shrink it.

libgadget/sfr_eff.c
```diff
@@ -359,24 +489,29 @@ starformation(int i, double *localsfr, double * sum_sm, double * sum_mass_stars)
...
 
-    if(P[i].Type == 0)	{
-        /* to protect using a particle that has been turned into a star */
+    /* A particle that will be turned into a star doesn't want the metals or to be made a wind.*/
+    if(newstar != i)	{
         SPHP(i).Metallicity += (1 - w) * METAL_YIELD * (1 - exp(-p));
 
         if(All.WindOn && HAS(All.WindModel, WIND_SUBGRID)) {
             /* Here comes the Springel Hernquist 03 wind model */
+            /*Notice that this is the mass of the gas particle after forking a star, 1/2 what it was before.*/
```
rainwoodman: Not necessarily half if GENERATIONS is not 2. It can be 1 / GENERATIONS

libgadget/sfr_eff.c
```diff
@@ -359,24 +489,29 @@ starformation(int i, double *localsfr, double * sum_sm, double * sum_mass_stars)
...
+        /* If we get a fraction of the mass we need to create
+         * a new particle for the star and remove mass from i.*/
+        if(P[i].Mass >= 1.1 * mass_of_star)
+            newstar = slots_split_particle(i, mass_of_star);
     }
 
-    if(P[i].Type == 0)	{
-        /* to protect using a particle that has been turned into a star */
+    /* A particle that will be turned into a star doesn't want the metals or to be made a wind.*/
+    if(newstar != i)	{
```
rainwoodman: This line updates the metalicity of the SPH but does not update the metalicity if it has been turned into a star. I think with newstar != i, you are checking for 1. split happened? 2. no sfr happened? Probably cleaner if you (can) explicitly check for both.

sbird: This was designed to match the behaviour before - if a star is formed it only includes a fraction of the metals added to the star-forming particle. Otherwise we can add the metals in one go a little earlier. I didn't want to change behaviour in this pull request - I don't know why it works like this though.

libgadget/sfr_eff.c
sbird: Fixed

libgadget/sfr_eff.c
sbird: This is true, but never the peak memory because we have a similar allocation whenever we use gadget_setup_thread_arrays(). So every time the treewalk gets initialised we get something like this. It is also only one int * NumPart * NumThreads which is not too much - I can put a report_memory_usage() around here so we know if this changes.

sbird: The single queue with the atomic capture would be very much slower, unfortunately.

sbird: Also static scheduling is slower by a factor of several: I checked this.

libgadget/sfr_eff.c
sbird: Oh, I see what you were saying - yes that's clearer. I'll fix it.

libgadget/partmanager.h
rainwoodman: 8 bits is a byte, not a word. Leaving some spares is always a good idea.

libgadget/sfr_eff.c
rainwoodman: By static scheduling you mean using openmp's static keyword? It uses same number of equal sized chunks as number of threads, and should have had bad imbalance between threads.

I was thinking more like this

for(i = tid; i < NumPart; i += nthreads) {

}

This would mean each thread will get into the hot regions equal likely, and we can also bound the number of particles added to the queue. I guess if this does not slow down badly here we can also shrink other places where such queue is needed.

The memory blow up is going to be a problem, if we are working with 64 threads this temporary queue is going to be bigger than the particle data.

libgadget/partmanager.h
sbird: I took them out - it is very easy to add them back if we need them. It isn't like struct particle_data is external ABI!

libgadget/sfr_eff.c
sbird: Ah, that I didn't think of doing that! It's a good idea and could be done elsewhere as well. I'll check the speed change.

sbird: We might break the pre-fetcher doing this though: even the dynamic scheduler is mostly contiguous.

libgadget/sfr_eff.c
sbird: It is still worthwhile (and it is only incrementing a pointer anyway) because there are usually much less than nactive stars and we may need memory for the Winds structure.

libgadget/sfr_eff.c
rainwoodman: is the new scheme slower? Yes the caches are likely per core and prefetching is going to be broken. But I am betting we do so much calculation in start forming particles, missing some caches is not a big problem.

libgadget/sfr_eff.c
```diff
@@ -66,56 +66,62 @@ void cooling_and_starformation(void)
     const int nactive = NumActiveParticle;
 
     if(All.StarformationOn) {
-        NewStars = mymalloc("NewStars", nactive * sizeof(int) * All.NumThreads);
-        gadget_setup_thread_arrays(NewStars, thrqueuesfr, nqthrsfr, nactive, All.NumThreads);
-        NewParents = mymalloc2("NewParents", nactive * sizeof(int) * All.NumThreads);
-        gadget_setup_thread_arrays(NewParents, thrqueueparent, nqthrsfr, nactive, All.NumThreads);
+        int narr = ceil(nactive/All.NumThreads);
```
rainwoodman: I would avoid floating point ops here. You can rely on integer divide and add 1 to make sure it is large enough.

libgadget/sfr_eff.c
sbird: It is a little slower but not noticeably so. I think you would win your bet.

libgadget/sfr_eff.c
sbird: Indeed! The current code in fact has an off-by-one error.

libgadget/sfr_eff.c
```diff
@@ -66,7 +66,7 @@ void cooling_and_starformation(void)
     const int nactive = NumActiveParticle;
 
     if(All.StarformationOn) {
-        int narr = ceil(nactive/All.NumThreads);
+        int narr = nactive/All.NumThreads+2;
```
rainwoodman: Add a comment about the magic number +2?

rainwoodman:
The new commits looks good to my eyes. But a comment about +2 would be nice.

sbird:
Done! Merging.

PR #240, opened 2018-11-06
Threadmem

sbird:
Following our discussion in the last pull, I noticed that most gadget_setup_thread_array calls were used with a schedule(static) which means they were allocating way more memory than they needed at high thread counts. Fix this.

### DIFF
--- a/libgadget/treewalk.c
+++ b/libgadget/treewalk.c
@@ -308,7 +308,10 @@ treewalk_build_queue(TreeWalk * tw, int * active_set, const int size, int may_ha
         return;
     }
 
-    tw->WorkSet = mymalloc("ActiveQueue", size * sizeof(int) * All.NumThreads);
+    /* Since we use a static schedule below we only need size / All.NumThreads elements per thread.
+     * Add 2 for non-integer parts.*/
+    int tsize = size / All.NumThreads + 2;
+    tw->WorkSet = mymalloc("ActiveQueue", tsize * sizeof(int) * All.NumThreads);
     tw->work_set_stolen_from_active = 0;
 
     int * queue = tw->WorkSet;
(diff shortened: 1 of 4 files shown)


libgadget/treewalk.c
```diff
@@ -318,7 +321,7 @@ treewalk_build_queue(TreeWalk * tw, int * active_set, const int size, int may_ha
     size_t *nqthr = ta_malloc("nqthr", size_t, All.NumThreads);
     int **thrqueue = ta_malloc("thrqueue", int *, All.NumThreads);
 
-    gadget_setup_thread_arrays(queue, thrqueue, nqthr, size, All.NumThreads);
+    gadget_setup_thread_arrays(queue, thrqueue, nqthr, tsize, All.NumThreads);
 
     /* We enforce schedule static to ensure that each thread executes on contiguous particles.*/
     #pragma omp parallel for schedule(static)
```
rainwoodman: I would suggest using a for loop in parallel block with explicity static scheduling rather than using the static clause here. Because the cause doesn't say it guarentees the number of iterations per thread is never greater than tsize.

sbird: intel's manual suggests it is: https://software.intel.com/en-us/articles/openmp-loop-scheduling

Are you just worried that chunk_size may be different? Would it be guaranteed if I manually set chunk_size?

rainwoodman:
More or a matter of readability.

I know the current standard guarantees the chunking to be single chunk as
equal as possible on different threads.
But this fact is hidden as a barely noticeable sentence in the openMP
documentation.

sbird:
That's true, but I think the extra nesting that #pragma omp parallel would also be not that readable. Would you accept more comments above the schedule(static) ?

rainwoodman:
I would still prefer avoiding trickery than documenting trickery. 

Forget about this then.  I also see a pattern that always occurs before you call gadget_setup_thread_array. Perhaps some can be merged into the function, or perhaps we can finally introduce a new structure for these -- poor man's TLS of sort. I will think about this after you merge this one, and put in the changes together. (I also felt tsize can be computed exactly).

rainwoodman:
something along these lines:

https://github.com/rainwoodman/fastpm/blob/master/libfastpm/pmapi.c#L152

which is used this way:

https://github.com/rainwoodman/fastpm/blob/master/libfastpm/transfer.c#L43

sbird:
That would be a nice cleanup indeed!

sbird:
I'll merge this and await your future cleanups.

PR #241, opened 2018-11-08
Vecadvise

sbird:
Small changes suggested by the vectorization part of intel's compiler. These give a 1% speedup on a small pure-threads problem, which is not large, but not nothing either. They made me tempted to remove the global NTasks and All.NumThreads entirely, replacing them with local const variables, but it was too much effort right now.

Vectorization advisor also made other suggestions which turned out to be not helpful or counter-productive.

### DIFF
--- a/libgadget/density.c
+++ b/libgadget/density.c
@@ -317,10 +317,10 @@ density_ngbiter(
         iter->base.symmetric = NGB_TREEFIND_ASYMMETRIC;
         return;
     }
-    int other = iter->base.other;
-    double r = iter->base.r;
-    double r2 = iter->base.r2;
-    double * dist = iter->base.dist;
+    const int other = iter->base.other;
+    const double r = iter->base.r;
+    const double r2 = iter->base.r2;
+    const double * dist = iter->base.dist;
 
     if(All.WindOn && HAS(All.WindModel, WIND_DECOUPLE_SPH)) {
         if(SPHP(other).DelayTime > 0)	/* partner is a wind particle */
@@ -336,11 +336,11 @@ density_ngbiter(
     if(r2 < iter->kernel.HH)
     {
 
-        double u = r * iter->kernel.Hinv;
-        double wk = density_kernel_wk(&iter->kernel, u);
-        double dwk = density_kernel_dwk(&iter->kernel, u);
+        const double u = r * iter->kernel.Hinv;
+        const double wk = density_kernel_wk(&iter->kernel, u);
+        const double dwk = density_kernel_dwk(&iter->kernel, u);
 
-        double mass_j = P[other].Mass;
+        const double mass_j = P[other].Mass;
 
         O->Rho += (mass_j * wk);
         O->Ngb += wk * iter->kernel_volume;
--- a/libgadget/treewalk.c
+++ b/libgadget/treewalk.c
@@ -503,10 +503,12 @@ static void ev_secondary(TreeWalk * tw)
             TreeWalkQueryBase * input = (TreeWalkQueryBase*) (tw->dataget + j * tw->query_type_elsize);
             TreeWalkResultBase * output = (TreeWalkResultBase*)(tw->dataresult + j * tw->result_type_elsize);
             treewalk_init_result(tw, output, input);
+#ifdef DEBUG
             if(!tw->UseNodeList) {
-                if(input->NodeList[0] != RootNode) abort(); /* root node */
-                if(input->NodeList[1] != -1) abort(); /* terminate immediately */
+                if(input->NodeList[0] != RootNode || input->NodeList[1] != -1)
+                     endrun(2, "Improper NodeList\n");
             }
+#endif
             lv->target = -1;
             tw->visit(input, output, lv);
         }
--- a/libgadget/treewalk.c
+++ b/libgadget/treewalk.c
@@ -43,7 +43,7 @@ static struct data_index *DataIndexTable;	/*!< the particles to be exported are 
 					   results to be disentangled again and to be
 					   assigned to the correct particle */
 
-static void ev_init_thread(TreeWalk * tw, LocalTreeWalk * lv);
+static void ev_init_thread(TreeWalk * tw, LocalTreeWalk * lv, const int NTask);
 static void ev_begin(TreeWalk * tw, int * active_set, int size);
 static void ev_finish(TreeWalk * tw);
 static int ev_primary(TreeWalk * tw);
@@ -103,7 +103,7 @@ static int data_index_compare(const void *a, const void *b)
 static TreeWalk * GDB_current_ev = NULL;
 
 static void
-ev_init_thread(TreeWalk * tw, LocalTreeWalk * lv)
+ev_init_thread(TreeWalk * tw, LocalTreeWalk * lv, const int NTask)
 {
     int thread_id = omp_get_thread_num();
     int j;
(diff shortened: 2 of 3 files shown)


libgadget/treewalk.c
```diff
@@ -608,7 +610,7 @@ treewalk_run(TreeWalk * tw, int * active_set, int size)
     if(tw->visit) {
         do
         {
-            ev_primary(tw); /* do local particles and prepare export list */
+            ev_primary(tw, NTask); /* do local particles and prepare export list */
```
rainwoodman: NTask looks a bit out of place here.

What if you add NTask into tw, and then add a local const alias to the variable in the struct?

libgadget/density.c
```diff
@@ -317,10 +317,10 @@ density_ngbiter(
         iter->base.symmetric = NGB_TREEFIND_ASYMMETRIC;
         return;
     }
-    int other = iter->base.other;
-    double r = iter->base.r;
-    double r2 = iter->base.r2;
-    double * dist = iter->base.dist;
+    const int other = iter->base.other;
```
rainwoodman: I suspect this is where most of the gain is from. 

A comment, so these consts are not removed in the future by accident.

sbird: Actually, no, most of the gain is in the treewalk and shows up in the short-range gravity walk (curiously).

libgadget/treewalk.c
sbird: How about const int NTasks = NTask, so we just reuse the current global?

sbird: and then there is no extra argument

rainwoodman:
It requires more changes. Also now you have different names for the same
variable in the same scope.

sbird:
True. What about a function like:

inline int get_NTask()
{
    int NTask;
    MPI_Comm_size(MPI_COMM_WORLD, &NTask);
    return NTask;
}

in a header

and use:
const int NTask = get_NTask();

wherever we need it and get rid of the global NTask variable entirely? That's the true global cleanup. I would guess that MPI_Comm_size just reads a global variable anyway, right?

rainwoodman:
-1 on this.

1. This would make it even harder to allow running on an arbitrary MPI
Communicator  -- it's just kicking the can down the road. MPI_COMM_WORLD is
a global too.

2. In some functions frequently called functions that uses NTask, this
would be slower than accessing the global variable.

I think comm will eventually get into many structures through their
constructions, and those structures can cache values like NTask
appropriately at initialization time.
Then the local routines can alias immutable values like NTask with const to
explicitly avoid a pointer deference.

sbird:
ok, good point. I'll add it to tw.

sbird:
How's that? I ended up removing the allvars.h import entirely so I didn't have two names for the same variable.

sbird:
ec57dd5 speeds up the hydro example on my desktop by a factor of 3! It moves so fast now I can't see the text. This is even applied as a single commit to the master branch. It totally dominates over the rest of the PR!

gadget/params.c
```diff
@@ -425,7 +426,7 @@ void read_parameter_file(char *fname)
         All.GravitySoftening = param_get_double(ps, "GravitySoftening");
         All.GravitySofteningGas = param_get_double(ps, "GravitySofteningGas");
 
-        All.ImportBufferBoost = param_get_double(ps, "ImportBufferBoost");
+        init_treewalk(param_get_double(ps, "ImportBufferBoost"));
```
rainwoodman: This is a good move! We may want to gradually standarlize to objects like TreeWalkConfiguration that are used to initialize different modules.

libgadget/density.c
rainwoodman: Likely because the compiler could figure out these are consts, but couldn't by itself in the other case.

libgadget/slotsmanager.c
```diff
@@ -61,7 +61,7 @@ int
 slots_convert(int parent, int ptype, int placement)
 {
     /*Set old slot as garbage*/
-    if(P[parent].PI > 0 && SLOTS_ENABLED(P[parent].Type))
+    if(P[parent].PI >= 0 && SLOTS_ENABLED(P[parent].Type))
```
rainwoodman: Wow. How could this happen...

libgadget/treewalk.c
```diff
@@ -21,6 +18,9 @@ static int *Exportnodecount;
 static int *Exportindex;
 static int *Send_offset, *Send_count, *Recv_count, *Recv_offset;
 
+/*!< Memory factor to leave for (N imported particles) > (N exported particles). */
+static double ImportBufferBoost;
```
rainwoodman: Just looking at this. I am thinking maybe in the future we can have a 
extern struct TreeWalkConfig {

} TreeWalkConfig

In treewalk.h

and 

struct TreeWalkConfig TreeWalkConfig;

in treewalk.c

Then instead of passing in unamed parameters with init_treewalk, we can explicitly set the parameters at in params.c before calling init_treewalk_module().

libgadget/treewalk.h
```diff
@@ -89,6 +89,8 @@ struct TreeWalk {
     TreeWalkNgbIterFunction ngbiter;     /* called for each pair of particles if visit is set to ngbiter */
     TreeWalkProcessFunction postprocess; /* postprocess finalizes quantities for each particle, e.g. divide the normalization */
     TreeWalkProcessFunction preprocess; /* Preprocess initializes quantities for each particle */
+    int NTask; /*Number of MPI tasks*/
```
rainwoodman: Where did you initialize this?

libgadget/density.c
sbird: That's probably true

gadget/params.c
sbird: Really good idea!

libgadget/treewalk.c
```diff
@@ -321,14 +321,13 @@ treewalk_build_queue(TreeWalk * tw, int * active_set, const int size, int may_ha
     tw->WorkSet = mymalloc("ActiveQueue", tsize * sizeof(int) * tw->NThread);
```
rainwoodman: What about adding a comment like "Watch out: tw->WorkSet may change a few lines later due to the realloc"?

libgadget/treewalk.c
```diff
@@ -135,8 +140,11 @@ ev_free_threadlocals()
 }
 
 static void
-ev_begin(TreeWalk * tw, int * active_set, int size)
+ev_begin(TreeWalk * tw, int * active_set, const int size)
 {
+    const int NumThreads = omp_get_max_threads();
+    MPI_Comm_size(MPI_COMM_WORLD, &tw->NTask);
```
sbird: tw->NTask initialised here, calling MPI_Comm directly.

libgadget/slotsmanager.c
sbird: I know, ouch! I guess it is just so rare it never happened in tests before. Introduced by the starform branch.

libgadget/treewalk.c
sbird: Also a good idea

libgadget/treewalk.c
sbird: Done!

rainwoodman:
Seems like we can get rid of Nexport's atomic operations with a static scheduling too (since no thread would export more than the number of particles assigned to it). But not sure if there will be gains of any sort -- I am expect Nexport to be accessed very sparsely.

rainwoodman:
All resolved.

PR #242, opened 2018-11-12
Update mpsort and bigfile to latest upstream.

rainwoodman:
This brings some speed improvements recently found from FastPM.


sbird:
You don't really need my signoff, but you have it!

rainwoodman:
OK. Merging then.

PR #243, opened 2018-11-24
Update bigfile

rainwoodman:
A typo fix, and a refactoring of the grouping routines.


sbird:
I don't really know the code, but looks ok to me

rainwoodman:
Yes. Let's hope there are no more silly typos. There was a bug when number of ranks > 2^16 in this version that got discovered in one of the large runs. It is substantially faster (a factor of 10) than the old version for small files (e.g. stars).

sbird:
What machine are you running on that has > 2^16 ranks?

rainwoodman:
Cori. 65536 if I use 1024 nodes.

rainwoodman:
It was FastPM, so I do not need to worry about the tree preferring threads over ranks.

sbird:
Ah, I see. Makes sense. You run on KNL with 1 rank per physical core and don't use the hyperthreading?

rainwoodman:
Sort of.

The optimal setting was 64x4. 64x2 is within error margin. 128x2 was
noticeably slower, for whatever reason.

sbird:
Thanks

PR #244, opened 2019-01-11
Dependfix

sbird:
A few small fixes from me: fix a (spurious) uninitialised variable warning, make it easier for the build system to regenerate the pfft dependency and fix a hang when it should crash with an error message.

### DIFF
--- a/depends/Makefile
+++ b/depends/Makefile
@@ -1,7 +1,8 @@
 CONFIG = ../Options.mk
 include $(CONFIG)
 
-LIBRARIES=lib/libbigfile-mpi.a lib/libmpsort-mpi.a lib/libpfft_omp.a
+LIBRARIES=lib/libbigfile-mpi.a lib/libmpsort-mpi.a
+FFTLIBRARIES=lib/libpfft_omp.a lib/libfftw3_mpi.a lib/libfftw3_omp.a
 depends: $(LIBRARIES)
 
 lib/libmpsort-mpi.a: mpsort/mpsort-mpi.c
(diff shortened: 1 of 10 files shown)


depends/Makefile
```diff
@@ -1,7 +1,7 @@
 CONFIG = ../Options.mk
 include $(CONFIG)
 
-LIBRARIES=lib/libbigfile-mpi.a lib/libmpsort-mpi.a lib/libpfft_omp.a
+LIBRARIES=lib/libbigfile-mpi.a lib/libmpsort-mpi.a lib/libpfft_omp.a lib/libfftw3_mpi.a lib/libfftw3_omp.a
```
rainwoodman: Do we link against these fftw3 libraries explicitly?

depends/Makefile
```diff
@@ -25,6 +25,7 @@ lib/libpfft_omp.a:
 
 clean:
 	rm -rf $(LIBRARIES)
+	rm -rf tmp-pfft-*/double
+	rm -rf tmp-pfft-*/single
```
rainwoodman: Sometimes I find it more frequent to clean fast libraries but keep the fft library.

Perhaps adding a clean-all target that cleans all and keep clean as clean-fast?

rainwoodman:
Wouldn't it be easier to simply fix cooling.c instead?

Before this change,
- positive : non-collective crash
- non-positive : collective crash

After this change, endrun 
- non-zero : non-collective crash
- zero : collective crash

If something should have been a non-collective crash but raised a collective crash code, it may get stuck, regardless of this fix.

However, after the PR we lose the ability to specify an error code in collective crash (always needs to be zero now).

In any case we do need to audit the existing endrun calls with negative parameter to see if they really collective.

sbird:
What is the use case for a collective crash? When MPI_Abort is called everything exists anyway.

I will audit the existing endrun( calls.

depends/Makefile
sbird: Yes - I took the list from Makefile.rules.

depends/Makefile
sbird: How about 
clean-all: clean clean-fft
clean-fft: rm -rf tmp-pfft-*/double
 That preserves back-compat because the meaning of clean does not change

sbird:
Maybe we should just make all crashes non-collective by making the barrier be 
if(!crash)
MPI_Barrier()

Remember I tried to remove the Barrier entirely once but the code became much slower

depends/Makefile
rainwoodman: What about:

clean: clean-fast clean-fft

back-compat is less important than getting it right in the long term.

sbird: Ok, I'll do that

sbird:
How's that? Implemented the suggestions and audited the collective endrun calls - most wanted to be non-collective anyway.

sbird:
(most with negative arguments: the 0 arguments are all fine being collective)

rainwoodman:
Great. This looks good to me now.

sbird:
Thanks, merged.

PR #245, opened 2019-01-20
Fof threading fix

sbird:
This branch fixes a deadlock in the FoF threading code, as well as generating FoF tables by using RestartFlag==3. The deadlock happens because we lock two particles at once, without any requirement on their ordering. The changes lock only one particle at a time. The double-locking ensured that no cycles occurred in the FoF tree: this is now ensured by merging trees always from the higher numbered one into the lower.

I tested that this works, does not deadlock, and gives the same answer as the master branch, but the code is a bit tricky, so you might want to give it a careful look-over.

### DIFF
--- a/libgadget/fof.c
+++ b/libgadget/fof.c
@@ -373,15 +373,15 @@ void fof_label_primary(void)
     myfree(FOF_PRIMARY_GET_PRIV(tw)->Head);
 }
 
-static void
-fofp_merge(int target, int other, TreeWalk * tw)
+/* This function does the actual merging of the trees. It takes two head particles, h1 and h2,
+ * and merges 2 into h1 by setting Head[h2] = h1. It does this if and only if h1 < h2 (to avoid cycles)
+ * and it does it under a lock for h2, so that h2 cannot be merged elsewhere. h1 is not locked, to avoid deadlocking.
+ * It can thus cease to be the head of the tree, but this is fine, because it will never be merged into h2.
+ * If h1 > h2, the code swaps them and tries again.
+ * returns: 1 if a merge happened, 0 if it did not. No locks are held after return.*/
+static int
+try_merge_trees(int * h1, int * h2, MyIDType * subminID, MyIDType * subminIDTask, int * Head)
 {
-    int * Head = FOF_PRIMARY_GET_PRIV(tw)->Head;
-    /* Get current HEAD of target. h1 is not locked, so by the time we use it,
-     * it may no longer really be the HEAD, but we could also link the
-     * tree by setting HEAD(h2) = target, so it doesn't matter.*/
-    int h1 = HEADa(-1, target, Head);
-    int h2 = HEADa(h1, other, Head);
     /* We are about to change the HEAD of h2, so lock it.
      * We need to make sure that h2 is really the head of the subtree,
      * or not all particles will be linked.
@@ -389,26 +389,59 @@ fofp_merge(int target, int other, TreeWalk * tw)
      * and loop until we can get a lock.*/
     while(1) {
         /*Stop if we encountered h1 along the head of this tree: means this is already merged.*/
-        if(h2 == -1)
-            return;
-        lock_particle(h2);
-        if(Head[h2] != h2) {
-            unlock_particle(h2);
-            h2 = HEADa(h1, h2, Head);
+        if(*h2 == -1)
+            return 0;
+        lock_particle(*h2);
+        if(Head[*h2] != *h2) {
+            unlock_particle(*h2);
+            *h2 = HEADa(*h1, *h2, Head);
         }
         else
             break;
     }
     /*Get the current subtree properties: this is fine because we are under the lock.*/
-    MyIDType subminID = HaloLabel[h2].MinID;
-    MyIDType subminIDTask = HaloLabel[h2].MinIDTask;
+    *subminID = HaloLabel[*h2].MinID;
+    *subminIDTask = HaloLabel[*h2].MinIDTask;
 
     /*Merge the subtree*/
-    #pragma omp atomic write
-    Head[h2] = h1;
+    if(*h2 > *h1) {
+        #pragma omp atomic write
+        Head[*h2] = *h1;
+        unlock_particle(*h2);
+        return 1;
+    }
+    else {
+        /*If h1 > h2, swap the particles and try again*/
+        unlock_particle(*h2);
+        int t = *h2;
+        *h2 = *h1;
+        *h1 = t;
+        return try_merge_trees(h1, h2, subminID, subminIDTask, Head);
+    }
+}
 
-    /*We are done with h2 and can unlock it.*/
-    unlock_particle(h2);
+static void
+fofp_merge(int target, int other, TreeWalk * tw)
+{
+    int * Head = FOF_PRIMARY_GET_PRIV(tw)->Head;
+    if(other < target)
+    {
+        int t = target;
+        target = other;
+        other = t;
+    }
+    /* Get current HEAD of target. h1 is not locked, so by the time we use it,
+     * it may no longer really be the HEAD, but we could also link the
+     * tree by setting HEAD(h2) = target, so it doesn't matter.*/
+    int h1 = HEADa(-1, target, Head);
+    int h2 = HEADa(h1, other, Head);
+
+    /*Get the current subtree properties: this is fine because we are under the lock.*/
+    MyIDType subminID, subminIDTask;
+
+    /*Try to merge h1 into h2. Return if it turns out they are already merged. h1 and h2 may change in this function.*/
+    if(!try_merge_trees(&h1, &h2, &subminID, &subminIDTask, Head))
+        return;
 
     /* Now we need to find the true current head of the new tree
      * and lock it to update minID properties.
--- a/libgadget/fof.c
+++ b/libgadget/fof.c
@@ -231,38 +231,29 @@ struct FOFPrimaryPriv {
 };
 #define FOF_PRIMARY_GET_PRIV(tw) ((struct FOFPrimaryPriv *) (tw->priv))
 
+/* Get the ultimate HEAD particle of the current tree.
+ * Read-only lockless function using atomics.*/
 static int
-HEADl(int stop, int i, TreeWalk * tw)
+HEADa(int stop, int i, const int * const Head)
 {
-    int r;
-    if (i == stop) {
-        return -1;
-    }
-//    printf("locking %d by %d in HEADl stop = %d\n", i, omp_get_thread_num(), stop);
-    lock_particle(i);
-//    printf("locked %d by %d in HEADl, next = %d\n", i, omp_get_thread_num(), FOF_PRIMARY_GET_PRIV(tw)->Head[i]);
+    int r, next;
+    #pragma omp atomic read
+    next = Head[i];
 
-    if(FOF_PRIMARY_GET_PRIV(tw)->Head[i] == i) {
-        /* return locked */
+    if(next == stop)
+        return -1;
+    /* Note that because this particle is not locked, the tree may get merged.
+     * before the next time we use this.
+     * Note that Head[i] == i is the initial state: once set it will never change.
+     * However, because this is a lock-free function, the
+     * head may have changed by the next time we want to read it.*/
+    if(next == i) {
         return i;
     }
-    /* this is not the root, keep going, but unlock first, since even if the root is modified by
-     * another thread, what we get here is on the path, */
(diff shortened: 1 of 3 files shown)


rainwoodman:
urghh multi-threading. we need to be extra careful here.  I read through the old algorithm and don't see the deadlock immediately. I will add the question inline.

libgadget/fof.c
```diff
@@ -382,40 +373,96 @@ void fof_label_primary(void)
...
+        *h1 = t;
+        return try_merge_trees(h1, h2, subminID, subminIDTask, Head);
+    }
+}
+
 static void
 fofp_merge(int target, int other, TreeWalk * tw)
 {
-    /* this will lock h1 */
-    int h1 = HEADl(-1, target, tw);
```
rainwoodman: This is the only place where we may two locks, h1, and h2.

The problem is fofp_merge can be called concurrently by two threads,
Thread 1: 
h1=a, h2=b
Thread 2:
h1=b h2=a

This tends to occur when threads are covering spatial volumes that severely overlap.

The motivation of the patch is to break this scenario.

rainwoodman: Oh. The case is wider.

Thread 1 locks h1=a (head of target/T1)
Thread 2 locks h2=b (head of target/T2)
Thread 1 waits for b between other/T1 and h2/T1,
Thread 2 waits for a between other/T1 and h2/T2

rainwoodman: Another confusion case:

Thread 1 locks h1=a (head of target/T1)
Thread 2 locks h1=b (head of target/T2)
Thread 1 waits for c between other/T1 and head of other/T1,
Thread 2 waits for b between other/T1 and head of other/T2

libgadget/fof.c
```diff
@@ -382,40 +373,96 @@ void fof_label_primary(void)
...
+        else
+            break;
+    }
+    /*Get the current subtree properties: this is fine because we are under the lock.*/
+    *subminID = HaloLabel[*h2].MinID;
+    *subminIDTask = HaloLabel[*h2].MinIDTask;
+
+    /*Merge the subtree*/
+    if(*h2 > *h1) {
+        #pragma omp atomic write
```
rainwoodman: This write seems to be atomic anyways. Why the atomic write? Do you need the flush implied by it?

libgadget/fof.c
```diff
@@ -382,40 +373,96 @@ void fof_label_primary(void)
...
+        unlock_particle(*h2);
+        return 1;
+    }
+    else {
+        /*If h1 > h2, swap the particles and try again*/
+        unlock_particle(*h2);
+        int t = *h2;
+        *h2 = *h1;
+        *h1 = t;
+        return try_merge_trees(h1, h2, subminID, subminIDTask, Head);
```
rainwoodman: Would it not work if passing in h2, h1 instead of h1, h2?

Also is it important this try is after ensuring h2 is the precise head? Given that you have unlocked `*h2`, it does not seem to be important. But then why is the section after the h2 promotion code.

libgadget/fof.c
```diff
@@ -231,38 +231,29 @@ struct FOFPrimaryPriv {
...
+HEADa(int stop, int i, const int * const Head)
 {
-    int r;
-    if (i == stop) {
-        return -1;
-    }
-//    printf("locking %d by %d in HEADl stop = %d\n", i, omp_get_thread_num(), stop);
-    lock_particle(i);
-//    printf("locked %d by %d in HEADl, next = %d\n", i, omp_get_thread_num(), FOF_PRIMARY_GET_PRIV(tw)->Head[i]);
+    int r, next;
```
rainwoodman: Perhaps you want next as a volatile, such that the browser does not try to re-read from Head[i]. The omp atomic read is confusing. Sounds like you just need a omp flush on next after this. But if next is declared a volatile I *think* you do not need the flush neither.

libgadget/fof.c
```diff
@@ -434,26 +481,23 @@ fof_primary_ngbiter(TreeWalkQueryFOF * I,
     }
     int other = iter->base.other;
 
-/* #pragma omp critical (_fofp_merge_) */
+    if(lv->mode == 0) {
```
rainwoodman: Nothing changes here other than indentation?

libgadget/fof.c
```diff
@@ -382,40 +373,96 @@ void fof_label_primary(void)
...
+            return 0;
+        lock_particle(*h2);
+        if(Head[*h2] != *h2) {
+            unlock_particle(*h2);
+            *h2 = HEADa(*h1, *h2, Head);
+        }
+        else
+            break;
+    }
+    /*Get the current subtree properties: this is fine because we are under the lock.*/
```
rainwoodman: This amounts to finding a snapshot of a tree branch. After the lock is obtained no other thread will modify the branch. Therefore you can assert the MinID of the branch is accurate before you release the lock.

libgadget/fof.c
```diff
@@ -382,40 +373,96 @@ void fof_label_primary(void)
...
+    *subminID = HaloLabel[*h2].MinID;
+    *subminIDTask = HaloLabel[*h2].MinIDTask;
+
+    /*Merge the subtree*/
+    if(*h2 > *h1) {
+        #pragma omp atomic write
+        Head[*h2] = *h1;
+        unlock_particle(*h2);
```
rainwoodman: You do not update the MinID at h1, because you are not owning a snapshot of h1. 

You may have skipped the splay step. The old algorithm ensures the tree is shallow by always setting the HEAD of any known node to the latest known root tree, to ensure HEAD() can almost always return in O(1). I don't think you are doing the splay here.

libgadget/fof.c
```diff
@@ -382,40 +373,96 @@ void fof_label_primary(void)
...
+    }
+}
+
 static void
 fofp_merge(int target, int other, TreeWalk * tw)
 {
-    /* this will lock h1 */
-    int h1 = HEADl(-1, target, tw);
-    /* stop looking if we find h1 along the path (because it is already owned by us) */
-    int h2 = HEADl(h1, other, tw);
```
rainwoodman: Seems like if we can change this to HEADa, and then unlock(h1) before locking(h2) a few lines later -- would pretty much capture the spirit of your patch?

libgadget/fof.c
sbird: That's right

libgadget/fof.c
sbird: That's right - I couldn't figure out how to avoid deadlocking in the splay. So the new code is slower unfortunately, but not deadlocking is useful

libgadget/fof.c
sbird: I wasn't sure I needed it, but I wanted any threads concurrently inside of HEADa() to get to h1 instead of h2. I was worried that if a read is in progress the other threads may pick up half of h2 and half of h1.

libgadget/fof.c
sbird: Passing in h2, h1 wouldn't swap the heads in the calling function, so MinID wouldn't be set correctly.

I unlock h2 because once we swap them it is the former h1 that needs to be promoted and locked. I think that if I swap first and them promote, the promoted h2 may again be less than h1 (because the indices decrease as we go up the tree). I might be wrong.

libgadget/fof.c
sbird: I'm not sure - I don't know the rules for volatile well enough

libgadget/fof.c
sbird: I'm worried a bit about a thread modifying the branch after we release the lock and before we get the lock on h1 later. I *think* it is ok, because we only care about MinID for the new tree head, h1, and the other thread will modify that itself if needed. But I'm not sure.

libgadget/fof.c
rainwoodman: OK. After reading a bit I am no longer opposing having  atomic annotations everywhere. The standard prefers doing so, and leave the detail to the compiler.

libgadget/fof.c
rainwoodman: I think the alternative algorithm I suggested may work.

Also keep in mind the splay does not have to be accurate: as long as we are always making the tree shallower it will help. If you do not splay, then FOF slows down quite badly once you get very massive halos. Of course this is still much better than using a double linked list. (n vs logn vs 1).

libgadget/fof.c
rainwoodman: I think this is OK because the reduction of MinID is associative: as long as the MinID is reduced to the true head at some point then the MinID of the final head will be correct.

rainwoodman:
Do you have a test case that reproduces the deadlock? I want to give this a
try.

libgadget/fof.c
sbird: Thanks, that sounds right.

libgadget/fof.c
sbird: I agree with this analysis

libgadget/fof.c
sbird: HEAD() still splays when called later, so it is only delayed. 

I suppose one could also do an omp atomic splay (ie, without locking each particle) while the lock on the head of the subtree is held. Then we would really need the atomic.

libgadget/fof.c
sbird: It is important that we have either both locked at once, or we enforce h1 < h2: otherwise we get cycles (which deadlock too and are much easier to hit than the original deadlock).

libgadget/fof.c
```diff
@@ -382,40 +373,96 @@ void fof_label_primary(void)
...
+        other = t;
     }
+    /* Get current HEAD of target. h1 is not locked, so by the time we use it,
+     * it may no longer really be the HEAD, but we could also link the
+     * tree by setting HEAD(h2) = target, so it doesn't matter.*/
+    int h1 = HEADa(-1, target, Head);
+    int h2 = HEADa(h1, other, Head);
 
-    /* h2 as a sub-tree of h1 */
-    FOF_PRIMARY_GET_PRIV(tw)->Head[h2] = h1;
```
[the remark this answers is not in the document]
sbird: Not sure. We need to enforce that there is an ordering somehow, or we can get h2 added below h1 by another thread while we are waiting for the lock.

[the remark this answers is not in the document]
rainwoodman: Here we only hold at most one lock at any time. 

Could you be more elaborate `waiting for the lock`? which lock?
h2 is locked and unlocked during the last three lines.

rainwoodman:
So a 32**3 problem on 256 cores. I can use a 2x64 configuration on a KNL node. It will not be exactly fast, but I think we can ask many snapshots and use very few time steps. dm-only equally likely as hydro?

sbird:
It is the fof for the blackhole seeding that deadlocks, so you do need hydro, unfortunately. Fof for snapshots doesn't run often enough. It is a very rare deadlock.

libgadget/fof.c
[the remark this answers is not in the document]
rainwoodman: Yes I just ran into this. The ordering would solve this. I am leaning towards your solution is the only solution. But let me think about this a bit more.

rainwoodman:
This sounds like implementing a dcas (two locks at the same time) with cas.  There seems to be a standard algorithm that we can adopt:  
https://en.wikipedia.org/wiki/Double_compare-and-swap

libgadget/fof.c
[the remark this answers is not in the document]
rainwoodman: Yes. I bet splays can be added. with easy.

My main reservation over your algorithm is that it feels trickier (harder to elaborate) than the existing one. This viewpoint must have been biased, as I am more familiar with the old algoirthm, but it enhanced my worry that there are corners that's not yet caught.

I think a retry algorithm is needed if we convert a double lock to two single locks. (not sure if there is a theorem, but it does smell like the right complexity mentioned in that paper). 

In this problem, the retry lock would be implemented like this: as we try to lock the second particle, we if we encounter any particle that is already locked, then we lock at the index:
- If the index of the newly found is less than the locked one we already have, wait on the lock.
- If not, unlock the first locked particle, and retry, optionally sleep.

This way if there is a loop of deadlocks, it will always untangle from one of the threads first. Notice that retry and ordering are both necessary, as the older algorithm was operating under the assumption that at the point of merge we do need to locks on both branches.

If we do not require both branches to be snapshots, then  we may be attaching the head to an older value that sits below the current branch, thus missing some merges. 

It is the lack of retry in your algorithm makes suspect there are corner cases. How do you guarantee if one of the branch advances the merge is still 'correct'? I think the ordering flip is thekey, but the ordering re-arrengement is getting into my way of understanding your algorithm.

PR #246, opened 2019-01-26
Fofthreadingfix2

sbird:
This implements the double compare and swap algorithm suggested in: 

https://github.com/MP-Gadget/MP-Gadget/pull/245#discussion_r250711455

Fixes the deadlock, passes the tests!

I like this algorithm much better - I was also unsure about the corner cases in the last one. I think I found one when testing this, in fact.

One thing I am unsure about is whether we need to take the lock in update_root. I think we do not need to if the write is atomic, because the head is preserved. But then I think we might need an omp atomic write annotation.

### DIFF
--- a/libgadget/fof.c
+++ b/libgadget/fof.c
@@ -259,7 +259,9 @@ HEADl(int stop, int i, int locked, int * Head)
     }
 //    printf("locking %d by %d in HEADl stop = %d\n", i, omp_get_thread_num(), stop);
     /* Try to lock the particle. Wait on the lock if it is less than the already locked particle,
-     * or if such a particle does not exist.*/
+     * or if such a particle does not exist. We could use atomic to avoid the lock, but this
+     * makes the locking code less clear, and doesn't lead to much of a speedup: the splay
+     * means that the tree is shallow.*/
     if(locked < 0 || i < locked) {
         lock_particle(i);
     }
@@ -285,6 +287,8 @@ HEADl(int stop, int i, int locked, int * Head)
     return r;
 }
 
+/* Rewrite a tree so that all values in it point directly to the true root.
+ * This means that the trees are O(1) deep and speeds up future accesses. */
 static void
 update_root(int i, int r, int * Head)
 {
@@ -445,7 +449,10 @@ fofp_merge(int target, int other, TreeWalk * tw)
     }
 
     /* h1 must be the root of other and target both:
-     * do the splay to speed up future accesses.*/
+     * do the splay to speed up future accesses.
+     * We do not need to have h2 locked, because h2 is
+     * now just another child of h1: these do not change the root,
+     * they make the tree shallow.*/
     update_root(target, h1, Head);
     update_root(other, h1, Head);
 
--- a/libgadget/fof.c
+++ b/libgadget/fof.c
@@ -298,11 +298,10 @@ update_root(int i, const int r, int * Head)
     int t = i;
     do {
         i = t;
-        #pragma omp atomic capture
-        {
-            t = Head[i];
-            Head[i]= r;
-        }
+        #pragma omp atomic read
+        t = Head[i];
+        #pragma omp atomic write
+        Head[i]= r;
     } while(t != i);
 }
 
--- a/libgadget/fof.c
+++ b/libgadget/fof.c
@@ -253,7 +253,7 @@ struct FOFPrimaryPriv {
 static int
 HEADl(int stop, int i, int locked, int * Head)
 {
-    int r;
+    int r, next;
     if (i == stop) {
         return -1;
     }
@@ -273,14 +273,17 @@ HEADl(int stop, int i, int locked, int * Head)
         }
     }
 //    printf("locked %d by %d in HEADl, next = %d\n", i, omp_get_thread_num(), FOF_PRIMARY_GET_PRIV(tw)->Head[i]);
+    /* atomic read because we may change
+     * this in update_root without the lock: not necessary on x86_64, but avoids tears elsewhere*/
+    #pragma omp atomic read
+    next = Head[i];
 
-    if(Head[i] == i) {
+    if(next == i) {
         /* return locked */
         return i;
     }
     /* this is not the root, keep going, but unlock first, since even if the root is modified by
      * another thread, what we get here is on the path, */
-    int next = Head[i];
     unlock_particle(i);
 //    printf("unlocking %d by %d in HEADl\n", i, omp_get_thread_num());
     r = HEADl(stop, next, locked, Head);
@@ -290,13 +293,17 @@ HEADl(int stop, int i, int locked, int * Head)
 /* Rewrite a tree so that all values in it point directly to the true root.
  * This means that the trees are O(1) deep and speeds up future accesses. */
 static void
-update_root(int i, int r, int * Head)
+update_root(int i, const int r, int * Head)
 {
-    while(Head[i] != i) {
-        int t = Head[i];
-        Head[i]= r;
+    int t = i;
+    do {
         i = t;
-    }
+        #pragma omp atomic capture
+        {
+            t = Head[i];
+            Head[i]= r;
+        }
+    } while(t != i);
 }
 
 static int
--- a/libgadget/fof.c
+++ b/libgadget/fof.c
@@ -231,6 +231,25 @@ struct FOFPrimaryPriv {
 };
 #define FOF_PRIMARY_GET_PRIV(tw) ((struct FOFPrimaryPriv *) (tw->priv))
 
+/* This function walks the particle tree starting at particle i until it reaches
+ * a particle which has Head[i] = i, the root node (particles are initialised in
+ * this state, so this is equivalent to finding a particle which has yet to be merged).
+ * Each particle is locked along the way, and unlocked once it is clear that it is not a root.
+ * Once it reaches a root, it returns that particle number with the lock still held.
+ * There are two other arguments:
+ *
+ * stop: When this particle is reached, return -1. We use this to find an already merged tree.
+ *
+ * locked: This particle is already locked (setting locked = -1 means no other locks are taken).
+ *         When we try to lock a particle locked by another thread, the code checks whether
+ *         it can safely take a second lock. If it cannot, it returns -2, with the expectation
+ *         that the first lock is released before a retry.
+ *
+ * Returns:
+ *      root particle if found
+ *      -1 if stop particle reached
+ *      -2 if locking might have deadlocked.
+ */
 static int
 HEADl(int stop, int i, int locked, int * Head)
 {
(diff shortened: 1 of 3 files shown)


libgadget/fof.c
```diff
@@ -232,35 +232,46 @@ struct FOFPrimaryPriv {
...
-HEADl(int stop, int i, TreeWalk * tw)
+HEADl(int stop, int i, int locked, int * Head)
 {
     int r;
     if (i == stop) {
         return -1;
     }
 //    printf("locking %d by %d in HEADl stop = %d\n", i, omp_get_thread_num(), stop);
-    lock_particle(i);
+    /* Try to lock the particle. Wait on the lock if it is less than the already locked particle,
```
rainwoodman: I think we may be able to add an optimization here (as done in your other PR), to advance towards the root without locking first. Because we only ever grow the tree in depth I think it should work; but perhaps leave this for the future?

libgadget/fof.c
```diff
@@ -232,35 +232,46 @@ struct FOFPrimaryPriv {
 #define FOF_PRIMARY_GET_PRIV(tw) ((struct FOFPrimaryPriv *) (tw->priv))
 
 static int
-HEADl(int stop, int i, TreeWalk * tw)
+HEADl(int stop, int i, int locked, int * Head)
```
rainwoodman: Now that it returns three possible values probably write a comment -1 is found the stop and -2 is lock failed.

sbird: Good idea, will add it.

libgadget/fof.c
sbird: I thought about this, but I couldn't immediately figure out how to get the lock retry logic written in an clear and obviously correct way. I think that with a splay it shouldn't lead to a big speedup: the tree should be O(1) deep so we don't lock many particles. In the other PR it was more important because there was no splay.

libgadget/fof.c
```diff
@@ -386,34 +397,38 @@ static void
...
+    int h1, h2;
+
+    do {
+        h1 = HEADl(-1, target, -1, Head);
+        /* stop looking if we find h1 along the path (because it is already owned by us) */
+        h2 = HEADl(h1, other, h1, Head);
+        /* We had a lock already taken on h2 by another thread.
+         * We need to unlock h1 and retry to avoid deadlock loops.*/
+        if(h2 == -2)
+            unlock_particle(h1);
```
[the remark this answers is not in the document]
sbird: Is it also true that in the retry h1 will no longer be the first root node?

libgadget/fof.c
```diff
@@ -386,34 +397,38 @@ static void
...
+        }
+        //printf("unlocking %d by %d in merge\n", h2, omp_get_thread_num());
+        unlock_particle(h2);
     }
-    //printf("unlocking %d by %d in merge\n", h2, omp_get_thread_num());
-    unlock_particle(h2);
 
-    update_root(target, h1, tw);
-    update_root(other, h1, tw);
+    /* h1 must be the root of other and target both:
```
rainwoodman: If h1 is locked, then any object on path from target to h1 can be splayed.

After h2 is unlocked, the root of any object under h2 becomes h1. As h1 is still locked, we can also safely splay any path along other to h1. That's why the second loop is after unlocking of h2. 

Does this make sense? If so, add a comment so we don't miss it next time.

rainwoodman:
Yes. This change set looks cleaner indeed. Go ahead and add the atomic read / write decorators. The standard actually requires them. We have been relying that on x64 aligned read / write of int64 are atomic. This assumption is not portable; but on x64 the decorators will not compile to instructions, I think.

libgadget/fof.c
rainwoodman: Indeed. Let's have comment to note the reason this apparently obvious optimization is skipped.

libgadget/fof.c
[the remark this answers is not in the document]
rainwoodman: Why not? If the thread locking h1 (t1) is swapped out of the context, then there wouldn't have been any progress, and we will run into the same loop. 

This situation explains why we never want to oversubscribe the cores -- the thread making progress may be out of context, and while the core spins.

If we add a sleep(which yields the context) then the second spinlock becomes somewhat of a mutex, and would have performed better under core over-subscription.

libgadget/fof.c
sbird: Yes, I see - there is no need to hold the snapshot of h2 because everything looking at it will reach h1 anyway. I will add a comment.

libgadget/fof.c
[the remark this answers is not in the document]
sbird: I hadn't thought about oversubscribed cores. Is sleep(0); enough?

sbird:
Looking at it closer, does update_root actually guarantee the trees are short? It only runs vertically, so if we have a very wide tree:
     h2
|  |  |  | 
a b c d

and we are starting at 'a', then after the root of h2 is set to h1, we will have still:
     h1 
  |     | 
  a   h2
      |  |  | 
     b  c d

so b, c, d can still have deep trees, if they acquired them before being merged to h2. The splay still makes things faster, but I'm not sure it is really O(1).

sbird:
I have some worries about the omp atomic annotations for the splay. 

The read/write form gives the following warning with gcc 8.2.1 (my fedora laptop):

fof.c:297:30: warning: parameter ‘r’ set but not used [-Wunused-but-set-parameter]
 update_root(int i, const int r, int * const Head)

which is alarming because the parameter is in fact used. If I remove the atomic annotations the warning disappears. godbolt suggests identical code is generated with and without the annotations. (https://gcc.godbolt.org/z/kEugNT : the capture form is there as well)

Seems like a buggy compiler warning. Shall we just ignore it? Or is it trying to tell me something important?

sbird:
More investigation reveals that the warning is only on with -Wextra and is also triggered by the innocuous:

int i;
void
set(int r)
{
    #pragma omp atomic write
    i = r;
}

so I will put it down to a harmless compiler false positive and ignore it.

rainwoodman:
Indeed. I can imagine such a bug easy to make and not so easy to fix -- though I didn't find a bug report.

libgadget/fof.c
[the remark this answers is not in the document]
sbird: ok, I will remove it again

sbird:
I think I'm ready to merge when travis goes green.

rainwoodman:
I thought about starving again. I think it is a good idea to add an empty loop of a few tens / hundreds of cycles before relocking the first particle in the recover / slow path, to increase the chance the proceeding thread gets the lock. Without this condition the proceeding thread would only have a 50% chance of getting the lock, while there is a 50% chance this thread gets the lock back immediately.

Not a sleep -- sleep and variants seem to cause a context switch (the man pages suggests it is synced to the context switch frequency) and is thus much slower.

rainwoodman:
it is not always available  -- here is a note from the man page.

       POSIX systems on which sched_yield() is available define _POSIX_PRIORITY_SCHEDULING in <unistd.h>.

The behavior is not clearly defined if there is exactly one thread on the core -- it may be a single empty cycle in this case and wouldn't help as much as a short busy loop.

sbird:
Ok. Is this a pure performance optimization or would it also fix a starvation deadlock?

Do we implement busy looping like:
volatile i=0;
for(1..100) i++;

or is there a better way?

rainwoodman:
It would be a pure performance optimization derived from a pure theoretical perspective. We don't have a clue if it helps at all or to what extent. And I can't think of a better way than your loop.

I did dig the implementation of unlock though. At least on some platforms (GNU Hurd) there is a `wake up` that seems to wake up other threads waiting on the same address after unlock. Let me run an experiment.

sbird:
heh, GNU Hurd has SMP now? Must be a stable release coming soon then. :)

rainwoodman:
Ah. it's possible that's not even SMP and hence they need the wake up. That explains why the wake up implementation looks like a system call.

My experiment shows on Haswell,
- 'not doing anything' is fastest when 32 threads are all compete against the same double lock. Using 100 ~ 1000 busy loops is roughly the same as doing a `sched_yield`. Using busy loop of 10 iterations is slightly faster than `sched_yield` but slower than not doing anything. 

on KNL, 
- an loop of 10 iteration is the fastest.

The difference is about 1%. The test case is much more extreme than we will ever encounter in real code. So the conclusion is we shall do nothing about the starvation.

Here is the code:

```
#include <stdio.h>
#include <unistd.h>
#include <pthread.h>
#include <omp.h>
#include <sys/time.h>
pthread_spinlock_t lock[2] = {0, 0};
int value = 0;
int mode = 'n';
int main(int argc, char * argv[0])
{

    struct timeval tv[2];

    if(argc == 2) {
        mode = argv[1][0];
    } 
    pthread_spin_init(lock, 0);
    pthread_spin_init(lock + 1, 0);

    omp_set_num_threads(32);
    #pragma omp parallel
    {
        int tid = omp_get_thread_num();
        printf(":Thread :%d\n", tid);

        int j;
        int values[200];
        int retries = 0;
        #pragma omp barrier
        #pragma omp single
        {
            gettimeofday(tv + 0, NULL);
        }
        #pragma omp barrier

        for(j = 0; j < 10; j ++) {
            while(1) {
                pthread_spin_lock(lock);
                if(!pthread_spin_trylock(lock + 1)) {
                    retries ++;
                    pthread_spin_unlock(lock);
                    if(mode == 's') {
                        sched_yield();
                    }
                    if(mode == 'l') {
                        volatile int i;
                        for(i = 0; i < 1000; i ++) {
                            continue;
                        }
                    }
                    continue;
                } else {
                    break;
                }
            }
            volatile int i;
            for(i = 0; i < 10000000; i ++) {
                continue;
            }
            value = value + 1;
            values[j] = value;
            pthread_spin_unlock(lock + 1);
            pthread_spin_unlock(lock);
        }
        #pragma omp barrier
        #pragma omp single
        {
            gettimeofday(tv + 1, NULL);
        }
        #pragma omp critical
        {
            printf(":Thread :%d retries %d\n", tid, retries);
            for(j = 0; j < 10; j ++) {
                printf("%d, ", values[j]);
            }
            printf("\n");
        }
    }
    printf("mode = %c, time = %g\n", mode, (tv[1].tv_sec - tv[0].tv_sec) * 1.0 + (tv[1].tv_usec - tv[0].tv_usec) * 1e-6);
    printf("value = %d\n", value);
    return 0;
}
```

sbird:
Ok - thanks for the testing! I think the fact that you looked up Hurd code means this PR took us past some new threshold for extreme performance analysis. 

How is our overall performance on KNL, by the way? Does it still suck?

rainwoodman:
I didn't rerun on KNL; but I don't think we are sucker than expected.

I recall a 100% increase in wall clock time comparing to Haswell is reasonable for many applications on the same core number. After some very difficult optimization it may be a comparable speed. Remember the clock speed is only half of haswell. There are 2x number of cores, so I'd think it won't be too bad given your threading improvements. 

I'll put a proper comparison high on my todo list. With small box hydro example that caused the deadlock?

sbird:
Can you try the lya example? 

The comparison I tried before was 1 stampede2 skylake node vs 1 stampede 2 knl node and the knl was about a factor of two slower. But that was before the threading improvements, so I hope it's better now. 

I should do a test too...

PR #248, opened 2019-01-30
Euclid Comparison

rainwoodman:
This ticket tracks the progress with Euclid Comparison.

Wish we have the z=2.0 power spectra. (seems obtainable)

TODO:
 - [x] convert IC to bigfile (A gadget to bigfile converter -- mind the unit of velocity) on NERSC
 - [x] subsample and run a small test?
 - [x] run the full run
 - [ ] compute Pk.


sbird:
Don't forget to set IncludeRadiation = 0

rainwoodman:
Good idea. We probably want a smaller box for debugging though.

sbird:
I will take this

sbird:
Yu: did you actually find a copy of their ICs? The link is broken for me.

rainwoodman:
These files looked familiar. At some point I might have thought they are the Euclid ICs. :

https://www.ics.uzh.ch/~aurel/euclidIC_2048/

sbird:
They are: I emailed Aurel and he put them back :-)

sbird:
Our results are much better than Gadget-3 (although there is some mode scatter because I didn't use the same P(k) estimator):

mpgadget-euclid

rainwoodman:
Wow. Not all TreePMs are made equal... Do they have high z comparisons? e.g. z=2?

sbird:
Alas, he does not: he said he would have to re-run all the sims.

sbird:
I think possibly helpful is that looking at the cpu.txt, MP-Gadget is really more of a 'PM' for this run: > 50% of runtime is in the PM module.

sbird:
Merging - I tried to make a P(k) from the snapshot with nbodykit but got an internal error from Frontera's MPI4py. Looks ok to me anyway.

PR #249, opened 2019-01-30
Do not use huge bunchsizes that causes overflow.

rainwoodman:
This fixes a crash when I am running with a single rank on a machine
with 16GB of memory.


sbird:
Good idea. Maybe a comment though?

sbird:
Looks good sign off from me.

PR #252, opened 2019-03-22
New cooling code.

sbird:
This branch reworks the cooling code. The main aim of this was to support the newer cooling rates and to add support for self-shielding following Rahmati 2012 (also used in Illustris, although the implementation I chose is a little different). To do this I had to understand the cooling code, and it was easier to rewrite that from scratch. To make sure I did this right, I also added a fair number of unit tests and comments. This follows heavily the python cooling code found in: https://github.com/sbird/fake_spectra/blob/master/fake_spectra/rate_network.py

The final test was to run the hydro example calling both the old cooling code and the new cooling code on each particle, to ensure they gave the same answer.

The cooling rates supported are now:
- The ones used in old Gadget (Cen 92)
- The ones used in Nyx
- The ones used in Jamie Bolton's Sherwood simulations, which are old Gadget with updated recombination tables and a bug fix.

I have not as yet changed the defaults to the new cooling rates. I wanted to let you review this code and check that it does give the same answer for you before doing this. If you agree I will also enable self shielding by default.

### DIFF
--- a/gadget/params.c
+++ b/gadget/params.c
@@ -339,16 +339,6 @@ create_gadget_parameter_set()
     return ps;
 }
 
-/* Structure to hold local parameter values so they can easily be broadcast.
- * These are for submodules where we do the initialization right here and so don't want to use All.*/
-struct Local
-{
-    /*Treewalk parameter*/
-    double ImportBufferBoost;
-    /* Cooling model parameters*/
-    struct cooling_params coolpar;
-};
-
 /*! This function parses the parameterfile in a simple way.  Each paramater is
  *  defined by a keyword (`tag'), and can be either of type douple, int, or
  *  character string.  The routine makes sure that each parameter appears
@@ -357,10 +347,9 @@ struct Local
  */
 void read_parameter_file(char *fname)
 {
-    struct Local locals;
-    if(ThisTask == 0) {
+    ParameterSet * ps = create_gadget_parameter_set();
 
-        ParameterSet * ps = create_gadget_parameter_set();
+    if(ThisTask == 0) {
 
         if(0 != param_parse_file(ps, fname)) {
             endrun(1, "Parsing %s failed.", fname);
@@ -449,7 +438,6 @@ void read_parameter_file(char *fname)
         All.GravitySoftening = param_get_double(ps, "GravitySoftening");
         All.GravitySofteningGas = param_get_double(ps, "GravitySofteningGas");
 
-        locals.ImportBufferBoost = param_get_double(ps, "ImportBufferBoost");
         All.PartAllocFactor = param_get_double(ps, "PartAllocFactor");
         All.TopNodeAllocFactor = param_get_double(ps, "TopNodeAllocFactor");
         All.SlotsIncreaseFactor = param_get_double(ps, "SlotsIncreaseFactor");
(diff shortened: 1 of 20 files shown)


gadget/params.c
```diff
@@ -321,6 +338,19 @@ create_gadget_parameter_set()
     return ps;
 }
 
+/* Structure to hold local parameter values so they can easily be broadcast.
+ * These are for submodules where we do the initialization right here and so don't want to use All.*/
```
rainwoodman: It is a bit strange initializing modules in params.c. params.c should be just parameter parsing I think. 

I think you were trying to identify per-module items in All, and split them out. If the goal is to modulize All, then probably directly do that instead?

libgadget/sfr_eff.c
```diff
@@ -324,9 +323,9 @@ cooling_direct(int i) {
     }
 #endif
 
-    struct UVBG uvbg;
-    GetParticleUVBG(i, &uvbg);
-    unew = DoCooling(unew, SPHP(i).Density * All.cf.a3inv, dtime, &uvbg, &ne, SPHP(i).Metallicity);
+    double redshift = 1./All.Time - 1;
+    struct UVBG uvbg = get_particle_UVBG(redshift, P[i].Pos);
```
rainwoodman: The function is no longer using a particle, so rename it to get_UVBG?

libgadget/utils/endrun.c
```diff
@@ -214,6 +214,7 @@ endrun(int where, const char * fmt, ...)
     va_start(va, fmt);
     messagev(1, where, fmt, va);
     va_end(va);
+    exit(1);
```
rainwoodman: This is for?

gadget/params.c
sbird: That's right. I needed somewhere to store the per-module config locations, since I don't have All to do it anymore. I could just declare them as extern in the per-module headers, in retrospect. Then I'd move the initialisation of ImportBufferBoost out of params as well.

libgadget/sfr_eff.c
sbird: get_local_UVBG?

libgadget/utils/endrun.c
sbird: My compiler would warn that some functions, specifically the switch statements in the new cooling code, left variables uninitialised in the default case, because it didn't know that endrun never returned. Putting the exit(1) in marks the function as __attribute__(noreturn) and stops the compiler warning.

sbird: I'll add a comment.

gadget/params.c
sbird: I want to avoid having to do a BCast on multiple structures, and I also want to avoid tons of extern globals. But then I agree I also don't really want to do all the init calls from params.

What about a set_*_params() function in the submodule, called from params? And then the init() is separate and called later?

sbird:
What do you think of the new initialisation stuff? If you like it I might do the same for the other modules and really modularize Allvars (in a separate pull request).

I ran this all the way down to z=1 with dual cooling routines. Exactly 4 calls to DoCooling came out with different answers, all for very cold particles, which is just because the new code treats the 0/0 that emerges when computing photoheating rate for neutral gas slightly differently.

gadget/params.c
rainwoodman: Since paramset is part of utils module, I think it is the lesser evil to have every module depend on the paramset.

The module can have a function to register params, and another function that loads param values (which will broadcast). Then the loaded param values can be a per-module local struct.

This does mean gradually we will move params.c into submodules inside libgadget piece by piece.

sbird: I dunno. That feels a bit like overkill for now. Also it makes it much harder to work out what the parameters of the simulation actually are: right now we just need to open params.c which is easy for the user (although documentation would also be good). 

To be honest, I'm not sure that leaving All intact isn't the best thing to do after all. It at least makes sorting out the multiprocessing simple.

gadget/params.c
```diff
@@ -489,10 +510,31 @@ void read_parameter_file(char *fname)
...
-        All.HeliumHeatThresh = param_get_double(ps, "HeliumHeatThresh");
-        All.HeliumHeatAmp = param_get_double(ps, "HeliumHeatAmp");
-        All.HeliumHeatExp = param_get_double(ps, "HeliumHeatExp");
+
+        param_get_string2(ps, "TreeCoolFile", All.TreeCoolFile);
+        param_get_string2(ps, "UVFluctuationfile", All.UVFluctuationFile);
+        param_get_string2(ps, "MetalCoolFile", All.MetalCoolFile);
+
+        /*Cooling rate network parameters*/
+        locals.coolpar.CMBTemperature = All.CP.CMBTemperature;
```
rainwoodman: What about moving these into `set_cooling_params(ps)`?

sbird: I think I was trying to make a clean separation between "parameters needed for initialisation" and "parameters needed for calculation". 

I think I want the parameters for calculation to be stored per-module, and the parameters for initialisation to continue to be stored in All (or something like it). At the moment one of the things that I don't like about All is that it combines "things for init" with "parameters for calculation" and even worse "parameters that change every timestep". It will take a while, but I'd ultimately like the three things to be separated (ideally the every-timestep parameters are local variables of course).

sbird:
What about that? The parameters are still registered in params.c, but they are parsed out in cooling_rates.c

rainwoodman:
Like a breeze!

gadget/params.c
rainwoodman: Would it make sense to have two levels then? first divide by module, then within a module, by frequency? Because the module decides which parameter is used more frequently.

sbird:
Thanks! Shall I merge when travis goes green? I will run more tests and change the default cooling rates in a separate PR when I have done so.

rainwoodman:
Sure.

PR #254, opened 2019-03-24
Fix Entropy reload

### LINKED ISSUE #251: Restarted sims deviate from uninterrupted.
I haven't used the restart feature much yet, but on a 512**3 hydro sim I ran a quick test earlier, restarting at z=2.25 and evolving to z=2 and below are the power spectra in the sim that had no restart vs the restarted one. The power on large scales agrees well, but in the non-linear regime there appears to be a non-negligible difference between the final outputs. 

mpgadget_restart

I guess this is could be some numerical rounding error that happens after the particle data is saved and reloaded, but not sure why this would result in consistently more non-linear power in the restarted sim, and to such a large degree.

sbird:
This fixes #251 by correctly initialising Entropy when restarting from a snapshot with DENSITY_INDEPENDENT_SPH off. We load it from the InternalEnergy array both with and without the compile flag to reduce regressions in future.

The SFR after a restart is still not exactly the same as not restarting (differs by about 0.1%) but it's close enough. I don't think we can get bit-exact equivalence because the act of forming stars changes the local density. Tested that it does fix #251 . This is an argument for removing the DENSITY_INDEPENDENT_SPH compile flag in the future.

Also a small refactoring of find_timestep that I did when I thought that was the problem.

### DIFF
--- a/libgadget/timestep.c
+++ b/libgadget/timestep.c
@@ -66,7 +66,8 @@ static inttime_t get_timestep_ti(const int p, const inttime_t dti_max);
 static int get_timestep_bin(inttime_t dti);
 static void do_the_short_range_kick(int i, inttime_t tistart, inttime_t tiend);
 static void do_the_long_range_kick(inttime_t tistart, inttime_t tiend);
-static inttime_t get_long_range_timestep_ti(const inttime_t dti_max);
+/* Get the current PM (global) timestep.*/
+static inttime_t get_PM_timestep_ti(inttime_t Ti_Current);
 
 /*Initialise the integer timeline*/
 void
@@ -122,9 +123,11 @@ set_global_time(double newtime) {
     IonizeParams();
 }
 
-/* This function assigns new timesteps to particles and PM */
+/* This function assigns new short-range timesteps to particles.
+ * It will also shrink the PM timestep to the longest short-range timestep.
+ * Returns the minimum timestep found.*/
 int
-find_timesteps(int * MinTimeBin)
+find_timesteps(inttime_t Ti_Current)
 {
     int pa;
     inttime_t dti_min = TIMEBASE;
@@ -132,21 +135,12 @@ find_timesteps(int * MinTimeBin)
     walltime_measure("/Misc");
 
     /*Update the PM timestep size */
-    const int isPM = is_PM_timestep(All.Ti_Current);
+    const int isPM = is_PM_timestep(Ti_Current);
+    inttime_t dti_max = PM.length;
+
     if(isPM) {
-        SyncPoint * next = find_next_sync_point(All.Ti_Current);
-        inttime_t dti_max;
-        if(next == NULL) {
-
-            endrun(0, "Trying to go beyond the last sync point. This happens only at TimeMax \n");
-
-            /* use a unlimited pm step size*/
-            dti_max = TIMEBASE;
-        } else {
-            /* go no more than the next sync point */
-            dti_max = next->ti - PM.Ti_kick;
-        }
-        PM.length = get_long_range_timestep_ti(dti_max);
+        dti_max = get_PM_timestep_ti(Ti_Current);
+        PM.length = dti_max;
         PM.start = PM.Ti_kick;
     }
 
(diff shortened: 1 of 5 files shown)


libgadget/timestep.c
```diff
@@ -122,31 +123,25 @@ set_global_time(double newtime) {
     IonizeParams();
 }
 
-/* This function assigns new timesteps to particles and PM */
+/* This function assigns new short-range timesteps to particles.
+ * If the argument restrictPMStep is true (on a PM step usually),
```
rainwoodman: restrictPMStep?

rainwoodman:
Why can't we save and load Entropy and InternalEnergy separately when there is no density_split_sph?

sbird:
Because trying to load Entropy will break compatibility with existing snapshots. Unless there is a way to express "try loading Entropy if it is there, otherwise fall back to IE"?

libgadget/timestep.c
sbird: Sorry, rebase error.

rainwoodman:
That's fair.

To address the data versioning problem in general, I think we shall add a Header field indicating the schema version -- the version number shall increase when the data schema is no longer compatible -- in which case we will simply die. Then in petaio.c we can hard code the version as a string, and to find the last version that can read the file we can look for that string in git history. 

This results a cleaner code than the 'if this then that' approach. Of course it means old data has to be manually migrated to the new software (or with separately provided python scripts) -- but it is easier to maintain than carrying the migration code with the mainline code base.

sbird:
The problem here though is not that the schema changed. (Entropy and InternalEnergy do also contain the same information for density-entropy SPH, so arguably we would just be wasting disc space writing both). The problem is really that our snapshot schema depends on compile-time flags.

What do you think about trying to fix this? (in a separate PR)? The Black Hole arrays can be written out unconditionally, since if there are no BHs they will just be empty. Then the difficulty will be only EgyWtDensity for the DENSITY_INDEPENDENT_SPH, and we could try and find a solution for that. We already did something similar with the paramfiles for the same reason.

sbird:
Also didn't mean to push that third commit yet.

sbird:
Oh, I see. We would need the schema versioning to make the extra arrays unconditional.

I support adding a schema version, and using it to make our snapshot format better defined. But I don't like the idea of having to rewrite the data formats when the schema changes, since it would make bisecting bugs across a format change almost impossible. I think ideally we would have the snapshot advertise its format number and the code deals with that by sending it to the right reader. I'm not quite sure how to do that though.

rainwoodman:
First, schema versioning is beyond this PR; I think this PR looks fine to me. It doesn't seem to break the pressure-entroppy restart.

You argue the disadvantage of lifting the schema migration out of the main application is losing the ability to bisect cross schema changes. However if we have the script to freely migrate snapshots to any schema version then we can add an extra conversion step in bisection. 

The conversion always happens, it's just maintaining the logic going from any version to the internal data schema in C (if blocks or vtable), or maintaining it in Python as a sequence of script files (data-migration-v1-v2, data-migration-v2-v3, and the reverses) -- I remember seeing similar structured tools in small web-frameworks. 

To minimize the need to add the data-migration step in bisection we'd prefer to change the schema when the code is stable; though this "principle" is unrealistic -- we likely only identify the need to change a schema when something is broken.

sbird:
That's true. Thinking about it: in practice if we are just adding or moving arrays we can I think easily preserve some level of forward compatibility, where schema version N can be read by the code expecting N-1 (it just ignores the new arrays). And the python script wouldn't have to rewrite the whole snapshot on disc either. So that seems fine.

sbird:
Shall I merge?

rainwoodman:
Yes. Merge please.

rainwoodman:
Switching from InternalEnergy for Entropy to Entropy for Entroy would bbreak forward compatibility.

sbird:
Not if InternalEnergy was still there: the old code would just read InternalEnergy, ignoring the now-present Entropy array, and the new code would read the Entropy.

Issue #255, opened 2019-03-25
Add option for primordial running in GenIC


andreufont:
Currently GenIC can handle different amplitude and slope of primordial power, but not running. We need to be able to modify this in the context of Lyman alpha emulators, and I think it should be fairly easy to fix, but before I get hands on adding this feature I'd like to hear from the experts.

PS: I was going to use PrimordialRunning as option in GenIC config files.

Chris-Pedersen:
I'll send a PR with this shortly!

PR #256, opened 2019-03-25
Added PrimordialRunning to MP-GenIC and make_class_power.py

Chris-Pedersen:
As mentioned in #255

### DIFF
--- a/genic/params.c
+++ b/genic/params.c
@@ -58,6 +58,7 @@ create_parameters()
     param_declare_double(ps, "Sigma8", OPTIONAL, -1, "Renormalise Sigma8 to this number if positive");
     param_declare_double(ps, "InputPowerRedshift", OPTIONAL, -1, "Redshift at which the input power is. Power spectrum will be rescaled to the initial redshift. Negative disables rescaling.");
     param_declare_double(ps, "PrimordialIndex", OPTIONAL, 0.971, "Tilting power, ignored for tabulated input.");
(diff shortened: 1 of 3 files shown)


sbird:
I'm not so sure about this - really what it is doing is adjusting the class power spectrum generation, and the GenIC changes are just because GenIC uses the same paramfile. In general I'd prefer to keep make_class_power.py minimal and have people with more complex cosmologies just roll their own power spectra (which I already do in my emulator code). However, this is just one extra parameter and is fairly generic, so I could probably live with it if you are really attached to keeping make_class_power.py.

Yu, what do you think?

rainwoodman:
@Chris-Pedersen, I think we'll at least need to add 'ignored for tabulated data' in the help string.

The problem is we already have the PrimordialIndex argument there to invite new parameters like this. So we shall either drop both, or include both.

Any capacity in the existing code is likely treated by new coming developers as mandatory rather than coincidental, regardless of the original intent.

Chris-Pedersen:
Whatever you guys feel is best! We can keep this on our own branch if that's best.

rainwoodman:
My vote is to add this.

sbird:
Ok, then merge it!

sbird:
but with a fixed help string saying that the parameter is unused and only for make_class_power.py

Chris-Pedersen:
Ok I've modified the parameter string.

sbird:
Looks good merge it from me.

sbird:
Oh wait...Chris does not have the power to merge. I will do it when travis goes green.

PR #257, opened 2019-03-25
follow up #254

rainwoodman:
I think this was the source of my difficult with 254.

I'm OK with using InternalEnergy as the primary thermal state variable. 

But why scattering the conversion in both init.c and petaio when we can do the conversion one next to the other?

### DIFF
--- a/libgadget/init.c
+++ b/libgadget/init.c
@@ -303,7 +303,7 @@ setup_smoothinglengths(int RestartSnapNum)
         #pragma omp parallel for
         for(i = 0; i < PartManager->NumPart; i++) {
             if(P[i].Type == 0) {
-                SPHP(i).Entropy = u_init;
+                SPHP(i).Entropy = GAMMA_MINUS1 * u_init / pow(SPHP(i).EOMDensity/a3 , GAMMA_MINUS1);
             }
         }
     }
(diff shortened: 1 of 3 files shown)


sbird:
Ah! I actually did this first but I wasn't sure about relying on qsort_openmp being stable.

sbird:
If that is ok, then by all means merge this!

rainwoodman:
Why would it rely on qsort_openmp being stable?

sbird:
There is a qsort_openmp done at petaio.c:836 that sorts the IO blocks by type. Since InternalEnergy and Density/EgyWtDensity have the same type, a stable sort will leave them in the same order that they were declared, but an unstable sort need not do this.

rainwoodman:
I see. I just made it stable.

sbird:
That works. Merge it!

rainwoodman:
BTW, do we have a standar location to add attribute accessor function of a struct?

I noticed the SPH_vel functions are defined locally when they needed. I am thinking of converting the EOMDensity macros to an accessor function, which will ease removal of the conditional compilation on the pressure-entropy hydro solver.

sbird:
I don't think we do but that sounds like a good idea. slotsmanager.h could work?

rainwoodman:
Agreed.

PR #258, opened 2019-03-27
Improvements to format conversion

sbird:
This commit fixes format conversion from bigfile to HDF5 for ICs and for non-double precision snapshots.

It also adds the ability to convert from HDF5 to bigfile.

Tested that both G3 and MPG simulations were able to start from a converted snapshot, and that G3 was able to start from a converted IC file.


tools/convert_bigfile_gadget_format3.py
```diff
@@ -0,0 +1,271 @@
...
+    """Maps names between HDF5 and Bigfile."""
+    def __init__(self):
+        #Map of (only changed) block names between HDF5 and bigfile snapshots.
+        self.hdf_to_bigfile_map = { "Coordinates" : "Position",
+                                    "Velocities": "Velocity", "Masses": "Mass",
+                                    "NeutralHydrogenAbundance": "NeutralHydrogenFraction",
+                                    "GFM_Metallicity": "Metallicity",
+                                    "ParticleIDs" : "ID",
+                                  }
+        #This requires python 2.7
```
rainwoodman: ? Python 3?

tools/convert_bigfile_gadget_format3.py
```diff
@@ -0,0 +1,271 @@
...
+    hattr["Flag_Sfr"] = flag_sfr
+    hattr["Flag_Cooling"] = flag_sfr
+    hattr["Flag_StellarAge"] = flag_sfr
+    hattr["Flag_Metals"] = flag_sfr
+    hattr["Flag_Feedback"] = 0
+    hattr["Flag_DoublePrecision"] = (bf["1/Position"].dtype == np.float64)
+    hattr["Flag_IC_Info"] = 0
+    hattr["Redshift"] = 1./battr["Time"] - 1
+    #Pass other keys through unchanged. We whitelist expected keys to avoid confusing Gadget.
+    hdfats = ["MassTable", "Time", "BoxSize", "Omega0", "OmegaLambda", "HubbleParam", "OmegaBaryon", "UnitLength_in_cm", "UnitMass_in_g", "UnitVelocity_in_cm_per_s"]
```
rainwoodman: It may be more verbose, but writing them explicitly may be easier to understand and maintain in such situation.

tools/convert_bigfile_gadget_format3.py
```diff
@@ -0,0 +1,271 @@
...
+            battr[attr] = hattr[attr]
+    #Fall back to default unit system
+    except KeyError:
+        battr["UnitLength_in_cm"] = 3.085678e+21
+        battr["UnitMass_in_g"] = 1.989e43
+        battr["UnitVelocity_in_cm_per_s"] = 100000.
+    #Some flags
+    battr["UsePeculiarVelocity"] = 0
+    #Missing a factor of 1/Hubble, but doesn't matter.
+    battr["RSDFactor"] = 1./hattr["Time"]**2
```
rainwoodman: If we don't have the right value, just set a nan is better than a wrong value; at least it means if the value is used we will crash rather than give wrong answers siliently.

tools/convert_bigfile_gadget_format3.py
```diff
@@ -0,0 +1,271 @@
...
+
+
+def write_big_file(bfname, hdf5name):
+    """Find all the HDF5 files in the snapshot and merge them into a bigfile."""
+    #Find all the HDF5 snapshot set.
+    hdf5_files = glob.glob(hdf5name)
+    if len(hdf5_files) == 0:
+        hdf5_files = glob.glob(hdf5name+".*.hdf5")
+    elif os.path.isdir(hdf5_files[0]):
+        hdf5_files = glob.glob(os.path.join(hdf5name,"*_[0-9][0-9][0-9].*.hdf5"))
```
rainwoodman: Suggestion : sort file names to make sure the ordering is always consistent every time you run this.

tools/convert_bigfile_gadget_format3.py
sbird: Er, yes, indeed. But I think we can assume python 3 nowdays anyway, so I'll just remove the comment (and the from __future__).

tools/convert_bigfile_gadget_format3.py
sbird: Done!

tools/convert_bigfile_gadget_format3.py
sbird: Yes! I always forget the output of glob isn't sorted. Well spotted!

tools/convert_bigfile_gadget_format3.py
sbird: if I write them explicitly I worry I will at some point copy-paste badly and write, eg:
battr["OmegaK"] = hattr["Omega0"]. But I can change it if you really like.

sbird:
Thank you, as always, for the quick review!

tools/convert_bigfile_gadget_format3.py
rainwoodman: Something to keep in mind in next round of refactoring.

tools/convert_bigfile_gadget_format3.py
```diff
@@ -0,0 +1,271 @@
...
+    if len(hdf5_files) == 0:
+        raise IOError("Could not find hdF5 snapshot as %s (.*.hdf5)" % hdf5name)
+    #Sort so we get a consistent answer each time.
+    hdf5_files.sort()
```
rainwoodman: glob.glob returns a list or a generator? may use sorted(....)

sbird: We could, is that preferred?

sbird: I thought they did the same thing, just one is in place and one is not?

rainwoodman: Yes. functional better than statement, when we can afford.

sbird: So you prefer hdf5_files = sorted(hdf5_files)? Really? Just because glob.glob might return a generator? I can change it but I feel that this egg should be broken at the other end! :-)

rainwoodman: sorted builtin function was added in Python 2.4, and it appears it is preferred than list.sort in most of the examples in https://docs.python.org/3/howto/sorting.html#sortinghowto (which could not be done with sort). 

The main difference is sorted is a function, while .sort() behaves like a statement. A function is clearer than a statement because a statement only performs side-effects, like modifying the argument in-place.

sbird: Ok, changed.

sbird:
The test is failing due to a flaky travis build. Will merge anyway.

PR #260, opened 2019-03-29
Modular tree

sbird:
This pull request is a pure cleanup. It removes all the mess of global variables in forcetree.c and replaces them with a single struct. Initially this is a global struct, but in the last commit it is passed around by all the functions using it. There are thus no longer any globals associated with the force tree.

I did this because I was trying to write tests for some of the tree walking routines (gravity and density) and the mess of globals was making it difficult. This PR also fixes runtests, which has been broken for ages because it didn't correctly initialize the tree.

For my next trick: the domain globals.

### DIFF
--- a/gadget/main.c
+++ b/gadget/main.c
@@ -114,7 +114,7 @@ int main(int argc, char **argv)
         case 3:
             begrun(RestartSnapNum);
             /*FoF needs a tree*/
-            struct OctTree Tree = {0};
+            ForceTree Tree = {0};
             force_tree_rebuild(&Tree);
             fof_fof(&Tree);
             force_tree_free(&Tree);
--- a/gadget/main.c
+++ b/gadget/main.c
@@ -114,8 +114,10 @@ int main(int argc, char **argv)
         case 3:
             begrun(RestartSnapNum);
             /*FoF needs a tree*/
-            force_tree_rebuild();
-            fof_fof();
+            struct OctTree Tree = {0};
+            force_tree_rebuild(&Tree);
+            fof_fof(&Tree);
+            force_tree_free(&Tree);
             fof_save_groups(RestartSnapNum);
             fof_finish();
             break;
--- a/libgadget/blackhole.c
+++ b/libgadget/blackhole.c
@@ -140,7 +140,7 @@ static double blackhole_soundspeed(double entropy, double pressure, double rho) 
     return cs;
 }
 
-void blackhole(struct OctTree * tree)
+void blackhole(ForceTree * tree)
 {
     if(!All.BlackHoleOn) return;
     int i;
--- a/libgadget/blackhole.c
+++ b/libgadget/blackhole.c
@@ -140,7 +140,7 @@ static double blackhole_soundspeed(double entropy, double pressure, double rho) 
     return cs;
 }
 
-void blackhole(void)
+void blackhole(struct OctTree * tree)
 {
     if(!All.BlackHoleOn) return;
     int i;
@@ -160,6 +160,7 @@ void blackhole(void)
     tw_accretion->UseNodeList = 1;
     tw_accretion->query_type_elsize = sizeof(TreeWalkQueryBHAccretion);
     tw_accretion->result_type_elsize = sizeof(TreeWalkResultBHAccretion);
+    tw_accretion->tree = tree;
 
     TreeWalk tw_feedback[1] = {{0}};
     tw_feedback->ev_label = "BH_FEEDBACK";
@@ -174,6 +175,7 @@ void blackhole(void)
     tw_feedback->UseNodeList = 1;
     tw_feedback->query_type_elsize = sizeof(TreeWalkQueryBHFeedback);
     tw_feedback->result_type_elsize = sizeof(TreeWalkResultBHFeedback);
+    tw_feedback->tree = tree;
 
     message(0, "Beginning black-hole accretion\n");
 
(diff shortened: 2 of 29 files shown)


sbird:
Tested that this gives identical positions to within fp accuracy and has negligible change to simulation time.

gadget/main.c
```diff
@@ -114,8 +114,10 @@ int main(int argc, char **argv)
         case 3:
             begrun(RestartSnapNum);
             /*FoF needs a tree*/
-            force_tree_rebuild();
-            fof_fof();
+            struct OctTree Tree = {0};
```
rainwoodman: Since the tree is slightly augmented, I suggest the name ForceTree than the more generic OctTree. It also agrees with the function and module names better.

libgadget/blackhole.c
```diff
@@ -160,6 +160,7 @@ void blackhole(void)
     tw_accretion->UseNodeList = 1;
     tw_accretion->query_type_elsize = sizeof(TreeWalkQueryBHAccretion);
     tw_accretion->result_type_elsize = sizeof(TreeWalkResultBHAccretion);
+    tw_accretion->tree = tree;
```
rainwoodman: I suggest the TreeWalk module takes a tree argument rather than setting the tree member of the TreeWalk structure this way we "formally decouple" the walking policy from the tree.

libgadget/forcetree.c
```diff
@@ -27,51 +27,33 @@
...
-force_tree_build_single(const struct TreeBuilder tb, const int npart);
+force_tree_build_single(const struct OctTree tb, const int npart);
 
 /*Next three are not static as tested.*/
 int
-force_tree_create_nodes(const struct TreeBuilder tb, const int npart);
+force_tree_create_nodes(const struct OctTree tb, const int npart);
 
-struct TreeBuilder
+struct OctTree
```
rainwoodman: shall we typedef?

libgadget/forcetree.c
```diff
@@ -81,47 +63,49 @@ force_tree_eh_slots_fork(EIBase * event, void * userdata)
     int parent = ev->parent;
     int child = ev->child;
     int no;
-    no = Nextnode[parent];
-    Nextnode[parent] = child;
-    Nextnode[child] = no;
-    Father[child] = Father[parent];
+    struct OctTree * tt = (struct OctTree * ) userdata;
```
rainwoodman: tt vs tree?

libgadget/forcetree.c
```diff
@@ -1020,7 +1016,7 @@ void force_treeupdate_pseudos(int no, const struct TreeBuilder tb)
  *  out just before the hydrodynamical SPH forces are computed, i.e. after
  *  density().
  */
-void force_update_hmax(int * activeset, int size)
+void force_update_hmax(int * activeset, int size, struct OctTree * tt)
```
rainwoodman: What is the difference between tt, tb, and tree -- which context do you use which?

libgadget/forcetree.h
```diff
@@ -37,64 +37,77 @@ extern struct NODE
...
+/*The node index is an integer with unusual properties:
+ * no = 0..OctTree.firstnode  corresponds to a particle.
+ * no = OctTree.firstnode..OctTree.lastnode corresponds to actual tree nodes,
+ * and is the only memory allocated in OctTree.Nodes_base. After the tree is built this becomes
+ * no = OctTree.firstnode..OctTree.numnodes which is the only allocated memory.
+ * no > OctTree.lastnode means a pseudo particle on another processor*/
+struct OctTree {
+    /*Is 1 if the tree is allocated*/
+    int tree_allocated_flag;
+    /*Index of first internal node. Difference between Nodes and Nodes_base. == MaxPart*/
```
rainwoodman: Are there any members used only during treebuild? If so, good idea to mark them, since it appears you have merged the tree builder interface with the tree.

rainwoodman:
This is a good move. I just have a bunch of style and clarification requests.

gadget/main.c
sbird: For sure! Actually, how about struct force_tree, since we also have struct particle_data?

libgadget/forcetree.c
sbird: My original notation was to use tt or tb when building the tree (tt when it was a pointer to the OctTree, tb when it was an OctTree passed by value) and to use tree outside of forcetree.c when the tree was being used rather than modified. But the distinction between tt and tb got lost when I converted several internal functions to taking pointers.

gadget/main.c
sbird: No, it shall be ForceTree to match TreeWalk

libgadget/forcetree.c
sbird: Done!

libgadget/forcetree.c
sbird: Now tree!

libgadget/forcetree.h
sbird: Done!

libgadget/blackhole.c
sbird: That's going to be kind of a pain because I would have to pass it down as an argument to every function in the TreeWalk module...

rainwoodman:
Nope I think we shall move away from lower case type names. Usually
function names are lower and type names are camel.

Yu

rainwoodman:
I see. We can leave that itch for the future. The tree walk can have an
internal structure during runtime.

Issue #261, opened 2019-03-29
Too many particles for exchange


Chris-Pedersen:
Got the following crash twice today running on 178e6849a0. This was on a very high res lya sim (512^3 with 10Mpc/h box), but I also had the same crash on a more usual sim (256^3, 30Mph/h). Let me know if you'd like full paramfiles & output files!

```
[ 053920.20 ] tree is done.
[ 053920.21 ] SFR: spawned 0 stars, converted 1 gas particles into stars
[ 053920.21 ] Forces computed.
[ 053920.21 ] PM timebin: 2000 dloga = 0.00399083  Max = (0.1)
[ 053920.21 ] HCI: Nothing happened.
[ 053920.56 ] domain decomposition... (presently allocated=950.652 MB)
[ 053921.19 ] Allocated 56.89 MByte for top-level domain structure
[ 053921.19 ] Attempting new domain decomposition policy: TopNodeAllocFactor=0.5, UseglobalSort=1, SubSampleDistance=16 UsePreSort=0
[ 053921.23 ] use of 71.1111 MB of temporary storage for domain decomposition... (presently allocated=1078.67 MB)
[ 053922.05 ] local topTree size before appending=3329
[ 053922.06 ] Final local topTree size = 3673 per segment = 25.5069.
[ 053922.07 ] NTopLeaves= 3214  NTopNodes=3673 (space for 1864136)
[ 053922.24 ] Expected segment cost 4.37452e+10
[ 053922.24 ] Created 144 segments in 1 rounds. Max leaf cost: 0.371135
[ 053922.24 ] Largest load: work=1.12689 particle=1.99995
[ 053922.37 ] Using 5776378752 bytes for exchange.
[ 053922.46 ] iter=0 exchange of 0000052359243 particles
[ 053922.42 ] Task 41: Too many particles for exchange: NumPart=2873075 count_get = 1574943 count_togo=716329 MaxPart=3728270
[ 053922.46 ] Task 0: Could not exchange particles
[ 053922.43 ] Task 1: Could not exchange particles
...
[ 053922.41 ] Task 47: Could not exchange particles
Could not exchange particles
Could not exchange particles
Could not exchange particles
Thread 2 (Thread 0x2ae102a9f700 (LWP 120316)):
#0  0x000000366e0df283 in poll () from /lib64/libc.so.6
#1  0x00002ae0f3d71a8d in ?? () from /usr/lib64/libpsm_infinipath.so.1
[...]

PR #262, opened 2019-03-29
Exchangegcfix

sbird:
Fix Issue #261 by checking for garbage, and a potential issue where domain leaves no space for star formation.

### DIFF
--- a/libgadget/domain.c
+++ b/libgadget/domain.c
@@ -471,10 +471,14 @@ domain_check_memory_bound(const int print_details, int64_t *TopLeafWork, int64_t
         ta_free(list_load);
     }
 
-    if(max_load > PartManager->MaxPart)
+    /*Leave a small number of particles for star formation */
+    double sfrfrac = 1.;
+    if(All.StarformationOn)
+        sfrfrac = 0.95;
+    if(max_load > PartManager->MaxPart * sfrfrac)
     {
         message(0, "desired memory imbalance=%g  (limit=%d, needed=%d)\n",
-                    (max_load * All.PartAllocFactor) / PartManager->MaxPart, PartManager->MaxPart, max_load);
+                    (max_load * All.PartAllocFactor) / PartManager->MaxPart, sfrfrac * PartManager->MaxPart, max_load);
 
         return 1;
     }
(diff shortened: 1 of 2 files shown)


libgadget/domain.c
```diff
@@ -471,10 +471,11 @@ domain_check_memory_bound(const int print_details, int64_t *TopLeafWork, int64_t
         ta_free(list_load);
     }
 
-    if(max_load > PartManager->MaxPart)
+    /*Leave a small number of particles for star formation */
+    if(max_load > PartManager->MaxPart * 0.99)
```
rainwoodman: I think you want to preserve more. Gut feeling is 0.90.

libgadget/exchange.c
```diff
@@ -169,11 +169,19 @@ static int domain_exchange_once(int (*layoutfunc)(int p), ExchangePlan * plan, i
 
     int bad_exh=0;
 
-    /*Check whether the domain exchange will succeed. If not, bail*/
+    /*Check whether the domain exchange will succeed. If not, check for garbage and then check again.*/
     if(PartManager->NumPart + plan->toGetSum.base - plan->toGoSum.base > PartManager->MaxPart){
-        message(1,"Too many particles for exchange: NumPart=%d count_get = %d count_togo=%d MaxPart=%d\n",
+        /* How much garbage have we got? Keep this serial so there is no memory allocation.
```
rainwoodman: Is the idea the garbages will be always be freed so we actually have more free slots?

sbird: That's right - we have to do a gc anyway to get rid of the particles that are leaving this processor, and the garbage gets swept up in that. But hang on, I can do this a better way.

libgadget/domain.c
sbird: That seems a little high - remember that only the star formation mode that forks an extra particle will increase the particle count. So we are talking the SH03 model, not the QuickLya star formation module. A 10% increase in particle count would mean 20% of the gas forming stars in one time step, which seems a lot. How about 0.95?

sbird:
That's a better fix, I think.

sbird:
Since this is confirmed to fix the problem, shall I merge?

PR #263, opened 2019-04-01
Modulardomain

sbird:
This is the equivalent to the ModularTree PR but for domain. Again tested it makes no difference to output.

### DIFF
--- a/libgadget/domain.c
+++ b/libgadget/domain.c
@@ -50,12 +50,7 @@ typedef struct {
     int PreSort; /** PreSort the local particles before subsampling, creating a fair subsample */
 } DomainDecompositionPolicy;
 
-struct topnode_data *TopNodes;
-struct topleaf_data *TopLeaves;
-
 int MaxTopNodes;		/*!< Maximum number of nodes in the top-level tree used for domain decomposition */
-
-int NTopNodes, NTopLeaves;
 
 static void * TopTreeTempMemory;
 
@@ -78,31 +73,29 @@ struct local_particle_data
     int64_t Cost;
 };
 
-struct task_data * Tasks;
-
 static int
 order_by_key(const void *a, const void *b);
 static void
 mp_order_by_key(const void * data, void * radix, void * arg);
 
 static void
-domain_assign_balanced(int64_t * cost);
+domain_assign_balanced(Domain * domain, int64_t * cost);
 
-static void domain_allocate(DomainDecompositionPolicy * policy);
+static void domain_allocate(Domain * domain, DomainDecompositionPolicy * policy);
 
 static int
-domain_check_memory_bound(const int print_details, int64_t *TopLeafWork, int64_t *TopLeafCount);
+domain_check_memory_bound(const int print_details, int64_t *TopLeafWork, int64_t *TopLeafCount, const Domain * domain);
 
-static int domain_attempt_decompose(DomainDecompositionPolicy * policy);
+static int domain_attempt_decompose(Domain * domain, DomainDecompositionPolicy * policy);
 
 static void
-domain_balance(void);
+domain_balance(Domain * domain);
 
 static int domain_determine_global_toptree(DomainDecompositionPolicy * policy, struct local_topnode_data * topTree, int * topTreeSize);
-static void domain_free(void);
+static void domain_free(Domain * domain);
 
 static void
-domain_compute_costs(int64_t *TopLeafWork, int64_t *TopLeafCount);
+domain_compute_costs(int64_t *TopLeafWork, int64_t *TopLeafCount, const Domain * domain);
 
 static void
 domain_toptree_merge(struct local_topnode_data *treeA, struct local_topnode_data *treeB, int noA, int noB, int * treeASize);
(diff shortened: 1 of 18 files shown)


libgadget/domain.c
```diff
@@ -78,31 +73,29 @@ struct local_particle_data
...
-struct task_data * Tasks;
-
 static int
 order_by_key(const void *a, const void *b);
 static void
 mp_order_by_key(const void * data, void * radix, void * arg);
 
 static void
-domain_assign_balanced(int64_t * cost);
+domain_assign_balanced(Domain * domain, int64_t * cost);
```
rainwoodman: The word Domain is confusing here. It can refer to the domain module or a single domain.

Here are three suggestions:

`DomainMaster * dmaster, DomainWizard * dwizard, DomainController * dcontroller`

libgadget/domain.c
```diff
@@ -417,7 +406,7 @@ domain_balance(void)
 }
 
 static int
-domain_check_memory_bound(const int print_details, int64_t *TopLeafWork, int64_t *TopLeafCount)
+domain_check_memory_bound(const int print_details, int64_t *TopLeafWork, int64_t *TopLeafCount, const Domain * domain)
```
rainwoodman: Is there a consistent policy when to put Domain* as the first argument, when to put it as the last?

libgadget/exchange.c
```diff
@@ -44,10 +43,10 @@ typedef struct {
  * exchange particles according to layoutfunc.
  * layoutfunc gives the target task of particle p.
 */
-static int domain_exchange_once(int (*layoutfunc)(int p), ExchangePlan * plan, int do_gc);
-static void domain_build_plan(int (*layoutfunc)(int p), ExchangePlan * plan);
+static int domain_exchange_once(ExchangePlan * plan, int do_gc);
```
rainwoodman: Shall we lift the function point type with a typedef? It feels like the right time to do it now; e.g. ExchangeLayoutFunc?

libgadget/exchange.c
```diff
@@ -110,7 +109,7 @@ int domain_exchange(int (*layoutfunc)(int p), int do_gc) {
 
         /* determine for each rank how many particles have to be shifted to other ranks */
         plan.last = domain_find_iter_space(&plan);
-        domain_build_plan(layoutfunc, &plan);
+        domain_build_plan(layoutfunc, layout_userdata, &plan);
```
rainwoodman: When did plan remember layoutfunc in a plan? It feels like you can store these before the loop rather than storing them at an arbitrary point in the loop.

Is there any other place where we need the layoutfunc without needing a plan?

rainwoodman: I see you simply figured some of the functions does not need the layoutfunc, you are not saving the layoutfunc in the plan object. Would it be a good idea to store it there?

libgadget/forcetree.c
```diff
@@ -48,13 +48,13 @@ static void
 force_treeupdate_pseudos(int no, const ForceTree * tree);
 
 static void
-force_create_node_for_topnode(int no, int topnode, struct NODE * Nodes, int bits, int x, int y, int z, int *nextfree, const int lastnode);
+force_create_node_for_topnode(int no, int topnode, struct NODE * Nodes, const Domain * domain, int bits, int x, int y, int z, int *nextfree, const int lastnode);
```
rainwoodman: If the treebuilder object remembers a pointer to the domain object, then we may be able to get rid of a lot of `Domain`s here.

rainwoodman: Whether if it is sane to the pointer or not depends on if domain object is used in the entire life cycle of the tree or only transiently during the build stage. It smells like it is the full life time -- since we also need the domain object to export ghosts and manage pseudo particles in treewalk.

rainwoodman:
The dependency between modules becomes clearer as the global variables are removed!

libgadget/domain.c
sbird: That is true. DomainWizard sounds too Harry Potter to me. How about DomainDecomp * ddecomp?
Otherwise I like DomainController best.

libgadget/forcetree.c
sbird: So the TopLeaves are used in the entire tree life cycle for that purpose, and I store a pointer to them in the Tree. But the rest of the domain (the TopNodes and the Task data) is only used during build. This function in particular is really using the TopNodes. It might be possible to remove a Domain from the forcetree functions that build pseudodata if we wanted, but I left it in so I didn't at some future point accidentally introduce bugs related to the tree->TopLeaves pointer not being initialised yet.

libgadget/exchange.c
sbird: Done!

libgadget/exchange.c
sbird: That's right. I thought about it and I think I'd like to keep them separate. The layoutfunc creates the plan and doesn't need to be called after it is made.

libgadget/domain.c
sbird: I put it last when it was const and used as input userdata, first when it was being created and built and behaving like a self pointer. I can make it consistently first if you like

sbird:
Yes! Isn't it nice! The domain was already pretty well structured so this one was easy to write.

sbird:
Will merge when travis goes green, thanks for the reviews.

rainwoodman:
I like Controller better than Decomp; but gadget has been using
DomainDecomp as a name, so I think we shall go with that.
But the variable name can be decomp rather than ddecomp?

sbird:
hmm, I guess I merged prematurely. I'll push the rename through another PR if you think it's worthwhile

rainwoodman:
My expected turnaround time is growing longer. I can live with ddecomp! no need another PR.

sbird:
Ok! Will keep it then. Thanks!

PR #265, opened 2019-04-09
Cleanups to glass files and libGenIC globals

sbird:
This cleans up our glass file support - MakeGlass is replaced with two options, MakeGlassCDM and MakeGlassBar for baryon and CDM glass file support respectively. Neutrinos can't be put into a glass (it makes little sense). The GenIC global particle table variable is removed in preparation (for something I am about to try, but it is a cleanup that makes sense on its own).

### DIFF
--- a/genic/main.c
+++ b/genic/main.c
@@ -88,6 +88,8 @@ int main(int argc, char **argv)
   }
 
   /*First compute and write CDM*/
+  int NumPart;
+  struct ic_part_data *ICP;
 
   if(!All2.MakeGlassCDM) {
       NumPart = setup_grid(All2.ProduceGas * shift_dm, All2.Ngrid, &ICP);
--- a/genic/main.c
+++ b/genic/main.c
@@ -90,12 +90,12 @@ int main(int argc, char **argv)
   /*First compute and write CDM*/
 
   if(!All2.MakeGlassCDM) {
-      setup_grid(All2.ProduceGas * shift_dm, All2.Ngrid);
+      NumPart = setup_grid(All2.ProduceGas * shift_dm, All2.Ngrid, &ICP);
   } else {
       setup_glass(0, All2.Ngrid, GLASS_SEED_HASH(All2.Seed));
   }
 
-  displacement_fields(DMType, ICP);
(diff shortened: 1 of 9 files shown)


genic/main.c
```diff
@@ -123,32 +127,30 @@ int main(int argc, char **argv)
...
-  write_particle_data(1, &bf, 0, All2.Ngrid);
-  free_ffts();
+  write_particle_data(1, &bf, 0, All2.Ngrid, ICP, NumPart);
 
   /*Now make the gas if required*/
   if(All2.ProduceGas) {
 
-    if(!All2.MakeGlass) {
-        setup_grid(shift_gas, All2.Ngrid);
+    if(!All2.MakeGlassBar) {
```
rainwoodman: You sure want to go with Bar not Gas here?

sbird: You're right! I want to go with Gas

libgenic/glass.c
```diff
@@ -91,7 +69,13 @@ setup_glass(double shift, int Ngrid, int seed)
 
     gsl_rng_free(rng);
 
+    evolve_glass(seed, ICP, NumPart);
+    return NumPart;
+}
 
+void evolve_glass(int seed, struct ic_part_data * ICP, const int NumPart)
```
rainwoodman: Perhaps mark this as static and remove it from the .h file, since it is not used outside this module.

sbird: I have a feature I'm trying out that uses it in main.c If it doesn't turn out to be useful I'll remove this from the header

rainwoodman:
Looks good. And minor suggestions -- Why Bar not Gas?

libgenic/glass.c
rainwoodman: If we are exposing more functions, then the functions in the module shall be properly prefixed ..

sbird:
Fixed! Thanks! Will merge when green.

libgenic/glass.c
sbird: glass_evolve ?

sbird: Prefixed and some comments added

PR #267, opened 2019-04-23
Fix the neutral fraction output

sbird:
get_neutral_fraction takes physical protons/cm^3, not internal gadget units.

Create new function that takes the internal units and use it.

### DIFF
--- a/libgadget/cooling_rates.h
+++ b/libgadget/cooling_rates.h
@@ -95,6 +95,6 @@ double get_heatingcooling_rate(double density, double ienergy, double helium, do
 density is gas density in protons/cm^3
 Internal energy is in J/kg == 10^-10 ergs/g.
 helium is a mass fraction.*/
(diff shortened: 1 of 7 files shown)


libgadget/cooling_rates.h
```diff
@@ -91,4 +91,10 @@ double get_ne_by_nh(double density, double ienergy, double helium, const struct
  */
 double get_heatingcooling_rate(double density, double ienergy, double helium, double redshift, double metallicity, const struct UVBG * uvbg, double * ne_equilib);
 
+/*Get the neutral hydrogen fraction at a given temperature and density.
+density is gas density in protons/cm^3
+Internal energy is in J/kg == 10^-10 ergs/g.
+helium is a mass fraction.*/
+double get_neutral_fraction(double density, double ienergy, double helium, const struct UVBG * uvbg, double * ne_init);
```
rainwoodman: I strongly suggest rename the function(s) or function arguments to indicate it consumes local / proper physical quantities.

sbird: Yes! Done.

PR #268, opened 2019-04-23
Yetmoreglass

sbird:
Some changes to the glass file generation to better implement matching linear growth functions.

### DIFF
--- a/genic/main.c
+++ b/genic/main.c
@@ -35,6 +35,7 @@ int main(int argc, char **argv)
   walltime_init(&All.CT);
 
   int64_t TotNumPart = (int64_t) All2.Ngrid*All2.Ngrid*All2.Ngrid;
+  int64_t TotNumPartGas = (int64_t) All2.ProduceGas*All2.NgridGas*All2.NgridGas*All2.NgridGas;
 
   init_cosmology(&All.CP, All.TimeIC);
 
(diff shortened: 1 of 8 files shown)


genic/main.c
```diff
@@ -88,15 +89,35 @@ int main(int argc, char **argv)
   }
 
   /*First compute and write CDM*/
+  double mass[6] = {0};
+  /*Can neglect neutrinos since this only matters for the glass force.*/
+  compute_mass(mass, TotNumPart, TotNumPartGas, 0, 0);
   /*Not used*/
   int size[3], offset[3];
   int NumPart = get_size_offset(size, offset, All2.Ngrid);
```
rainwoodman: Currently NumPart in GenIC and NumPart in Gadget have different meanings. You may want to replace NumPart\> with NumPartCDM globally to avoid confusion.

genic/main.c
```diff
@@ -88,15 +89,35 @@ int main(int argc, char **argv)
...
+  /*Make the table for the baryons if we need, using the second half of the memory.*/
+  if(All2.ProduceGas) {
+    if(!All2.MakeGlassGas) {
+        setup_grid(shift_gas, All2.NgridGas, mass[0], NumPart, ICP+NumPart);
+    } else {
+        setup_glass(0, All2.NgridGas, GLASS_SEED_HASH(All2.Seed + 1), mass[0], NumPart, ICP+NumPart);
+    }
+    /*Do a single glass evolution timestep to avoid close pairs*/
+    if(All2.MakeGlassGas || All2.MakeGlassCDM)
+        glass_evolve(14, 0xDEADBEEF, ICP, NumPart+NumPartGas);
```
rainwoodman: Be aware, this says 14 steps, not a single step. 

Is it feasible to split ICP to ICPCDM, ICPGas and ICPNu? It may look leaner than passing around both the base pointer and the offset.

sbird: I will fix the comment (it is a relic). 

I prefer not to split ICP, I think, to make it explicit that it is the same allocation and thus one can run the glass on the combined particle list.

sbird:
Both fixed! Will merge when green, thanks.

PR #269, opened 2019-04-24
Sfrneutralfix

sbird:
This fixes a very old bug in Gadget: that the neutral fraction on the effective equation of state is reported as
GetNeutralFraction(egyeff) (which is usually << 1) when it should be GetNeutralFraction(egycold) * cold cloud fraction + GetNeutralFraction(egyhot) * (1- cold cloud fraction) (which is usually ~ 1). I fix this now because Jamie Bolton got caught by it in Illustris.

This normally only affects the NH0 saved in the snapshots for star-forming particles, and I would not even ask for a review if that were all. However, the second commit changes the BH_OPTHIN AGN feedback by reweighting the feedback energy more towards dense particles (in that method feedback is weighted by neutral fraction). Do we use that method anywhere? Does that seem plausible? It is certainly more physical, but AGN feedback is not based on physics anyway so it might have negative consequences.

### DIFF
--- a/libgadget/blackhole.c
+++ b/libgadget/blackhole.c
@@ -461,7 +461,7 @@ blackhole_accretion_ngbiter(TreeWalkQueryBHAccretion * I,
             double mass_j;
             if(HAS(All.BlackHoleFeedbackMethod, BH_FEEDBACK_OPTTHIN)) {
                 double redshift = 1./All.Time - 1;
-                double nh0 = get_neutral_fraction_sfr(other, redshift);
+                double nh0 = get_neutral_fraction_sfreff(other, redshift);
                 if(r2 > 0)
                     O->FeedbackWeightSum += (P[other].Mass * nh0) / r2;
             } else {
--- a/libgadget/blackhole.c
+++ b/libgadget/blackhole.c
@@ -16,6 +16,7 @@
 #include "blackhole.h"
 #include "timestep.h"
 #include "hydra.h"
+#include "sfr_eff.h"
 /*! \file blackhole.c
  *  \brief routines for gas accretion onto black holes, and black hole mergers
  */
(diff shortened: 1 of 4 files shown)


libgadget/sfr_eff.c
```diff
@@ -370,6 +370,34 @@ sfreff_on_eeqos(int i)
     return flag;
 }
 
+/*Get the neutral fraction of a particle correctly, accounting for being on the star-forming equation of state*/
+double get_neutral_fraction_sfr(int i, double redshift)
```
rainwoodman: sfreff_get_neutral_fraction or get_neutral_fraction_sfreff?

libgadget/sfr_eff.c
```diff
@@ -529,26 +558,30 @@ starformation(int i, double *localsfr, double * sum_sm)
 
 double get_starformation_rate(int i) {
     /* returns SFR in internal units */
-    return get_starformation_rate_full(i, 0, NULL, NULL, NULL);
+    return get_starformation_rate_full(i, 0, NULL, NULL, NULL, NULL);
 }
 
-static double get_starformation_rate_full(int i, double dtime, MyFloat * ne_new, double * trelax, double * egyeff) {
+static double get_starformation_rate_full(int i, double dtime, MyFloat * ne_new, double * trelax, double * egyhot_i, double * coolfrac) {
```
rainwoodman: egyhot_out, or rename the local variable to avoid a name conflict?

Also, coolfrac really is cloud_fraction, if you look at line 610.

libgadget/blackhole.c
```diff
@@ -460,10 +461,7 @@ blackhole_accretion_ngbiter(TreeWalkQueryBHAccretion * I,
             double mass_j;
             if(HAS(All.BlackHoleFeedbackMethod, BH_FEEDBACK_OPTTHIN)) {
                 double redshift = 1./All.Time - 1;
-                double InternalEnergy = DMAX(All.MinEgySpec, SPHP(other).Entropy / GAMMA_MINUS1 * pow(SPHP(other).EOMDensity * All.cf.a3inv, GAMMA_MINUS1));
-                double physdens = SPHP(other).Density * All.cf.a3inv;
-                struct UVBG uvbg = get_local_UVBG(redshift, P[other].Pos);
-                double nh0 = GetNeutralFraction(InternalEnergy, physdens, &uvbg, SPHP(other).Ne);
+                double nh0 = get_neutral_fraction_sfr(other, redshift);
```
rainwoodman: This is fine. I don't think we use this in production. It was a very crude extreme case scenario that we do not think is of reality.

rainwoodman:
The OPTTHIN branch is not the canonical AGN model, thus it doesn't matter much if the behavior changes drastically. I left a few styling comments as usual.

sbird:
Ok, good! I will fix the style comments and merge.

PR #272, opened 2019-04-26
Change cooling defaults to new rates.

sbird:
The new rates are the ones used in Sherwood and recommended by Jamie Bolton. They are more accurate. Also enable the self-shielding correction by default, which has a slight effect at IGM densities.

This will change the output, albeit slightly. The mean flux changes by a few percent and the flux power by a similar amount (once rescaled to the same mean flux). The effect would be smaller if the temperature were also marginalized out.

@andreufont  @Chris-Pedersen : This is a flag day. Be warned!

### DIFF
--- a/gadget/params.c
+++ b/gadget/params.c
@@ -206,9 +206,9 @@ create_gadget_parameter_set()
         {"Badnell06", Badnell06},
         {NULL, Cen92 },
     };
-    param_declare_enum(ps, "CoolingRates", CoolingTypeTable, OPTIONAL, "KWH92", "Which cooling rate table to use. Options are KWH92 (old gadget default), Enzo2Nyx and Sherwood (new default).");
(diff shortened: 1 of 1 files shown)


sbird:
Merging soon, if any persons present have any reason why these branches should not be joined together, speak now or forever hold your peace.

andreufont:
Hi @sbird - We have frozen versions of the code that we use for our project, so we are happy to merge if you have tested the new branch and it gives reasonable thermal and ionization histories with the default TREECOOL file.

sbird:
I have and it does: the thermal histories change by a couple of percent but it is known to be a better fit, I think. I would very much like you to either merge or to use the updated rates, so that it is easier to compare all published MP-Gadget simulations.

sbird:
(and to compare them to Nyx or Sherwood)

PR #273, opened 2019-04-28
Localparams

sbird:
This is a cleanup branch which makes domain and tree global parameters local to each module.

### DIFF
--- a/libgadget/domain.c
+++ b/libgadget/domain.c
@@ -115,17 +115,17 @@ static void
 mp_order_by_key(const void * data, void * radix, void * arg);
 
 static void
-domain_assign_balanced(DomainDecomp * ddecomp, int64_t * cost, const int NSegment, const int NTask);
+domain_assign_balanced(DomainDecomp * ddecomp, int64_t * cost, const int NsegmentPerTask);
 
-static void domain_allocate(DomainDecomp * ddecomp, DomainDecompositionPolicy * policy, const int NTask);
+static void domain_allocate(DomainDecomp * ddecomp, DomainDecompositionPolicy * policy);
 
 static int
-domain_check_memory_bound(const DomainDecomp * ddecomp, const int print_details, int64_t *TopLeafWork, int64_t *TopLeafCount, const int NTask);
+domain_check_memory_bound(const DomainDecomp * ddecomp, const int print_details, int64_t *TopLeafWork, int64_t *TopLeafCount);
 
 static int domain_attempt_decompose(DomainDecomp * ddecomp, DomainDecompositionPolicy * policy);
 
 static void
-domain_balance(DomainDecomp * ddecomp, const int NTask);
+domain_balance(DomainDecomp * ddecomp);
 
 static int domain_determine_global_toptree(DomainDecompositionPolicy * policy, struct local_topnode_data * topTree, int * topTreeSize, MPI_Comm DomainComm);
 static void domain_free(DomainDecomp * ddecomp);
@@ -150,7 +150,7 @@ domain_create_topleaves(DomainDecomp * ddecomp, int no, int * next);
 static int domain_layoutfunc(int n, const void * userdata);
 
 static int
-domain_policies_init(DomainDecompositionPolicy policies[], const int NTask,
+domain_policies_init(DomainDecompositionPolicy policies[],
         const int NincreaseAlloc,
         const int SwitchToGlobal);
 
@@ -165,17 +165,12 @@ void domain_decompose_full(DomainDecomp * ddecomp)
     static DomainDecompositionPolicy policies[16];
     static int Npolicies = 0;
 
-    /*Number of tasks to decompose to*/
-    int NTask;
-    ddecomp->DomainComm = MPI_COMM_WORLD;
-    MPI_Comm_size(ddecomp->DomainComm, &NTask);
-
     /* start from last successful policy to avoid retries */
     static int LastSuccessfulPolicy = 0;
 
     if (Npolicies == 0) {
         const int NincreaseAlloc = 16;
-        Npolicies = domain_policies_init(policies, NTask, NincreaseAlloc, 8);
+        Npolicies = domain_policies_init(policies, NincreaseAlloc, 8);
     }
 
     double t0, t1;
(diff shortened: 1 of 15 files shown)


libgadget/domain.c
```diff
@@ -129,12 +164,16 @@ void domain_decompose_full(DomainDecomp * ddecomp)
     static DomainDecompositionPolicy policies[16];
     static int Npolicies = 0;
 
+    /*Number of tasks to decompose to*/
+    int NTask;
+    MPI_Comm_size(MPI_COMM_WORLD, &NTask);
```
rainwoodman: Can we save NTask, ThisTask and MPI_Comm as mebers of ddecomp?

Most of the functions require consistent values of these variables.

libgadget/domain.c
```diff
@@ -377,22 +418,22 @@ domain_balance(DomainDecomp * ddecomp)
     walltime_measure("/Domain/Decompose/Sumcost");
 
     /* first try work balance */
-    domain_assign_balanced(ddecomp, TopLeafWork);
+    domain_assign_balanced(ddecomp, TopLeafWork, NTask, NTask);
```
rainwoodman: This looks strange. It looks like you wanted NSubDomains or NSegment here.

libgadget/domain.c
```diff
@@ -377,22 +418,22 @@ domain_balance(DomainDecomp * ddecomp)
...
-    int status = domain_check_memory_bound(ddecomp, 0, TopLeafWork, TopLeafCount);
+    int status = domain_check_memory_bound(ddecomp, 0, TopLeafWork, TopLeafCount, NTask);
     walltime_measure("/Domain/Decompose/memorybound");
 
     if(status != 0)		/* the optimum balanced solution violates memory constraint, let's try something different */
     {
         message(0, "Note: the domain decomposition is suboptimum because the ceiling for memory-imbalance is reached\n");
 
-        domain_assign_balanced(ddecomp, TopLeafCount);
+        domain_assign_balanced(ddecomp, TopLeafCount, NTask, NTask);
```
rainwoodman: This looks strange too. It looks like you wanted NSubDomains or NSegment here.

libgadget/domain.c
rainwoodman: Another possible convention is that any function that needs MPI, pass a MPI_Comm. Getting ThisTask and NTask from MPI_Comm is fast, though we may want a helper macro to reduce the code clutter.

libgadget/forcetree.c
```diff
@@ -126,11 +153,11 @@ ForceTree force_tree_build(int npart, DomainDecomp * ddecomp)
         {
             force_tree_free(&tree);
 
-            message(0, "TreeAllocFactor from %g to %g\n", All.TreeAllocFactor, All.TreeAllocFactor*1.15);
+            message(0, "TreeAllocFactor from %g to %g\n", ForceTreeParams.TreeAllocFactor, ForceTreeParams.TreeAllocFactor*1.15);
 
-            All.TreeAllocFactor *= 1.15;
+            ForceTreeParams.TreeAllocFactor *= 1.15;
```
rainwoodman: This param is mutable. We may want to add a comment to indicate this -- usually 'param's are immutable.

libgadget/forcetree.h
```diff
@@ -76,13 +76,16 @@ typedef struct ForceTree {
...
+void init_forcetree_params(const int FastParticleType, const double * GravitySofteningTable);
+
 int force_tree_allocated(const ForceTree * tt);
 
 /* This function propagates changed SPH smoothing lengths up the tree*/
 void force_update_hmax(int * activeset, int size, ForceTree * tt);
 
-/*This is the main constructor for the tree structure. Pass in something empty.*/
-void force_tree_rebuild(ForceTree * tree, DomainDecomp * ddecomp);
+/*This is the main constructor for the tree structure. Pass in something empty for tree.*/
```
rainwoodman: `tree shall be either zero-filled (due to force_tree_allocated), or a valid ForceTree.`

libgadget/domain.c
```diff
@@ -48,6 +65,7 @@ typedef struct {
     int UseGlobalSort; /** Apply a global sorting on the subsamples before building the top level tree. */
     int SubSampleDistance; /** Frequency of subsampling */
     int PreSort; /** PreSort the local particles before subsampling, creating a fair subsample */
+    int NSubDomains; /** Number of sub-domain segments to create before balancing. Should be DomainOverDecompositionFactor * NTask*/
```
rainwoodman: I avoided using SubDomain because I thought it was confusing -- a more suitable candidate for the name is TopLeave, which is a octtree nodes augumented with workload accounting, and the spatial region element iterated by ForceTree / PM.

Segments are transient objects created only during the load balancing stage.

rainwoodman: `Number of Hilbert-Peano segments` rather than sub-domain segments.

sbird: I will change the comment - I like SubDomain better than TopLeave. I always found TopLeave confusing because it sounds like it is part of a Tree, but it is not.

libgadget/domain.c
sbird: This is a good idea. It would also let us hypothetically use MPI communicators for some future domain feature.

libgadget/domain.c
sbird: Ah but it is NTopLeaves in ddecomp. I will make it NTopLeaves in policy also.

libgadget/domain.c
sbird: NSubDomains is used only at an earlier stage of the process, to build more TopLeaves than we need Segments. The assign_balanced code then multiple TopLeaves to Segments so that each Segment is of roughly equal weight, and then assigns these Segments to Tasks so that the Tasks are also of roughly equal weight.

The thing is, in the current algorithm the Segments are superfluous: because each Segment is of equal weight anyway there is no point making them separate from Tasks. This is specifying that a Segment and a Task are the same. I could simplify by removing the code for multiple Segments per Task but I don't want to yet because the (finely-tuned) algorithm may still turn out to be useful.

sbird: The new code has this clarified.

sbird:
Making the functions pass around a communicator was a good idea! It makes the code much more conceptually easy.

libgadget/domain.c
rainwoodman: Sorry but I think you got this wrong. In general NSegment <= NTopLeaves. Several neighbouring Leaves are combined into a segment during load balancing.

libgadget/exchange.c
```diff
@@ -45,10 +47,10 @@ typedef struct {
  * exchange particles according to layoutfunc.
  * layoutfunc gives the target task of particle p.
 */
-static int domain_exchange_once(ExchangePlan * plan, int do_gc);
+static int domain_exchange_once(ExchangePlan * plan, int do_gc, MPI_Comm ExchangeComm);
```
rainwoodman: Probably just call it comm, since the routine takes exactly one comm and it is for exchange. `MPI_Bcast` doesn't take a BcastComm, for example. It can reduce effort rewriting future code following this style.

libgadget/domain.c
sbird: No, I think this variable is supposed to be NTopLeaves. NSegment is still there in domain_assign_balanced, and is == NTask for now,

libgadget/exchange.c
sbird: Fixed!

PR #274, opened 2019-04-28
Barriers

sbird:
Remove MPI_Barriers, including the one in a collective message().

The last time I tried this the code got a factor of 3 slower - I never worked out why and when I tried again the slowdown has disappeared. Removing the barriers now makes the code a small (~1%) amount faster.

So let's just remove them.

### DIFF
--- a/libgadget/fof.c
+++ b/libgadget/fof.c
@@ -130,7 +130,6 @@ static MPI_Datatype MPI_TYPE_GROUP;
 void fof_fof(ForceTree * tree)
 {
     int i;
-    double t0, t1;
 
     MPI_Type_contiguous(sizeof(Group[0]), MPI_BYTE, &MPI_TYPE_GROUP);
     MPI_Type_commit(&MPI_TYPE_GROUP);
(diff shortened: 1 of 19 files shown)


libgadget/fof.c
```diff
@@ -150,26 +149,20 @@ void fof_fof(ForceTree * tree)
         HaloLabel[i].Pindex = i;
     }
     /* Fill FOFP_List of primary */
-    t0 = second();
     fof_label_primary(tree);
-    t1 = second();
 
-    message(0, "group finding took = %g sec\n", timediff(t0, t1));
+    message(0, "Group finding done.\n");
```
rainwoodman: I think the main reason it were there was because if the code crashes inside a step we never see the wall-clock report of that step. Perhaps we shall add a signal handler that saves the current per rank wall-clock report at crash?

libgadget/utils/endrun.c
```diff
@@ -189,9 +189,6 @@ messagev(int crash, int where, const char * fmt, va_list va)
         sprintf(prefix, "[ %09.2f ] ", MPI_Wtime() - _timestart);
     }
 
-    if(!crash && where <= 0)
-        MPI_Barrier(MPI_COMM_WORLD);
```
rainwoodman: The main purpose is to assert that if we see this message, then all ranks have proceeded to this point. 

I think this is still a useful assertion to have, but I agree we may have abused it.

What about not having the barrier in message(0), but adding a MPI_Barrier() before any message(0) when it is clearly an assertion of code progress?

rainwoodman:
When we enter an long running / fragile subroutine we want to ensure all ranks are consistent.

I think having a clear picture of the running code is worthy of a 1% of performance. We can keep a few explicit barriers entering and leaving long running subroutines, and remove the barrier in message(0). This would remove a vast number of barriers, and yet preserve the clear picture.

libgadget/fof.c
sbird: We do still know how long the group finding took, because we see in the output log a timestamp for each message and we can do sums. If we see:

[10.01] Starting FoF
[10000.5] Group finding done

then we know what that means.

libgadget/utils/endrun.c
sbird: That makes sense - I think the problem is that message(0) means two separate things: 1) Assert all ranks are here. 2) Output this information only on Rank 0.

sbird:
There we go. A few Barriers before long-running subroutines so that the "X is done" message shows clear progress. I rebased on top of the localparams pull so I could use the DomainComm for that barrier (which shows an advantage of removing the Barrier in endrun) so that merge should go in first.

rainwoodman:
Your call when to merge these two.

I have some idea converting those barriers to something fancier, e.g. giving a warning if waited too long, and if mismatched barrier is found, die with a message. I will put this on my next week's todo list.

sbird:
That's pretty cool! I will give these a quick regression test and then merge.

PR #275, opened 2019-04-29
Localwindparams

sbird:
Do the same localisation of parameters from domain and forcetree to winds, sfr and blackhole.

This was pretty easy. The only tricky part was moving the wind-related code scattered around the place into the winds module.

### DIFF
--- a/gadget/params.c
+++ b/gadget/params.c
@@ -188,7 +188,7 @@ create_gadget_parameter_set()
     param_declare_double(ps, "HydroCostFactor", OPTIONAL, 1, "Cost factor of hydro calculation, default to 1.");
 
     param_declare_int(ps, "BytesPerFile", OPTIONAL, 1024 * 1024 * 1024, "number of bytes per file");
-    param_declare_int(ps, "NumWriters", OPTIONAL, NTask, "Max number of concurrent writer processes. 0 implies Number of Tasks; ");
+    param_declare_int(ps, "NumWriters", OPTIONAL, 0, "Max number of concurrent writer processes. 0 implies Number of Tasks; ");
     param_declare_int(ps, "MinNumWriters", OPTIONAL, 1, "Min number of concurrent writer processes. We increase number of Files to avoid too few writers. ");
     param_declare_int(ps, "WritersPerFile", OPTIONAL, 8, "Number of Writer groups assigned to a file; total number of writers is capped by NumWriters.");
 
@@ -423,6 +423,8 @@ void read_parameter_file(char *fname)
         All.IO.BytesPerFile = param_get_int(ps, "BytesPerFile");
         All.IO.UsePeculiarVelocity = 0; /* Will be set by the Initial Condition File */
         All.IO.NumWriters = param_get_int(ps, "NumWriters");
+        if(All.IO.NumWriters == 0)
+            MPI_Comm_size(MPI_COMM_WORLD, &All.IO.NumWriters);
         All.IO.MinNumWriters = param_get_int(ps, "MinNumWriters");
         All.IO.WritersPerFile = param_get_int(ps, "WritersPerFile");
         All.IO.AggregatedIOThreshold = param_get_int(ps, "AggregatedIOThreshold");
--- a/gadget/params.c
+++ b/gadget/params.c
@@ -11,6 +11,7 @@
 #include <libgadget/cooling_rates.h>
 #include <libgadget/winds.h>
 #include <libgadget/sfr_eff.h>
+#include <libgadget/blackhole.h>
 
 /* Optional parameters are passed the flag 0 and required parameters 1.
  * These macros are just to document the semantic meaning of these flags. */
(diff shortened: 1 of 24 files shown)


sbird:
We still have dependencies on allvars because we need unit systems and the Time parameters.

sbird:
And now the Fof module uses local variables! Maybe someday soon I will write the tests this PR makes possible :-)

gadget/params.c
```diff
@@ -9,6 +9,10 @@
 #include <libgadget/utils.h>
 #include <libgadget/treewalk.h>
 #include <libgadget/cooling_rates.h>
+#include <libgadget/winds.h>
+#include <libgadget/sfr_eff.h>
+#include <libgadget/blackhole.h>
+#include <libgadget/fof.h>
```
rainwoodman: yikes. This is how it should looks like. Real header files!

gadget/params.c
```diff
@@ -419,6 +423,8 @@ void read_parameter_file(char *fname)
         All.IO.BytesPerFile = param_get_int(ps, "BytesPerFile");
         All.IO.UsePeculiarVelocity = 0; /* Will be set by the Initial Condition File */
         All.IO.NumWriters = param_get_int(ps, "NumWriters");
+        if(All.IO.NumWriters == 0)
```
rainwoodman: We shall add an `unset` state to the parameters. not urgent, but I will put this on my todo list.

libgadget/allvars.c
```diff
@@ -22,7 +22,6 @@
 
 
 int ThisTask;			/*!< the number of the local processor  */
```
rainwoodman: It looks weird NTask is gone but ThisTask is still here. I guess it still used in too many places?

rainwoodman: Perhaps add a TODO / FIXME about planning to refactor it away in the future.

libgadget/allvars.h
```diff
@@ -176,6 +144,11 @@ extern struct global_data_all_processes
     /* some SPH parameters */
 
     int DesNumNgb;		/*!< Desired number of SPH neighbours */
+    /*Look like black hole parameters but really for neighbour finding*/
```
rainwoodman: Blackhole parameters used in neighbour finding? Also could you add a space between * and the text?

libgadget/allvars.c
sbird: It is not that many places but I ran out of steam...we shall get rid of it at some point.

gadget/params.c
sbird: Yup!

libgadget/fof.c
```diff
@@ -172,23 +200,23 @@ void fof_fof(ForceTree * tree)
     /* We create the smaller 'BaseGroup' data set for this. */
     struct BaseGroup * base = (struct BaseGroup *) mymalloc("BaseGroup", sizeof(struct BaseGroup) * NgroupsExt);
 
-    fof_compile_base(base);
+    fof_compile_base(base, Comm);
 
-    MPI_Barrier(MPI_COMM_WORLD);
+    MPI_Barrier(Comm);
```
rainwoodman: MPI always used lower case 'Comm', btw. I can do this in a formatting PR later.

libgadget/allvars.h
sbird: Yes - neighbour finding for black holes is different for SPH particles. These parameters belong in the density module.

libgadget/fof.c
```diff
@@ -823,6 +858,8 @@ static void fof_reduce_groups(
     MPI_Type_contiguous(elsize, MPI_BYTE, &dtype);
     MPI_Type_commit(&dtype);
 
+    /*Set global data for the comparison*/
+    MinIDCompareThisTask = ThisTask;
```
rainwoodman: Let's name this `_fof_compare_Group_MinIDTask_ThisTask`, to indicate it shall only be used in that function.

libgadget/init.c
```diff
@@ -93,7 +93,7 @@ void init(int RestartSnapNum, DomainDecomp * ddecomp)
 #ifdef BLACK_HOLES
         if(RestartSnapNum == -1 && P[i].Type == 5 )
         {
-            BHP(i).Mass = All.SeedBlackHoleMass;
+            BHP(i).Mass = P[i].Mass;
```
rainwoodman: This is the initial condition and we are setting the initial mass of the blackholes to the seed accretion mass -- your change would have set it directly to the gravity mass. I think this is reasonable.

rainwoodman: Could you add a comment `/*  Watchout: Gadget=3 sets this to the seed mass. */

libgadget/blackhole.c
```diff
@@ -141,7 +180,8 @@ static double blackhole_soundspeed(double entropy, double pressure, double rho)
     return cs;
 }
 
-void blackhole(ForceTree * tree)
+void
+blackhole(ForceTree * tree, double * TimeNextSeedingCheck)
```
rainwoodman: Wht about spliting to two arguments: (double TimeSeedingCheck, double * TimeNextSeedingCheck)? Then the first one is a const and the second one is out-only.

libgadget/init.c
sbird: Right - we never actually do this, because we never have BHs in the ICs. I considered just removing it entirely and saying that the person putting BHs in the ICs should specify a mass for them, but decided against it. Will add a comment.

libgadget/blackhole.c
sbird: I tried that first, with blackhole returning the time of the next seeding check but it looked sort of ugly. I'll add a comment.

sbird:
Thanks for the quick review! I will double check with a test and merge when travis goes green.

libgadget/init.c
rainwoodman: We probably don't even read BHP(i).Mass from IC files..

sbird: Ah yes probably true.

sbird:
Testing caught a couple bugs that sneaked into the wind module. Now bit identical on a pure MPI run to z=9. Very small floating point roundoff change to one of the particles by z = 8.

rainwoodman:
Cool!

sbird:
Merrrrging!

PR #281, opened 2019-05-06
fix a corruption in sfr_eff when MyFloat is not double

rainwoodman:


sbird:
Oops, this is because I used a temporary double in get_neutral_fraction_sfreff ?

sbird:
It isn't actually necessary to set ne_new on entry: it is never read. get_starformation_rate_full reads SPHP(i).Ne into a temporary double itself.

rainwoodman:
Indeed. I fixed another warning related to this as well. Hope it didn't affect any runs? (if MyFloat is set to float it would trigger undefined values)

sbird:
I don't think so. I haven't made a run with MyFloat == float for a while. Thanks for the fix

sbird:
(just to be explicit: feel free to merge!)

PR #282, opened 2019-05-06
More code towards using the UNDEFINED state of optional parameters.

rainwoodman:
- trace the origin of a parameter value (line number of default value)
- warn if a undefined optional is cast to a concrete value.

The third state is not yet used anywhere. A few cases I tried it only makes the situation harder to deal with.

I think we may want to get rid of most of the duplicated items in All / xxx params and ParameterSet at some point, the modules can cache the concrete values before entering expensive loops, and read from the paramset in other occasions; but that's more work than I can do this week.

### DIFF
--- a/gadget/params.c
+++ b/gadget/params.c
@@ -14,11 +14,6 @@
 #include <libgadget/blackhole.h>
 #include <libgadget/fof.h>
 
-/* Optional parameters are passed the flag 0 and required parameters 1.
- * These macros are just to document the semantic meaning of these flags. */
-#define OPTIONAL 0
-#define REQUIRED 1
-
 static int
 BlackHoleFeedbackMethodAction (ParameterSet * ps, char * name, void * data)
 {
--- a/genic/params.c
+++ b/genic/params.c
@@ -8,9 +8,6 @@
 #include <libgadget/physconst.h>
 #include <libgadget/utils.h>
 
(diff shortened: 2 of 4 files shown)


sbird:
This makes sense and the code look fine but we should find a use for it. You said you tried to use it for some of the parameters, but it made things worse? Why was that? It seems a natural fit for Nmesh and PhysDensThresh.

Removing duplicates between ParamSet and All might be a good idea (although now I think we free the paramset once we're done with it, so it is not really duplicated) but we should finish the modularization first, I think.

rainwoodman:
We defer the use of the value too late to the code, which means we need to maintain the tri-state all the way down to the very internal of the implementation in 'All.', in the case of PhysCritDensity. 

If we can broadcast a paramset, then most of all and xxx_params can go away, as we can use the paramset api to query the paramset on demand in almost all cases. It would also mean dividing up params.c into submodules becomes more reasonable. Each module will have a preample section that defines the parameters and their docstrings. A global variable points to the paramset where we can make the queries, and a few static variables that points to cached parsed parameter values -- which can be packed into a module context argument to all module functions and initiated during module initialization.

sbird:
I understand. That sounds like a good plan to work towards, and would remove a bunch of globals. I'm happy for this to be merged.

Issue #278, opened 2019-04-30
Smarter MPI_Barrier


rainwoodman:
Replace `MPI_Barrier` with something that calls `MPI_Iallgather`, such that we can detect mismatch and issue warnings on extremely load imbalance.

PR #283, opened 2019-05-06
Fancy barrier.

rainwoodman:
This implements the fancier barrier in #278.

It currently will issue a warning if the root ranks waits for 5 seconds. 

Some refactoring of the endrun library was done as well.

### DIFF
--- a/libgadget/utils/system.c
+++ b/libgadget/utils/system.c
@@ -543,7 +543,7 @@ get_physmem_bytes()
 }
 
 /**
- * A facny MPI barrier (use MPIU_Barrier macro)
+ * A facny MPI barrier (use MPIU_Barrier macro) mainly for -DDEBUG
  *
  *  - aborts if barrier mismatch occurs
  *  - warn if some ranks are very imbalanced.
@@ -557,34 +557,32 @@ _MPIU_Barrier(const char * fn, const int line, MPI_Comm comm)
     MPI_Comm_size(comm, &NTask);
     MPI_Comm_rank(comm, &ThisTask);
     int * recvbuf = ta_malloc("tags", int, NTask);
-    MPI_Request request;
     int tag = 0;
     int i;
-    for(i = 0; i < strlen(fn); i ++) {
+    for(i = 0; fn[i]; i ++) {
         tag += (int)fn[i] * 8;
     }
     tag += line;
 
+    MPI_Request request;
     MPI_Igather(&tag, 1, MPI_INT, recvbuf, 1, MPI_INT, 0, comm, &request);
-// #ifdef DEBUG TODO if this is a performance drag, change it to debug only
     i = 0;
     int flag = 1;
     int tsleep = 0;
     while(flag) {
         MPI_Test(&request, &flag, MPI_STATUS_IGNORE);
-        usleep(i * 20);
-        tsleep += i * 20;
+        if(flag) break;
+        usleep(i * 1000);
+        tsleep += i * 1000;
         i = i + 1;
         if(i == 50) {
             if(ThisTask == 0) {
-                MPIU_Trace(comm, 0, "Waited more than %g seconds during barrier %s : %d \n", tsleep / 1000., fn, line);
+                MPIU_Trace(comm, 0, "Waited more than %g seconds during barrier %s : %d \n", tsleep / 1000000., fn, line);
             }
             break;
         }
     }
-// #endif
     MPI_Wait(&request, MPI_STATUS_IGNORE);
-
     /* now check if all ranks indeed hit the same barrier. Some MPIs do allow them to mix up! */
     if (ThisTask == 0) {
         for(i = 0; i < NTask; i ++) {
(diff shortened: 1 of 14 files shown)


libgadget/utils/system.h
```diff
@@ -74,4 +74,9 @@ static inline int atomic_add_and_fetch(int * ptr, int value) {
     return k;
 }
 
+void MPIU_trace(MPI_Comm comm, int where, const char * fmt, ...);
+void MPIU_tracev(MPI_Comm comm, int where, const char * fmt, va_list va);
```
sbird: These need some comments. Also, do we need to expose two functions for this?

libgadget/utils/system.c
```diff
@@ -453,6 +528,50 @@ get_physmem_bytes()
     return 64 * 1024 * 1024;
 }
 
+int
```
sbird: Need a comment explaining what this does.

libgadget/utils/system.c
```diff
@@ -453,6 +528,50 @@ get_physmem_bytes()
...
+_MPIU_Barrier(const char * fn, const int line, MPI_Comm comm)
+{
+    int ThisTask, NTask;
+    MPI_Comm_size(comm, &NTask);
+    MPI_Comm_rank(comm, &ThisTask);
+    int * recvbuf = malloc(sizeof(int) * NTask);
```
sbird: system malloc? Should this be ta_malloc?

libgadget/utils/system.c
```diff
@@ -453,6 +528,50 @@ get_physmem_bytes()
...
+    int i;
+    for(i = 0; i < strlen(fn); i ++) {
+        tag += (int)fn[i] * 8;
+    }
+    tag += line;
+
+    MPI_Igather(&tag, 1, MPI_INT, recvbuf, 1, MPI_INT, 0, comm, &request);
+// #ifdef DEBUG TODO if this is a performance drag, change it to debug only
+    i = 0;
+    while(flag) {
```
sbird: This code never runs? Did you mean to initialise flag = 1 above?

libgadget/utils/system.c
```diff
@@ -453,6 +528,50 @@ get_physmem_bytes()
...
+        MPI_Test(&request, &flag, MPI_STATUS_IGNORE);
+        usleep(50);
```
sbird: Can we make the performance hit less by gradating the usleep? eg usleep(MIN(80, 2 * i))? Or similar?

libgadget/utils/system.c
rainwoodman: No it shouldn't because  system.c sits lower on the stack

libgadget/utils/system.c
rainwoodman: Yes!

libgadget/utils/system.c
sbird: We use mymalloc and ta_malloc elsewhere in system.c

libgadget/utils/system.h
rainwoodman: Yes. endrun calls the v version. just like sprintf / sprintfv.

rainwoodman: I added docstring in the C file.

libgadget/utils/system.c
rainwoodman: I am surprised. We probably wanted to minimize fragmentation since these routines are called frequently.  I changed it to use ta_malloc.

sbird: Yes, I think that was it. Thanks.

sbird:
The code looks good. Can you check the performance implications of the hang check? If you are too busy I run some tests here. It might also be worth shrinking the threshold for emitting the message(): 5 s of imbalance in a single timestep is a *lot*; maybe we could report after 1 s. But this needs a test to see how often these things happen in practice.

sbird:
hmm, I ran a test and this loses us 25% of performance, which is too much. Probably the hang check, I will see what happens if I hide it behind DEBUG

sbird:
Also usleep takes an argument in microseconds, not milliseconds, so the warning is printing after 24.5 microseconds, which is often!

sbird:
Even without the hang check we lose 6% performance, which seems a little high too.

sbird:
Part of this may be my running on a shared memory machine where MPI_Barrier can probably be implemented very very quickly. But it may also indicate that MPI_Barrier is so optimized that it isn't worth trying to replace it.

rainwoodman:
Could be the async gather call that's much slower than Barrier.

But I am still surprised that we really call the Barrier function that frequently to cause such a huge difference. We can revert to the fast MPI_Barrier if not in DEBUG mode.

sbird:
I agree it is weird. I think it is because we call MPI_Barrier every timestep, and the short timesteps are sometimes really short. So anything that has a constant cost per timestep can become important. The extra cost could indeed be the Igather, with the associated memory allocation.

sbird:
Can you also fix the spacing of the usleeps? Warning on 24.5 microseconds is too often. Maybe the gradient of the increase should be a little longer too.

sbird:
With the latest code I cannot reproduce any of the slowdown. I suspect that it was just due to the hang check being overly aggressive and thus printing something every time there was a barrier. Or maybe it was random noise from something else. Anyway, I committed a re-enablement of the fancy barrier and I approve merging it at any time.

rainwoodman:
I will merge it then. Thanks!

PR #286, opened 2019-05-09
Runtimesph

sbird:
This removes the DENSITY_INDEPENDENT_SPH compile time option and replaces it with a runtime option. Costs us two floats in sph_particle_data if it is enabled, but that seems worth it to me.

Lightly tested - it compiles and runs but there might be some bugs I have yet to find. I would like a review first.

One transition issue I had is this causing all config files for Lyman-alpha runs to suddenly switch to density independent SPH. So I have used the newly merged third state of the parameters to set an undefined DensityIndependentSph to zero iff kernel == cubic.

### DIFF
--- a/libgadget/density.c
+++ b/libgadget/density.c
@@ -36,10 +36,11 @@ typedef struct
 
 typedef struct {
     TreeWalkResultBase base;
-#ifdef DENSITY_INDEPENDENT_SPH
+
+    /*These are only used for density independent SPH*/
     MyFloat EgyRho;
     MyFloat DhsmlEgyDensity;
-#endif
+
     MyFloat Rho;
     MyFloat DhsmlDensity;
     MyFloat Ngb;
@@ -284,10 +285,11 @@ density_reduce(int place, TreeWalkResultDensity * remote, enum TreeWalkReduceMod
         TREEWALK_REDUCE(SPHP(place).GradRho[2], remote->GradRho[2]);
 #endif
 
-#ifdef DENSITY_INDEPENDENT_SPH
-        TREEWALK_REDUCE(SPHP(place).EgyWtDensity, remote->EgyRho);
-        TREEWALK_REDUCE(SPHP(place).DhsmlEgyDensityFactor, remote->DhsmlEgyDensity);
-#endif
+        /*Only used for density independent SPH*/
+        if(All.DensityIndependentSphOn) {
+            TREEWALK_REDUCE(SPHP(place).EgyWtDensity, remote->EgyRho);
+            TREEWALK_REDUCE(SPHP(place).DhsmlEgyDensityFactor, remote->DhsmlEgyDensity);
+        }
     }
 
 }
@@ -352,13 +354,15 @@ density_ngbiter(
 
         /* Hinv is here because O->DhsmlDensity is drho / dH.
          * nothing to worry here */
-        O->DhsmlDensity += mass_j * density_kernel_dW(&iter->kernel, u, wk, dwk);
+        double density_dW = density_kernel_dW(&iter->kernel, u, wk, dwk);
+        O->DhsmlDensity += mass_j * density_dW;
 
-#ifdef DENSITY_INDEPENDENT_SPH
-        const double EntPred = SPHP(other).EntVarPred;
-        O->EgyRho += mass_j * EntPred * wk;
-        O->DhsmlEgyDensity += mass_j * EntPred * density_kernel_dW(&iter->kernel, u, wk, dwk);
-#endif
+        if(All.DensityIndependentSphOn) {
+            const double EntPred = SPHP(other).EntVarPred;
+            O->EgyRho += mass_j * EntPred * wk;
+            O->DhsmlEgyDensity += mass_j * EntPred * density_dW;
+        }
+
 
 #ifdef SPH_GRAD_RHO
         if(r > 0)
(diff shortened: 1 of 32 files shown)


rainwoodman:
On the last commit.

I feel the default of DensityKernelType shall be derived from DensityIndependentSphOn rather than the other way around as you have now. 

If you worry about silently breaking runs, then you can make it a required parameter for a few months, till the code propagates to everywhere and all files are updated, then mark it optional again.

sbird:
You mean set DensityKernelType = cubic if DensityIndependentSphOn = 0 and DensityKernelType = quintic if DensityIndependentSphOn = 1 ?

We can make them derive from each other so that setting one of them sets the other. Would that work?

sbird:
Like that?

sbird:
I could also just make it a required parameter, but if I do that I think I should make it permanently required. Many of the other master switches are REQUIRED.

rainwoodman:
Yes. I agree required on most 'xxxOn' makes more sense.. I will go over the
rest of the PR tomorrow or sunday.

- Yu

sbird:
Thanks, take your time. I have another branch that removes the compile-time BLACK_HOLES but I wanted to check you agree with the approach for this one before I posted it.

libgadget/density.c
```diff
@@ -284,10 +285,11 @@ density_reduce(int place, TreeWalkResultDensity * remote, enum TreeWalkReduceMod
         TREEWALK_REDUCE(SPHP(place).GradRho[2], remote->GradRho[2]);
 #endif
 
-#ifdef DENSITY_INDEPENDENT_SPH
-        TREEWALK_REDUCE(SPHP(place).EgyWtDensity, remote->EgyRho);
-        TREEWALK_REDUCE(SPHP(place).DhsmlEgyDensityFactor, remote->DhsmlEgyDensity);
-#endif
+        /*Only used for density independent SPH*/
+        if(All.DensityIndependentSphOn) {
```
rainwoodman: Shall we zero out the values if it is not On? We haven't been doing that on other branches, but I do worry about undefined values tainting valgrind reports very late down stream.

libgadget/hydra.c
```diff
@@ -281,32 +296,34 @@ hydro_ngbiter(
...
+                    /* apply the limit if it is enabled > 0*/
+                    r1 = DMIN(r1, All.DensityContrastLimit);
+                    r2 = DMIN(r2, All.DensityContrastLimit);
+                }
             }
         }
-#endif
-        /* grad-h corrections: enabled if DENSITY_INDEPENDENT_SPH is off, or DensityConstrastLimit >= 0 */
+
+        /* grad-h corrections: enabled if DensityIndependentSphOn = 0 is off, or DensityConstrastLimit >= 0 */
```
rainwoodman: Typo = 0.

rainwoodman: DensityContractLimit caught my eye. It controls three behaviours if and only if DensityIndepdentSPH is enabled:
- > 0 Limit r1, and r2, enable the grad-h term
- = 0 no limit r1 and r2, enable the grad-h term
- < 0 disable the grad-h term.

Could you also update the doc string of this in params.c?

libgadget/init.c
```diff
@@ -265,50 +264,51 @@ setup_smoothinglengths(int RestartSnapNum, DomainDecomp * ddecomp)
 
         u_init /= molecular_weight;
 
-#ifdef DENSITY_INDEPENDENT_SPH
-        /* initialization of the entropy variable is a little trickier in this version of SPH,
-           since we need to make sure it 'talks to' the density appropriately */
-        message(0, "Converting u -> entropy, with density split sph\n");
+        if(All.DensityIndependentSphOn) {
```
rainwoodman: Does it make sense to move this as another function? The processing is quite long for a if branch.

libgadget/hydra.c
```diff
@@ -23,26 +23,41 @@
  *  (via artificial viscosity) is computed.
  */
 
+MyFloat SPH_EOMDensity(int i)
```
rainwoodman: Brilliant solution! This is where I got stuck last few attempts.

rainwoodman:
Looks good. I spotted a mis-documented DensityContrastLimit parameter; and maybe a slight refactor of the large section in initialization; but I have the impression it may not be easy to lift it out as a function.

libgadget/density.c
sbird: hmm, I think not. EgyWtDensity is initialised to -1 in init.c and should never be written to, so it is already a canary. Both DhsmlDensityFactor variables should be explicitly initialized to zero though.

sbird: Both these variables are already initialized, but in a non-obvious way.

libgadget/hydra.c
sbird: Done!

libgadget/init.c
sbird: Done!

libgadget/hydra.c
sbird: Thanks! I iterated on the name a few times

sbird:
Done! Thanks for the review. I will test and merge.

sbird:
Confirmed that the results with and without this branch are the same to within floating point accuracy. We lose 3% of performance on a Lyman-alpha forest run with DENSITY_INDEPENDENT_SPH off. I think this is mostly because of the increased size of the structures, rather than the function calls, as it happens in the gravity tree (and we are memory bandwidth limited). I think this is a reasonable price to pay.

PR #287, opened 2019-05-13
Runtimebh

sbird:
Remove the BLACK_HOLES runtime option. Overall cost is one float in sph_particle_data.

The BlackHoleOn parameter was a little strange before: if BLACK_HOLES was compiled but BlackHoleOn = 0 then existing type 5 particles would in places be treated as black holes. This would never really happen but I fixed it anyway, as there might be a performance hit for non black hole runs otherwise.

### DIFF
--- a/libgadget/blackhole.c
+++ b/libgadget/blackhole.c
@@ -225,9 +225,19 @@ blackhole(ForceTree * tree, double * TimeNextSeedingCheck)
 
     /* Let's determine which particles may be swalled and calculate total feedback weights */
     SPH_SwallowID = mymalloc("SPH_SwallowID", SlotsManager->info[0].size * sizeof(MyIDType));
-    #pragma omp parallel for
-    for(i = 0; i < SlotsManager->info[0].size; i ++)
-        SPH_SwallowID[i] = -1;
+    if(ActiveParticle) {
+        #pragma omp parallel for
+        for(i = 0; i < NumActiveParticle; i ++) {
+            int p_i = ActiveParticle[i];
+            SPH_SwallowID[P[p_i].PI] = -1;
+        }
+    }
+    else {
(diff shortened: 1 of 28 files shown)


sbird:
Tested that this makes no difference to the output with the hydro example. It marginally speeds up the no-black-hole case if -DBLACK_HOLES is defined, and has a 3% performance hit compared to the case where it is not defined.

sbird:
Once this series of PRs removing the last of the compile-time options is in, I think I will look into shrinking sph_particle_data!

sbird:
I spoke too soon: it has a 3% performance hit over the first few timesteps and a 1% performance *increase* over the first 30!

sbird:
I think the performance changes I am measuring are really just random noise caused by running on whatever firefox is doing at the time. This PR passes tests and I think can be merged straightforwardly after review.

libgadget/blackhole.c
```diff
@@ -224,6 +224,20 @@ blackhole(ForceTree * tree, double * TimeNextSeedingCheck)
     N_sph_swallowed = N_BH_swallowed = 0;
 
     /* Let's determine which particles may be swalled and calculate total feedback weights */
+    SPH_SwallowID = mymalloc("SPH_SwallowID", SlotsManager->info[0].size * sizeof(MyIDType));
```
rainwoodman: Is this a bug fix? Seems to be related to Yueying's Ptype=5 Mass not tracking BlackholeMass issue.

sbird: I don't think it is a bug fix - it is just allocating this memory within blackhole.c instead of in sph_particle_data

sbird: It might be that there is a bug there, but this does not change it as far as I can see.

PR #288, opened 2019-05-14
Cleanups

sbird:
Cleanups of the remaining random compile-time options, delete a couple of obsolete files, and a couple of fixes for compilation on stampede.


depends/Makefile
```diff
@@ -17,7 +18,7 @@ lib/libbigfile-mpi.a: bigfile/src/bigfile-mpi.c
 	cd bigfile/src; \
 	make install PREFIX=$(PWD) CC="$(MPICC)" MPICC="$(MPICC)" CFLAGS="$(OPTIMIZE)"
 
-lib/libpfft_omp.a:
+pfft:
```
rainwoodman: Ah. Good catch.

libgadget/cooling_rates.c
```diff
@@ -174,12 +184,12 @@ load_treecool(const char * TreeCoolFile)
                 continue;
             Gamma_log1z[i] = atof(retval);
             /*Get the rest*/
-            Gamma_HI.ydata[i] = log10(atof(strtok(NULL, " \t")));
-            Gamma_HeI.ydata[i] = log10(atof(strtok(NULL, " \t")));
-            Gamma_HeII.ydata[i] = log10(atof(strtok(NULL, " \t")));
-            Eps_HI.ydata[i] = log10(atof(strtok(NULL, " \t")));
-            Eps_HeI.ydata[i] = log10(atof(strtok(NULL, " \t")));
-            Eps_HeII.ydata[i] = log10(atof(strtok(NULL, " \t")));
+            Gamma_HI.ydata[i]   = load_tree_value();
```
rainwoodman: Better to use a locally defined macro than hiding strtok, which is a weird function that uses a hidden state. We may forget load_tree_value also has that strange behaviour sooner or later.

sbird: Even better to not use the weird strtok function: the last commit has strtok_r

depends/Makefile
sbird: Thanks! It was really Yueying who found it. Unfortunately I needed an extra commit to tell Make not to keep trying to create this file...

PR #291, opened 2019-05-16
Runtime SPH_GRAD_RHO

sbird:
This removes the compile flag SPH_GRAD_RHO which is the LAST COMPILE TIME SCIENCE FLAG!! Hooray! No longer will we need to recompile MP-Gadget for different tests! Everything is runtime.

If you agree with the approach for sph_extra_data I might move some other things that are recomputed every timestep into it. eg, VelPred.

### DIFF
--- a/libgadget/run.c
+++ b/libgadget/run.c
@@ -211,8 +211,6 @@ void run(DomainDecomp * ddecomp)
             cooling_and_starformation(&Tree);
         }
 
-        slots_free_sph_scratch_data(SphP_scratch);
-
         /* Update velocity to Ti_Current; this synchonizes TiKick and TiDrift for the active particles */
 
         if(is_PM) {
@@ -252,6 +250,8 @@ void run(DomainDecomp * ddecomp)
         }
 
         write_checkpoint(WriteSnapshot, WriteFOF, &Tree);
+
+        slots_free_sph_scratch_data(SphP_scratch);
 
         write_cpu_log(NumCurrentTiStep);		/* produce some CPU usage info */
 
--- a/libgadget/run.c
+++ b/libgadget/run.c
@@ -195,6 +195,8 @@ void run(DomainDecomp * ddecomp)
         ForceTree Tree = {0};
         force_tree_rebuild(&Tree, ddecomp, All.BoxSize, HybridNuGrav);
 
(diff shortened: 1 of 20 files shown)


libgadget/run.c
```diff
@@ -209,6 +211,8 @@ void run(DomainDecomp * ddecomp)
             cooling_and_starformation(&Tree);
         }
 
+        slots_free_sph_extra_data(SphP_extra);
```
rainwoodman: What about calling this sph_scratch_data? 
It is some data that we do not need to preserve them; the word extra doesn't convey this message. However I do have the impression we probably use 'extra' for this meaning everywhere else anyways.

libgadget/sfr_eff.c
```diff
@@ -830,7 +830,14 @@ find_star_mass(int i)
  * You may need a license to run with these modess.
 
  * */
-#if defined SPH_GRAD_RHO
+
+int need_to_compute_sph_grad_rho(void)
+{
+    if (HAS(sfr_params.StarformationCriterion, SFR_CRITERION_MOLECULAR_H2)) {
```
rainwoodman: Better to add an sfr prefix now than later, since it is not static.

libgadget/slotsmanager.h
```diff
@@ -124,6 +128,9 @@ struct sph_particle_data
 #define BHP(i) BhP[P[i].PI]
 #define STARP(i) StarP[P[i].PI]
 
+/*Shortcut for the extra data*/
+#define SphP_extra ((struct sph_extra_data*) SlotsManager->info[0].extradata)
```
rainwoodman: Can we call it something like SphPE?

sbird: SphP_scratch. I don't like names that differ from another name by just one character.

sbird:
Names done! Will test.

libgadget/slotsmanager.h
rainwoodman: I guess I've been parsing too much pure consonant names...

sbird: Ah yes, the vowel makes a difference...I've seen several people confuse DESI and DES, for example.

sbird:
Ok, a few segfaults later, this passes my tests...will merge when travis goes green.

PR #292, opened 2019-05-20
PetaPM as an object.

rainwoodman:
This PR allows reusing PetaPM with different Nmesh in the same application.

Additional changes

- Add IDGenerator interface in libgenic to route the interfaces to PetaPM object.
- neutrino IC uses NGridNu for thermal velocity. (potential BUG)
- stop doing the parallel for in the glass initial smearing. (potential BUG, race condition on RNG)
- split out the PMesh gaussian file routine to a new file, since locally it uses PM* pm to adapt to FastPM's source code.

[x] compiles
[ ] no crash in nbody
[ ] no crash in lya
[ ] sane in nbody
[ ] sane in lya


sbird:
Do you mind if I take this and finish it off? It is a good idea and you've already done most of the work.

rainwoodman:
No objections. Go ahead!

I remember it still crashes in nbody. Let me see if I recall any details.

sbird:
Ok! There is a rebase to do too - I'll let you now how it goes.

sbird:
This version confirmed to produce the same ICs (after the first commit fixing the race) on the travis example in both OMP and MPI modes. I will make a longer test now. Did you mean to remove the DEBUG guards around verify_density_field?

libgadget/powerspectrum.c
```diff
@@ -78,8 +76,8 @@ void powerspectrum_sum(struct _powerspectrum * PowerSpectrum, const double BoxSi
         PowerSpectrum->Power[i] /= PowerSpectrum->Norm;
         PowerSpectrum->kk[i] /= PowerSpectrum->Nmodes[i];
         /* Mpc/h units */
-        PowerSpectrum->kk[i] *= 2 * M_PI / (BoxSize_in_cm / CM_PER_MPC );
-        PowerSpectrum->Power[i] *= pow(BoxSize_in_cm / CM_PER_MPC , 3.0);
+        PowerSpectrum->kk[i] *= 2 * M_PI / (PowerSpectrum->BoxSize_in_cm / CM_PER_MPC );
+        PowerSpectrum->Power[i] *= pow(PowerSpectrum->BoxSize_in_cm / CM_PER_MPC , 3.0);
```
rainwoodman: Perhaps BoxSize_in_MPC is cleaner than cm.

libgadget/powerspectrum.h
```diff
@@ -24,7 +26,7 @@ struct _powerspectrum {
 };
 
 /*Allocate memory for the power spectrum*/
-void powerspectrum_alloc(struct _powerspectrum * PowerSpectrum, const int nbins, const int nthreads, const int MassiveNuLinResp);
+void powerspectrum_alloc(struct _powerspectrum * PowerSpectrum, const int nbins, const int nthreads, const int MassiveNuLinResp, const double BoxSize_in_cm);
```
rainwoodman: powerspectrum_init?

libgadget/petapm.h
```diff
@@ -16,7 +16,46 @@ typedef struct Region {
...
+    int * DcSend;
+    int * DcRecv;
+
+    double * BufSend;
+    double * BufRecv;
+    int * ibuffer;
+};
+
+/* Data which is private to the PetaPM structure. Don't access from outside.*/
+typedef struct PetaPMPriv {
```
rainwoodman: The reason of keeping a Priv is to avoid exposing unnecessary interface to the public file. I think the benefit of keeping a smaller API outweights saving an allocation -- and the obligation to maintain the life cycle of these variables the way it currently is (which is not very well documented.)

Unless there are other reasons a client must see all of the fields. If it is just one or two fields, we can always add the accessor functions.

rainwoodman:
I remember nbody was likely crashing with an memory alloc/unalloc mismatch error. Is it still the case?

libgadget/powerspectrum.c
sbird: Indeed!

libgadget/powerspectrum.h
sbird: It does do allocations.

libgadget/petapm.h
sbird: I agree that a small API is generally desirable, but we have no obligation to keep binary compatibility or do actual data member hiding in this code. We can keep the API small by just not using priv outside of petapm. Whereas the extra allocation is another thing to ensure gets freed in the right order, which is a bit of a pain. So while I agree with you as a general rule of code, in this specific case I think the stack allocation is the cleaner path.

sbird:
No, I fixed it. It was because the random offset in the glassfile was missing a units factor, so the initial random positions of the glassfile had particles spread over 3* the box, instead of 3* a cell.

libgadget/petapm.h
rainwoodman: Sure.

On the other hand, priv can be managed by the standard C malloc. Only `large` and frequently realloced objects (fragment makers) really 'need' to use mymalloc. `priv` is unlikely a fragmenter in this sense.

I think when I was hacking in FastPM, there were several occasions I hoped an object was an opaque pointer. I don't quite recall the reason.  Probably related to forward declarations or embedding structures. I'll bring this up again if I remember the situation.

PR #293, opened 2019-05-21
Add parameter to rescale hydrogen heat amplitude

Chris-Pedersen:
There might be a neater way to implement this, let me know what you think.


sbird:
I'm happy with the implementation, but please fix the tests.

Chris-Pedersen:
Ok will do!

Chris-Pedersen:
Had to take the log out of the loop as it was breaking some tests, probably faster this way. This should go green now.

sbird:
Can you explicitly set CoolingParams.HydrogenHeatAmp = 0 in both test_cooling.c and test_cooling_rates.c? Just to be safe. Then I will merge.

PR #296, opened 2019-05-30
Fixes from today's telecon

sbird:
Here are the fixes from today's telecon for the small black hole seed mass and for the ScaleDepVelocity. There is one more fix to make address sanitizer happy.

### DIFF
--- a/genic/main.c
+++ b/genic/main.c
@@ -110,9 +110,9 @@ int main(int argc, char **argv)
   /*Make the table for the baryons if we need, using the second half of the memory.*/
   if(All2.ProduceGas) {
     if(!All2.MakeGlassGas) {
-        setup_grid(shift_gas, All2.NgridGas, mass[0], NumPartCDM, ICP+NumPartCDM);
+        setup_grid(shift_gas, All2.NgridGas, mass[0], NumPartGas, ICP+NumPartCDM);
     } else {
-        setup_glass(0, All2.NgridGas, GLASS_SEED_HASH(All2.Seed + 1), mass[0], NumPartCDM, ICP+NumPartCDM);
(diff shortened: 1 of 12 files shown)


genic/main.c
```diff
@@ -43,7 +43,7 @@ int main(int argc, char **argv)
 
   petapm_init(All.BoxSize, All.Nmesh, All.NumThreads);
   /*Initialise particle spacings*/
-  const double meanspacing = All.BoxSize / All2.Ngrid;
+  const double meanspacing = All.BoxSize / DMAX(All2.Ngrid, All2.NgridGas);
```
rainwoodman: This looks strange. perhaps need meanspacing for each species instead.

sbird: It is only used for computing the shifts - the important thing is actually that the shift is less than one grid cell for all particles, so this is the right thing to do.

libgadget/fof.c
```diff
@@ -796,7 +796,7 @@ fof_compile_catalogue(struct Group * group, double BoxSize, int BlackHoleInfo, M
     fof_finish_group_properties(Group, BoxSize);
 
     int64_t TotNids;
-    MPI_Allreduce(&Ngroups, &TotNgroups, 1, MPI_UINT64, MPI_SUM, Comm);
+    sumup_large_ints(1, &Ngroups, &TotNgroups);
```
rainwoodman: It's cleaner to use TotNgroups = Ngroups and MPI_IN_PLACE.
Most use cases of sumup_large_ints can be replaced by this.

sbird: Ah yes - that way the right communicator is used. Thanks.

PR #297, opened 2019-06-02
Memoryopts

sbird:
Shrink sph_particle_data and particle_data by making some array elements local to other modules. Move EntVarPred into sph_scratch_data. This provides a modest speed win (10% of hydro) because of the temporary array for Pressure, but the big point is to enable future work. I have checked that these commits make no difference to output of a hydro example run.

### DIFF
--- a/libgadget/density.c
+++ b/libgadget/density.c
@@ -55,6 +55,7 @@ typedef struct {
 
 struct DensityPriv {
     MyFloat *Left, *Right;
+    MyFloat *Rot;
     int NIteration;
     int *NPLeft;
     int update_hsml;
@@ -148,7 +149,7 @@ density_internal(int update_hsml, ForceTree * tree)
 
     DENSITY_GET_PRIV(tw)->Left = (MyFloat *) mymalloc("DENSITY_GET_PRIV(tw)->Left", PartManager->NumPart * sizeof(MyFloat));
     DENSITY_GET_PRIV(tw)->Right = (MyFloat *) mymalloc("DENSITY_GET_PRIV(tw)->Right", PartManager->NumPart * sizeof(MyFloat));
-
+    DENSITY_GET_PRIV(tw)->Rot = (MyFloat *) mymalloc("DENSITY_GET_PRIV(tw)->Rot", 3 * SlotsManager->info[0].size * sizeof(MyFloat));
     DENSITY_GET_PRIV(tw)->update_hsml = update_hsml;
 
     DENSITY_GET_PRIV(tw)->NIteration = 0;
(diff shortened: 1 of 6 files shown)


libgadget/density.c
```diff
@@ -55,6 +55,7 @@ typedef struct {
 
 struct DensityPriv {
     MyFloat *Left, *Right;
+    MyFloat *Rot;
```
rainwoodman: MyFloat (*Rot)[3], and sizeof can be  sizeof(priv->Rot[0])
Then you can get get rid of those 3* formality in array indexing (easier to align lines without them.)

libgadget/fofpetaio.c
```diff
@@ -127,8 +127,10 @@ struct PartIndex {
 };
 
 static int fof_sorted_layout(int i, const void * userdata) {
-    return P[i].targettask;
+    int * targetTask = (int *) userdata;
```
rainwoodman: Was it important to carry targetTask with P in exchange? I think not. Did you check this?

libgadget/hydra.c
```diff
@@ -120,14 +139,27 @@ void hydro_force(ForceTree * tree)
...
+    for(i = 0; i < SlotsManager->info[0].size; i++)
+        HYDRA_GET_PRIV(tw)->PressurePred[i] = PressurePred(i);
 
     double timeall = 0, timenetwork = 0;
     double timecomp, timecomm, timewait;
 
     walltime_measure("/Misc");
 
+    /* Initialize some time factors*/
+    fac_mu = pow(All.cf.a, 3 * (GAMMA - 1) / 2) / All.cf.a;
```
rainwoodman: Move them into priv as well?

libgadget/hydra.c
```diff
@@ -273,6 +303,8 @@ hydro_ngbiter(
             if(vsig > O->MaxSignalVel)
                 O->MaxSignalVel = vsig;
 
+            /* Note this uses the CurlVel of an inactive particle, which may not be
```
rainwoodman: DivVel as well? In principle we could have drifted them to current time since the acceleration is known? Did you check if we drift these quantities at all?

libgadget/density.c
rainwoodman: To case malloc result use (MyFloat (*) [3])

libgadget/hydra.c
sbird: Yes,  DivVel as well, I think. As far as I can see we don't drift them. I don't know that we can easily drift them because they depend on the relative velocities between the particle and neighbours. It might be worthwhile implementing a timestep limiter as a fix.

libgadget/fofpetaio.c
sbird: I checked the output was the same and it doesn't seem important from looking at the code. Carrying origintask is important.

libgadget/hydra.c
sbird: Done!

libgadget/density.c
sbird: Done, I think. Like 9181ae3 ?

PR #298, opened 2019-06-02
Fixorder

sbird:
This PR changes output. Please review it quite carefully (after memoryopts is reviewed and merged) because I am not sure it is the right idea.

Commit 62bd178 and f982f91 make a difference to output because they fix a bug - before this commit dloga for EntVarPred and gravkick_factor for VelPred are computed from Ti_drift - Ti_kick, but Ti_drift has not yet been updated to the present timestep, so it is finding the predicted quantity at the drift time from the *last* timestep, not the current timestep. 

These commits move the predicted quantities to sph_scratch_data and their initialization to the init of density. At this time Ti_drift has already been updated, so the predicted quantities are computed for the current timestep.

Commit 2785d89 computes the densities after the gravitational forces. This matches what is done in bluetides-ii and in P-Gadget3. It means that because we do not need the sph_scratch_data allocated
during PM, there is a lot more memory available then.

However, UNLIKE bluetides-ii and P-Gadget3, it means that VelPred is computed using the gravitational acceleration from the current timestep, rather than the last timestep (in bluetides-ii VelPred was computed at kick time). This seems more consistent to me. It does mean that the HydroAccel in VelPred is from a different timestep than the GravAccel, but since GravAccel and HydroAccel are decoupled I think this should be ok and maybe more accurate. If you disagree let me know and I will drop the commit. I am not sure about this.

The place VelPred is actually used is to compute DivVel and CurlVel. These are both a little weird. Although they are computed using the velocity at the current drift time, they are only re-computed for active particles. So when they are used in hydra.c to compute artificial viscosity between two particles, if one of the particles is inactive the DivVel and CurlVel for the viscosity are strictly speaking being evaluated at an earlier time than the current drift. This doesn't make a lot of sense but is (probably) ok because if the velocity were changing significantly the particle would be active. My guess is that since viscosity is kind of wacky in SPH anyway the true fix is to switch to MFM.

I considered trying to compute density() and hydra() after adding the kick, so we wouldn't need velocity prediction at all, but I decided against it because it would make hydroaccel use a different timestepping scheme than gravaccel.

### DIFF
--- a/libgadget/blackhole.c
+++ b/libgadget/blackhole.c
@@ -481,7 +481,7 @@ blackhole_accretion_ngbiter(TreeWalkQueryBHAccretion * I,
             O->Rho += (mass_j * wk);
 
             O->SmoothedPressure += (mass_j * wk * PressurePred(P[other].PI));
-            O->SmoothedEntropy += (mass_j * wk * SPHP(other).Entropy);
+            O->SmoothedEntropy += (mass_j * wk * SphP_scratch->EntVarPred[P[other].PI]);
             O->GasVel[0] += (mass_j * wk * SphP_scratch->VelPred[3 * P[other].PI]);
             O->GasVel[1] += (mass_j * wk * SphP_scratch->VelPred[3 * P[other].PI+1]);
             O->GasVel[2] += (mass_j * wk * SphP_scratch->VelPred[3 * P[other].PI+2]);
--- a/libgadget/blackhole.c
+++ b/libgadget/blackhole.c
@@ -482,9 +482,9 @@ blackhole_accretion_ngbiter(TreeWalkQueryBHAccretion * I,
 
             O->SmoothedPressure += (mass_j * wk * PressurePred(P[other].PI));
             O->SmoothedEntropy += (mass_j * wk * SPHP(other).Entropy);
-            O->GasVel[0] += (mass_j * wk * SPHP(other).VelPred[0]);
-            O->GasVel[1] += (mass_j * wk * SPHP(other).VelPred[1]);
-            O->GasVel[2] += (mass_j * wk * SPHP(other).VelPred[2]);
+            O->GasVel[0] += (mass_j * wk * SphP_scratch->VelPred[3 * P[other].PI]);
+            O->GasVel[1] += (mass_j * wk * SphP_scratch->VelPred[3 * P[other].PI+1]);
+            O->GasVel[2] += (mass_j * wk * SphP_scratch->VelPred[3 * P[other].PI+2]);
 
             /* here we have a gas particle; check for swallowing */
 
--- a/libgadget/density.c
+++ b/libgadget/density.c
@@ -16,6 +16,20 @@
 
 #define MAXITER 400
 
+/* The evolved entropy at drift time: evolved dlog a.
+ * Used to predict pressure and entropy for SPH */
+static MyFloat
+SPH_EntVarPred(int i)
+{
+        double dloga = dloga_from_dti(P[i].Ti_drift - P[i].Ti_kick);
+        double EntVarPred = SPHP(i).Entropy + SPHP(i).DtEntropy * dloga;
+        /*Entropy limiter for the predicted entropy: makes sure entropy stays positive. */
+        if(dloga > 0 && EntVarPred < 0.5*SPHP(i).Entropy)
+            EntVarPred = 0.5 * SPHP(i).Entropy;
+        EntVarPred = pow(EntVarPred, 1/GAMMA);
+        return EntVarPred;
+}
+
 /*! Structure for communication during the density computation. Holds data that is sent to other processors.
 */
 typedef struct {
@@ -159,15 +173,14 @@ density_internal(int update_hsml, ForceTree * tree)
      * The iteration will gradually turn DensityIterationDone on more particles.
      * */
     #pragma omp parallel for
-    for(i = 0; i < NumActiveParticle; i++)
+    for(i = 0; i < PartManager->NumPart; i++)
     {
-        int p_i = i;
-        if(ActiveParticle)
-            p_i = ActiveParticle[i];
-        P[p_i].DensityIterationDone = 0;
-        P[p_i].NumNgb = 0;
-        DENSITY_GET_PRIV(tw)->Left[p_i] = 0;
-        DENSITY_GET_PRIV(tw)->Right[p_i] = All.BoxSize;
+        P[i].DensityIterationDone = 0;
+        P[i].NumNgb = 0;
+        DENSITY_GET_PRIV(tw)->Left[i] = 0;
+        DENSITY_GET_PRIV(tw)->Right[i] = All.BoxSize;
+        if(P[i].Type == 0)
+            SphP_scratch->EntVarPred[P[i].PI] = SPH_EntVarPred(i);
     }
 
     /* allocate buffers to arrange communication */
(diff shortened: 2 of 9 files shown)


rainwoodman:
This is quite confusing to me. The word 'current' and 'previous' is dubious because we are updating the acceleration with kick. Since you have been thinking about this, what about writing down the operators in a more formal way? Sounds like you are using
```
H_v(t_i + 1) ~ H_a(t_i), G_a(t_i); H_f(t_i+1) ~ H_v(t_i +1)
```
But old MP-Gadget is using
```
H_v(t_i + 1) ~ H_a(t_i), G_a(t_i - 1); H_f(t_i+1) ~ H_v(t_i +1)
```
Could it be old gadget is using
```
H_v(t_i) ~ H_a(t_i - 1), G_a(t_i - 1); H_f(t_i) ~ H_v(t_i)
```
but we moved almost everything by 1 by mistake at some point?

sbird:
So there are two separate things going on. This only affects the predicted quantities, which are drift time quantities, not kick time quantities. It does not affect position changes, only the viscosity term in SPH. 

Commits 62bd178 and f982f91 change the predicted quantities from being evaluated at:
H_pv(t_drift - 1) ~ H_a(t_i), G_a(t_i);
to:
H_pv(t_drift) ~ H_a(t_i), G_a(t_i);

These quantities are evaluated by adding the kick-time quantity to the acceleration:
H_pv(t_drift) ~ H_v(t_kick) + H_a(t_i) * F1(t_drift- t_kick) + G_a(t_i) * F2(t_drift- t_kick)

where F1 and F2 are kick factors from the cosmology.

Commit 2785d89 changes the predicted velocity only (because the other predicted quantities feel gravity) from being: 
H_pv(t_drift) ~ H_v(t_kick) + H_a(t_i) * F1(t_drift- t_kick) + G_a(t_i) * F2(t_drift- t_kick)
to
H_pv(t_drift) ~ H_v(t_kick) + H_a(t_i) * F1(t_drift- t_kick) + G_a(t_i+1) * F2(t_drift- t_kick)

Let me double-check what old Gadget does.

sbird:
I checked Gadget-2 because it is the simpler code. Gadget-2 computes VelPred when drifting particles, and uses the post-drift time. So it has:

H_pv(t_drift) ~ H_a(t_i), G_a(t_i);

and commit 177f23f which computes VelPred as above thus fixes a bug (or at least, brings us into agreement with Gadget-2).

Gadget-2 does not have EntVarPred (this came in with DensityEntropy SPH) but it does do something strange with PressurePred: it computes it at the end of the current particle timestep, irrespective of what the drift or kick times are. My copy of Gadget-3 does not do this. It computes EntropyPred and PressurePred at the drift time,  

H_pe(t_drift) ~ H_a(t_i), G_a(t_i);

so 7bf5170 fixes a bug also, I think.

I cannot find anyone who does computes VelPred as 92d28aa does. The paper for the viscosity switch is:
http://adsabs.harvard.edu/abs/1997JCoPh.136...41M
It is behind a paywall so I naturally cannot read it. Notoriously, this artificial viscosity formulation is not very accurate. It is also only important during a shock, when HydroAccel (which is not changed) will dominate. So I suspect it doesn't really matter. But I think it may be worthwhile keeping agreement with P-Gadget 3, and thus I should drop 92d28aa.

sbird:
I will check how much difference that specific commit makes.

sbird:
This whole series makes a moderate but not large difference. If I had enabled threading in my tests it would be lost in the noise. The mean shift in particle positions is 0.5 kpc (because most particles are in the IGM and adiabatic anyway so artificial viscosity is not important). Max shift is ~ 100 kpc in a 4000 kpc box. Star formation rates change by a couple of percent. Commit 92d28aa makes a difference of the order of the rest of the series.

So I think it really does not matter that much.

rainwoodman:
std may be a better measurement of discrenpacy than the mean. 100kpc does
seem to be a bit large as well.

Let me read through your comments more carefully tonight.

Yu

rainwoodman:
What does your t_i / t_drift / t_kick mean? Terms like " G_a(t_i+1) * F2(t_drift- t_kick)" doesn't look right to my eyes. Did you imply t_i is one of t_drift or t_kick?

I dug out Daniel Price's SPH leap frog scheme:

http://users.monash.edu.au/~dprice/ndspmhd/index.html

This is called a predictor/corrector scheme.

[stepND_leapfrog_mhd.txt](https://github.com/MP-Gadget/MP-Gadget/files/3255452/stepND_leapfrog_mhd.txt)

Relevant lines are 
- predictor: 130 131 The predicted quantities shall have the same time stamp as the desired force output
- corrector: 175 Leapfrog is applied to all force terms consistently.

I think in BT-i/ii code I may have messed up the Predictor-Corrector scheme more than what's already there.

If we summarize Price's code, then:

- Predictor takes a particle with whatever t_drift / t_kick and evolve any interested property using the current particle state to t_current. It can be a null operation of t_drift == t_current. 
- ComputeACC always consume predicted quantities (line 152)
- kicks and drifts commute with the predictor, since it is linear.
- PM ACC is assumed a constant between PM steps, thus commute as well.

Is there a way to consolidate?

sbird:
> What does your t_i / t_drift / t_kick mean? Terms like " G_a(t_i+1) * F2(t_drift- t_kick)" doesn't look right to my eyes. Did you imply t_i is one of t_drift or t_kick?

No - the forces are evaluated at t_drift, yes, but the force from the last timestep during which this particle was active is used for prediction (t_drift - 1). This must be the case for hydro force because the predicted quantity is used to compute the force at t_drift. So it is not yet available for prediction. The question is: should we use the now available gravitational force from t_drift for prediction? It seems that no-one else does so I think the answer is probably no.

> I dug out Daniel Price's SPH leap frog scheme:
> 
> http://users.monash.edu.au/~dprice/ndspmhd/index.html
> 
> This is called a predictor/corrector scheme.
> 
> [stepND_leapfrog_mhd.txt](https://github.com/MP-Gadget/MP-Gadget/files/3255452/stepND_leapfrog_mhd.txt)
> 
> Relevant lines are
> 
>     * predictor: 130 131 The predicted quantities shall have the same time stamp as the desired force output
> 
>     * corrector: 175 Leapfrog is applied to all force terms consistently.
> 
> 
> I think in BT-i/ii code I may have messed up the Predictor-Corrector scheme more than what's already there.
> 
> If we summarize Price's code, then:
> 
>     * Predictor takes a particle with whatever t_drift / t_kick and evolve any interested property using the current particle state to t_current. It can be a null operation of t_drift == t_current.
> 
>     * ComputeACC always consume predicted quantities (line 152)
> 
>     * kicks and drifts commute with the predictor, since it is linear.
> 
>     * PM ACC is assumed a constant between PM steps, thus commute as well.
> 
> 
> Is there a way to consolidate?

Yes - if I understand you correctly the predictor-correlator is what is now implemented in this PR. I don't think the bug exists in Bluetides-ii: I think we introduced it more recently than that.

rainwoodman:
Great! Let's merge this then -- it would be nicer if you can add a description of the predictor-correlator somewhere in the code, so we don't drift away from this in the future.

sbird:
Ok, I will merge! 

I'm not sure I understand the predictor/correlator stuff well enough to write the documentation for it. I am really going on "just do what G2/3 does" here. I think I would like to implement MFM at some point. Maybe doing that will help me understand hydro solvers better.

PR #299, opened 2019-06-05
Injectedbhenergy

sbird:
Instead of creating Injected_BH_Energy and adding it in the SFR code,
just add the BH feedback energy directly into blackhole.c

The previous code was doing something complicated: first it collected the total energy injected by the BH in a temporary variable, then it adds it to the total entropy when doing cooling. This means that the BH can never prevent a star forming region from being star forming. I think that this is unphysical: if you hit gas with a quasar it should stop forming stars.

Was there a reason we were doing it the previous way that I have missed?

### DIFF
--- a/libgadget/sfr_eff.c
+++ b/libgadget/sfr_eff.c
@@ -351,28 +351,6 @@ cooling_direct(int i) {
             (SPHP(i).Entropy + SPHP(i).DtEntropy * dloga) /
             GAMMA_MINUS1 * pow(SPH_EOMDensity(i) * All.cf.a3inv, GAMMA_MINUS1));
 
-    if(All.BlackHoleOn && SPHP(i).Injected_BH_Energy)
-    {
-        if(P[i].Mass == 0) {
-            endrun(12, "Encoutered zero mass particle during sfr;"
-                      " We haven't implemented tracer particles and this shall not happen\n");
-            /* This shall not happend */
-            SPHP(i).Injected_BH_Energy = 0;
-        }
-
-        unew += SPHP(i).Injected_BH_Energy / P[i].Mass;
-        const double u_to_temp_fac = (4 / (8 - 5 * (1 - HYDROGEN_MASSFRAC))) * PROTONMASS / BOLTZMANN * GAMMA_MINUS1
-        * All.UnitEnergy_in_cgs / All.UnitMass_in_g;
-
-        double temp = u_to_temp_fac * unew;
-
-
-        if(temp > 5.0e9)
-            unew = 5.0e9 / u_to_temp_fac;
-
-        SPHP(i).Injected_BH_Energy = 0;
-    }
-
     double redshift = 1./All.Time - 1;
     struct UVBG uvbg = get_local_UVBG(redshift, P[i].Pos);
     unew = DoCooling(redshift, unew, SPHP(i).Density * All.cf.a3inv, dtime, &uvbg, &ne, SPHP(i).Metallicity);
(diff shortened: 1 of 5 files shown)


sbird:
Request comments from Tiziana

libgadget/sfr_eff.c
```diff
@@ -476,34 +454,7 @@ static void cooling_relaxed(int i, double egyeff, double dtime, double trelax) {
...
-
-        double temp = u_to_temp_fac * egycurrent;
-
-        if(temp > 5.0e9)
-            egycurrent = 5.0e9 / u_to_temp_fac;
-
-        if(egycurrent > egyeff)
-        {
-            double ne = SPHP(i).Ne;
-            double tcool = GetCoolingTime(redshift, egycurrent, SPHP(i).Density * All.cf.a3inv, &uvbg, &ne, SPHP(i).Metallicity);
```
rainwoodman: trelax is modified here. I don't think you still capture this logic after the removal.. You would compute trelax the regular way from two phase model. Here trelax is set to tcool if the BH makes the particles too hot.  I am not sure about the importance of doing this.

But I also don't quite understand the purpose of having this trelax -- I don't think I found this treatment in the SH03 multi-phase paper. Did you understand this part of the model?

sbird: No, I don't understand this part of the model! It is a bit mysterious. 

Hotter gas generally has a longer cooling time, but trelax is only modified if the new cooling time is *shorter* than the old one. I don't think that would ever happen if the BH heats the gas. So the new trelax I think doesn't do much?

It looks like the code is trying very hard to prevent the black hole feedback from shutting down star formation in star forming regions, but I don't understand why.

rainwoodman: - tcool is compared against treleax, which seems to be mostly come from tsfr. So this is sort of deciding if sfr or BH is dominating the entropy change.

- egycurrent can be so much less than egyeff that after injecting BH feedback it is still less.

- also looks like the injected energy is not conserved, since we only use it in the first step, and then clears it.

Notice that the _eff part of the file name says this is about effective eq of state, so it makes sense the eq of state depends on if the particle is in the BH feedback region, which introduces an explicit coupling with the BH module. 

Is there a better way to model this coupling than passing in this weird injected_BH_energy? Is it possible to combine it with other transient variables you introduced, or use a similar technique to lift it out of SphP?

sbird: tsfr notice is not coming from egyeff, but from the free-fall time based on the density. So the code is finding places where the effective cooling time is shorter than the free-fall time...but the effective equation of state is supposed to avoid that.

I *think* that this part of the model is trying to allow for the black hole feedback to stimulate star formation in already star forming regions. But I could be wrong.

It is absolutely possible to move injected_BH_energy to sph_scratch_data. I actually wrote that code first. I'm just not sure the coupling makes that much sense: upstream doesn't seem to do it.

sbird:
This code dates back to the original svn commit of P-Gadget2 in 2006, and has not been modified since. I will check if the condition is ever actually hit.

sbird:
Looking at the numbers: this happens all the time. In practice we have dtime ~ 1e-3, trelax ~ 1e-3, tcool ~ 1e-5. So we go from egy =  egyeff - (egycurrent - egyeff) * 0.5 to egy =  egyeff - (egycurrent - egyeff) * 1e-42
So indeed black holes just do not heat any particles on the effective equation of state *at all* in the Gadget black hole model.

However, I bet this doesn't make a difference anyway, because I also do not see any instances in which the temperature of the particle is more than 20% above egyeff even with the black hole heating.

sbird:
So a broadly equivalent model would be a patch in blackhole.c reading:
if(!sfreff_eeqos(other))
 add_blackhole_heating();

sbird:
Sorry, it is more than that, because the change in cooling time also affects the energy that was around before the BH hit the particle. So in fact a simpler version of this model is:
cooling_relaxed(other) {
if(SPHP(other).Injected_BH_energy > 0)
SPHP(other).egycurrent = egyeff;
else
SPHP(other).egycurrent = egyeff + (egycurrent - egyeff) * exp(-dtime/trelax); 
}

However, this still makes an incredibly small difference because a simpler version of cooling_relaxed is actually: 
cooling_relaxed(other) {
SPHP(other).egycurrent = egyeff;
}
ie, all star forming particles cool to the effective energy very quickly.

sbird:
So removing this in fact decreases star formation by about 0.5%. Which is a very small effect. I think this PR is a reasonable simplification.

rainwoodman:
The difference is going to be bigger near massive blackholes -- but I think we shall give this a go.

Worst case we shall add something in blackhole.c to reduce the feedback of massive blackholes to compensate this (we do have full access to the gas particle state there). It might be even better than than the pre-PR code, which we couldn't make sense of anyways.

sbird:
If anything I think larger changes would help match observations better; these kinds of models generally expel too much gas to achieve a given star formation rate suppression. I will test again with the full change, rather than just removing the conditional by hand and check that I'm not missing something before merging.

sbird:
Turns out I had missed something: I had inverted the unit conversion! It is fixed now and gives reasonable results on a test run. I also made the locking a bit lighter. I will merge when travis goes green.

sbird:
The final numbers are about a 6% reduction in star formation rate. It does also slow the code down by ~ 10%, apparently because some of the heated particles (which would before have formed stars) have very short timesteps. So I think I will drop this and replace it with the scratch variable version of it, with a comment detailing what the model does.

PR #300, opened 2019-06-06
Use the same velocities in snapshots as in Gadget-2/3 and as in ICs.

sbird:
This introduces a new header flag specifying whether Gadget-formatted
velocities are used, so that downstream analysis codes can tell the
difference. By default the velocities output are the same as those read
in the ICs. If the flag is not present in the ICs then it is assumed
the ICs have the 1/sqrt(a) factor, but the snapshots do not.

### DIFF
--- a/genic/params.c
+++ b/genic/params.c
@@ -142,6 +142,7 @@ void read_parameterfile(char *fname)
 
     /*Simulation parameters*/
     All.IO.UsePeculiarVelocity = param_get_int(ps, "UsePeculiarVelocity");
+    All.IO.UseGadgetVelocity = 1;
     All.BoxSize = param_get_double(ps, "BoxSize");
     All.Nmesh = param_get_int(ps, "Nmesh");
     All2.Ngrid = param_get_int(ps, "Ngrid");
--- a/libgadget/allvars.h
+++ b/libgadget/allvars.h
@@ -116,6 +116,7 @@ extern struct global_data_all_processes
         int EnableAggregatedIO;  /* Enable aggregated IO policy for small files.*/
         size_t AggregatedIOThreshold; /* bytes per writer above which to use non-aggregated IO (avoid OOM)*/
         int UsePeculiarVelocity;
+        int UseGadgetVelocity; /* If this is true, output velocities in the Gadget-2/3 format, which is physical velocity divided by a^{1/2}. This matches the units of MP-GenIC*/
     } IO;
 
     double PartAllocFactor;	/*!< in order to maintain work-load balance, the particle load will usually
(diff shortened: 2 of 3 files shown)


rainwoodman:
I dislike adding a new flag about velocity. I was hoping we make UsePeculiarVelocity the default and get rid of that flag at some point. 
It's easier and less confusing to do the conversion  in the downstream than supporting multiple velocity conventions. 

Try to keep the number of switches in the simulation software minimal, and delegate as much as sensible to downstream.

sbird:
On thinking it over I agree with you. Can you explain why UsePeculiarVelocity = 0 by default? 
It seems to me like the correct usability fix is to set UsePeculiarVelocity = 1 as a default and I would like to know why you didn't do this a while ago.

I think that MP-Gadget uses V = a * dx/dt internally. When UsePeculiarVelocity = 1, the Velocity field is just physical dx/dt in km/s for particles (which is the velocity: the thing limited by 3x10^5 km/s).

When UsePeculiarVelocity = 0, the velocity field is physical dx/dt/sqrt(a) for MP-GenIC, but a * dx/dt for MP-Gadget.

Gadget-2 & 3 uses a * dx/dt internally but outputs dx/dt / sqrt(a) for all snapshots. 

So the UsePeculiarVelocity = 0 case for us is really just weird. It doesn't match past usage and it doesn't match the natural thing the user is expecting.

sbird:
Incidentally, there is code in timestep.c that reads:
        /* Code here imposes a hard limit (default to speed of light)
         * on the gas velocity. Then a limit on the change in entropy
         * FIXME: This should probably not be needed!*/
        const double velfac = sqrt(All.cf.a3inv);
        double vv=0;
        for(j=0; j < 3; j++)
            vv += P[i].Vel[j] * P[i].Vel[j];
        vv = sqrt(vv);

        if(vv > All.MaxGasVel * velfac) {
            for(j=0;j < 3; j++)
            {
                P[i].Vel[j] *= All.MaxGasVel * velfac / vv;
            }
but this is not correct: the internal velocity units are a * dx/dt so velfac should be 1/All.cf.a

I would be completely amazed if this condition were ever hit though! Perhaps we should remove the condition completely, or replace it with a warning or a crash dump.

sbird:
I created https://github.com/MP-Gadget/MP-Gadget/pull/301 which I think does what is needed.

rainwoodman:
MaxGasVel can occur when BH feedback goes crazy -- yes we shall write a message about the velocity cap being hit. If the cap happens too frequently the simulation can be swamped by the message, but at that point we don't want to continue the run anyways.

PR #301, opened 2019-06-10
Make UsePeculiarVelocities = 1 the default.

sbird:
This replaces https://github.com/MP-Gadget/MP-Gadget/pull/300

Makes UsePeculiarVelocities = 1 the default, fixes the format conversion script and adds some comments on the velocity format.

### DIFF
--- a/libgadget/petaio.c
+++ b/libgadget/petaio.c
@@ -481,7 +481,10 @@ petaio_read_header_internal(BigFile * bf) {
         endrun(0, "Failed to read required cosmology from IC header\n");
     }
 
-    /* Fall back to use a**2 * dx/dt if UsePeculiarVelocity is not set in IC */
+    /* If UsePeculiarVelocity = 1 then snapshots save to the velocity field the physical peculiar velocity, v = a dx/dt (where x is comoving distance).
+     * If UsePeculiarVelocity = 0 then the velocity field is a * v = a^2 dx/dt in snapshots
+     * and v / sqrt(a) = sqrt(a) dx/dt in the ICs. Note that snapshots never match Gadget-2, which
+     * saves physical peculiar velocity / sqrt(a) in both ICs and snapshots. */
     All.IO.UsePeculiarVelocity = _get_attr_int(&bh, "UsePeculiarVelocity", 0);
 
     if(0 != big_block_get_attr(&bh, "TotNumPartInit", All.NTotalInit, "u8", 6)) {
--- a/libgadget/timestep.c
+++ b/libgadget/timestep.c
@@ -314,18 +314,17 @@ do_the_short_range_kick(int i, inttime_t tistart, inttime_t tiend)
         }
 
         /* Code here imposes a hard limit (default to speed of light)
-         * on the gas velocity. Then a limit on the change in entropy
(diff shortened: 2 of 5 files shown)


libgadget/timestep.c
```diff
@@ -314,15 +314,15 @@ do_the_short_range_kick(int i, inttime_t tistart, inttime_t tiend)
...
-         * FIXME: This should probably not be needed!*/
-        const double velfac = sqrt(All.cf.a3inv);
+         * on the gas velocity. This should rarely be hit.*/
+        const double velfac = 1./All.cf.a;
         double vv=0;
         for(j=0; j < 3; j++)
             vv += P[i].Vel[j] * P[i].Vel[j];
         vv = sqrt(vv);
 
         if(vv > All.MaxGasVel * velfac) {
```
rainwoodman: Shouldn't this always be vv * ainv > All.MaxGasVel?
Then in parameters, MaxGasVel shall be defined in terms of peculiar velocity, a dx / dt and thus 3e5 in km/s.

libgadget/petaio.c
```diff
@@ -481,7 +481,10 @@ petaio_read_header_internal(BigFile * bf) {
         endrun(0, "Failed to read required cosmology from IC header\n");
     }
 
-    /* Fall back to use a**2 * dx/dt if UsePeculiarVelocity is not set in IC */
+    /* If UsePeculiarVelocity = 1 then snapshots save to the velocity field the physical peculiar velocity.
+     * If UsePeculiarVelocity = 0 then snapshots save a * physical peculiar velocity in snapshots
```
rainwoodman: I settled to explicitly write down whenever these are mentioned the exact formula 
- a dx / dt for 'physical peculiar velocity', 
- a^2 dx / dt for conjugate momentum (per unit mass internal integration quantity, BlueTides-i/ii) and
- a^{1/2} dx / dt for Gadget-1 velocity.

I think the formula makes conversion clear and avoids confusion.

tools/convert_bigfile_gadget_format3.py
```diff
@@ -113,7 +116,14 @@ def write_hdf_file(bf, hdf5name, fnum, nfiles):
                 continue
             ptype = int(ptype)
             hname = names.get_hdf5_name(bname)
-            hdf5["PartType"+str(ptype)][hname] =  bf[block][startpart[ptype]:endpart[ptype]]
+            bfdata = bf[block][startpart[ptype]:endpart[ptype]]
+            #Velocities need units adjusted
```
rainwoodman: # Gadget-3 uses a^{1/2} dx / dt. 

if pecvel: # pecvel is a dx / dt.

else # MP-Gadget@Bluetides snapshot, a^2 dx / dt.

rainwoodman:
Looks good! I request a few changes in comments.

libgadget/petaio.c
sbird: That is a good idea. Just to explicitly check that I have this right: x is comoving distance in the above, right?

libgadget/timestep.c
sbird: Oops, yes it should! Nicely caught.

PR #304, opened 2019-06-24
Fixfofsegfault

sbird:
This PR fixes a segfault in the FOF code on particle write. The segfault was happening because the star formation rate calculation may use SPH.GradRho, and after the exchange in FOF SPH.GradRho may be on a different processor. There are two fixes needed. The first is to compute the neutral fraction without computing the star formation rate and the second is to store the star formation rate in the sph_particle_data struct again so that it goes with the FOF exchange.

Commit 1e47aad implements the first by refactoring the star formation rate code so that the effective equation of state parameters are computed in one function and the actual SFR in another. It does marginally change the output of the hydro run, slightly increasing the number of stars. Looking at sfr.txt it seems that the change is in particles with a small sfr, which have their sfrs scattered slightly and symmetrically with a median of exactly zero. I *think* therefore that this is just the fp optimizations being applied a bit differently, but please review carefully in case I missed something.

### DIFF
--- a/libgadget/fof.c
+++ b/libgadget/fof.c
@@ -579,7 +579,7 @@ static void add_particle_to_group(struct Group * gdst, int i, double BoxSize, in
 
 
     if(P[index].Type == 0) {
-        gdst->Sfr += get_starformation_rate(index);
+        gdst->Sfr += SPHP(index).Sfr;
     }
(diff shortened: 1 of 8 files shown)


sbird:
@rainwoodman ? Yueying has confirmed this fixes a segfault is the extra star formation models are enabled.

PR #309, opened 2019-07-06
Pass some of the errors as return values.

rainwoodman:
Then they can be printed with endrun / message, thus not swallowed
by the super computers at crash.

### DIFF
--- a/libgadget/utils/paramset.c
+++ b/libgadget/utils/paramset.c
@@ -169,7 +169,7 @@ int param_validate(ParameterSet * ps, char **error)
     for(i = 0; i < ps->size; i ++) {
         ParameterSchema * p = &ps->p[i];
         if(p->required == REQUIRED && ps->value[p->index].nil) {
-            char error1 = fastpm_strdup_printf("Parameter `%s` is required, but not set.", p->name);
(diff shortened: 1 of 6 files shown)


rainwoodman:
Hacking session with @cheng-zhao and Albert Chuang.

libgadget/utils/paramset.c
```diff
@@ -103,7 +103,7 @@ static void
 param_set_from_string(ParameterSet * ps, char * name, char * value, int lineno);
 
 
-static int param_emit(ParameterSet * ps, char * start, int size, int lineno)
+static int param_emit(ParameterSet * ps, char * start, int size, int lineno, char **error)
```
sbird: This could use a comment about the allocation lifetime of error

libgadget/utils/paramset.c
```diff
@@ -198,9 +202,16 @@ int param_parse (ParameterSet * ps, char * content)
     char * p1 = content; /* begining of a line */
     int flag = 0;
     int lineno = 0;
+    *error = "";
```
sbird: So *error is never NULL?

rainwoodman:
I tested it works. Will merge if this turns green.

sbird:
A fixed travis, nice.

rainwoodman:
Green!

PR #311, opened 2019-07-07
Densitysplit

sbird:
This moves one of the DhsmlDensityFactor variables into a local variable of density.c, further reducing the overhead of DENSITY_INDEPENDENT_SPH. No change to the hydro example, modest performance increase.

### DIFF
--- a/libgadget/density.c
+++ b/libgadget/density.c
@@ -458,10 +458,7 @@ density_postprocess(int i, TreeWalk * tw)
         else
             DhsmlDens = &(SPHP(i).DhsmlEgyDensityFactor);
         *DhsmlDens *= P[i].Hsml / (NUMDIMS * SPHP(i).Density);
-        if(*DhsmlDens > -0.9)	/* note: this would be -1 if only a single particle at zero lag is found */
-            *DhsmlDens = 1 / (1 + *DhsmlDens);
-        else
-            *DhsmlDens = 1;
+        *DhsmlDens = 1 / (1 + *DhsmlDens);
 
         /*Compute the EgyWeight factors, which are only useful for density independent SPH */
         if(DENSITY_GET_PRIV(tw)->DoEgyDensity) {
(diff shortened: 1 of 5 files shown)


libgadget/density.c
```diff
@@ -88,6 +88,11 @@ typedef struct {
 struct DensityPriv {
     MyFloat *Left, *Right;
     MyFloat (*Rot)[3];
+    /* This is the DhsmlDensityFactor for the pure density,
+     * not the entropy weighted density. It is used here and
+     * copied to DhsmlEgyDensityFactor if DensityIndependentSphOn = 1
+     * If DensityIndependentSphOn = 0 it is used to set DhsmlEgyDensityFactor. */
+    MyFloat * DhsmlDensityFactor;
```
rainwoodman: This comment is a bit confusing. If DensityIndependentSphOn is false there is no DhsmlEgyDensityFactor in the force equations. Probably you mean DhsmlEgyDensityFactor and DhslmDensityFactor are the same if not doing densityindependentSph?

libgadget/density.c
```diff
@@ -528,7 +551,13 @@ void density_check_neighbours (int i, TreeWalk * tw) {
             {
                 if(P[i].Type == 0 && fabs(P[i].NumNgb - desnumngb) < 0.5 * desnumngb)
                 {
-                    double fac = 1 - (P[i].NumNgb - desnumngb) / (NUMDIMS * P[i].NumNgb) * SPHP(i).DhsmlDensityFactor;
+                    MyFloat DensFac;
+                    if(DENSITY_GET_PRIV(tw)->DhsmlDensityFactor)
```
rainwoodman: Maybe you want to use the master flag (cleary shows the motivation of the branching ) instead of a derived condition.

sbird: Yes, I do!

libgadget/density.c
sbird: Thanks, clarified.

rainwoodman:
Green. Merge?

sbird:
Just checking the Lya example is still good...

sbird:
Uh....I just found something worrying.

sbird:
This passes the lya example too, merging.

PR #313, opened 2019-07-07
Add a density test and fix a bunch of density bugs

sbird:
This adds a test for density and then fixes a few bugs connected with it. It warns if the difference between the upper and lower HSML bounds < 0.1 % of the lower bound, which is a catchall stopping condition for the code. It should never be hit in practice unless it is hiding another bug.

I also found a bug in the stopping condition for too small Hsml values. If an intermediate Hsml value was less than MinGasHsml the code would stop, even if the true Hsml should have been larger. Then there is a roundoff error induced by using the wrong condition to choose Right and Left which was causing wrong Hsml values in a small number of particles in the hydro run.

Finally I build the density redo queue as we go, which should help large simulations a bit (although I cannot measure any speed change).

### DIFF
--- a/libgadget/density.c
+++ b/libgadget/density.c
@@ -510,11 +510,9 @@ void density_check_neighbours (int i, TreeWalk * tw) {
         }
 
         /* If we need more neighbours, move the lower bound up. If we need fewer, move the upper bound down.*/
-        if(P[i].NumNgb < (desnumngb - All.MaxNumNgbDeviation)) {
-            if(Left[i] < P[i].Hsml)
+        if(P[i].NumNgb < desnumngb) {
                 Left[i] = P[i].Hsml;
         } else {
-            if(Right[i] > P[i].Hsml)
                 Right[i] = P[i].Hsml;
         }
 
--- a/libgadget/density.c
+++ b/libgadget/density.c
@@ -303,7 +303,6 @@ density_reduce(int place, TreeWalkResultDensity * remote, enum TreeWalkReduceMod
     {
         TREEWALK_REDUCE(SPHP(place).Density, remote->Rho);
 
-
         TREEWALK_REDUCE(SPHP(place).DivVel, remote->Div);
         int pi = P[place].PI;
         TREEWALK_REDUCE(DENSITY_GET_PRIV(tw)->Rot[pi][0], remote->Rot[0]);
@@ -324,7 +323,6 @@ density_reduce(int place, TreeWalkResultDensity * remote, enum TreeWalkReduceMod
         }
         else
             TREEWALK_REDUCE(SPHP(place).DhsmlEgyDensityFactor, remote->DhsmlDensity);
-
     }
 
 }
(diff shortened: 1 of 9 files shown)


sbird:
Thinking it over, this is worrying but less severe than I thought. The code is still I think wrong, but it will only be a problem if individual particles have a very large smoothing length, not if they are near the upper edge of the box.

sbird:
Something else is wrong here...

sbird:
Obvious now I found it! I think I might write a density() test to catch more issues.

sbird:
Ok, this passes tests now and seems to work. I'm done with the density code for a bit. It is impressive - even small, seemingly inconsequential changes make it perform noticeably worse or introduce bugs!
Please review for merging.

libgadget/density.c
```diff
@@ -202,29 +199,63 @@ density(int update_hsml, int DoEgyDensity, ForceTree * tree)
...
+        if(update_hsml) {
+            if(!alloc_high) {
+                ReDoQueue = (int *) mymalloc2("ReDoQueue", tsize * sizeof(int) * NumThreads);
+                alloc_high = 1;
+            }
+            else {
+                ReDoQueue = (int *) mymalloc("ReDoQueue", tsize * sizeof(int) * NumThreads);
+                alloc_high = 0;
+            }
+            gadget_setup_thread_arrays(ReDoQueue, DENSITY_GET_PRIV(tw)->NPRedo, DENSITY_GET_PRIV(tw)->NPLeft, tsize, NumThreads);
```
rainwoodman: BTW, what was our verdict on jemalloc?

libgadget/density.c
```diff
@@ -365,7 +396,7 @@ density_ngbiter(
 
     if(All.WindOn) {
         if(winds_is_particle_decoupled(other))
-            if(!(I->DelayTime > 0))	/* if I'm not wind, then ignore the wind particle */
+            if(!(I->Type == 0 && I->DelayTime > 0))	/* if I'm not wind, then ignore the wind particle */
```
rainwoodman: So Bh were randomly not getting a density update?

libgadget/density.c
```diff
@@ -374,17 +405,24 @@ density_ngbiter(
...
+
     if(r2 < iter->kernel.HH)
     {
-
         const double u = r * iter->kernel.Hinv;
         const double wk = density_kernel_wk(&iter->kernel, u);
+        O->Ngb += wk * iter->kernel_volume;
+
+        /* For the BH only Ngb is used*/
+        if(I->Type == 5)
```
rainwoodman: ?? BH model uses density. Did you recompute it later?

libgadget/density.c
```diff
@@ -450,14 +479,20 @@ density_postprocess(int i, TreeWalk * tw)
...
-            endrun(12, "Particle %d has bad density: %g\n", i, SPHP(i).Density);
         int PI = P[i].PI;
         MyFloat * DhsmlDens;
         if(DENSITY_GET_PRIV(tw)->DoEgyDensity)
             DhsmlDens = &(DENSITY_GET_PRIV(tw)->DhsmlDensityFactor[PI]);
         else
             DhsmlDens = &(SPHP(i).DhsmlEgyDensityFactor);
+        if(SPHP(i).Density <= 0) {
+            if(P[i].NumNgb == 0) {
+                SPHP(i).Density = 1;
```
rainwoodman: Applying the MINHSML bound?

rainwoodman: There is almost no particle nearby, but in certain 'symmetriziation' of SPH pairs the density may contribute to other particles. I suspect 0 here may be safter than a undimensioned number 1.

libgadget/tests/test_density.c
```diff
@@ -0,0 +1,335 @@
...
+    }
+    assert_true(isfinite(minHsml));
+    assert_true(minHsml >= All.MinGasHsml);
+    assert_true(maxHsml <= All.BoxSize);
+
+}
+
+static void do_density_test(void ** state, const int numpart, double expectedhsml, double hsmlerr)
+{
+    /*Sort by peano key so this is more realistic*/
```
rainwoodman: Does it still work without sorting?

libgadget/tests/test_forcetree.c
```diff
@@ -47,6 +47,10 @@ struct forcetree_testdata
 
 void dump_snapshot() { }
 
+double walltime_measure_full(char * name, char * file, int line) {
+    return MPI_Wtime();
+}
+
 /*End dummies*/
```
rainwoodman: I recently learned the proper name of these dummies are mocks.

libgadget/density.c
sbird: Sorry, I didn't get a chance to look into it.

libgadget/density.c
sbird: BH were randomly including wind particles in their density estimation

libgadget/tests/test_forcetree.c
sbird: Well sure but as a british chap I delight in using unusual arguably incorrect names for things :-)

libgadget/tests/test_density.c
sbird: The sorting is not actually done because it was too much of a pain to keep Slots and P in sync in the test. Sorting only makes the treebuild benchmark more realistic, it does not affect correctness.

libgadget/density.c
sbird: Ah. Haha. It does! But because this was not in Reduce, this was reading from an uninitialised variable all along! Doh! Nice catch!

libgadget/density.c
rainwoodman: OK. It looks like jemalloc doesn't record the line number of memory allocation / tag; so it is mostly out -- our allocation are mostly 'huge allocation's, an it seems jemalloc is mostly optimized for small alloc, and may actually perform bad regarding segmentation. 

But I think I found a way to avoid requiring exact ordering of malloc and free; and perhaps also a way to record full stack information when an allocation occurs. I'll think about this more.

libgadget/density.c
rainwoodman: Oops. Now BH always include wind particles in there density, which doesn't sound sensible but was the way it was supposed to be, I think.

libgadget/tests/test_forcetree.c
rainwoodman: I don't actually know the origin of Mock. Dummy actually reads more lively.

libgadget/density.c
rainwoodman: No it looks like blackhole.c recomputes the density based on the Hsml computed in density.c -- thus we do not need density of BH here at all. Actually we could have splited density.c to two tree walks, Hsml that works on BH and Gas, and  sphdensity that only works on Gas, and caleed after Hsml iteration is done.

libgadget/density.c
sbird: Done - this almost puts back the original condition from gadget-2. The MINHSML bound should not affect this - if NumNgb == 0, the code will increase Hsml.

libgadget/density.c
sbird: Oh no, that's not true. BH Density is computed in blackhole.c, during accretion treewalk. The Density computed here is not used. The current code is fine. I have added a comment.

sbird: I did try (hard) to split the density.c into two tree walks, one for BH and one for gas but it was unfortunately always slower.

libgadget/density.c
sbird: Ok not jemalloc then. One related thing that occurred to me is to generalise the algorithm in sfr_eff.c to allow realloc() to succeed even when called on not the top allocation.

rainwoodman: As long as we ensure the heap state after a time-step is identical then there is no long term fragmentation. So we don't really need ordered alloc/free. Time for a new allocator :-)

libgadget/density.c
rainwoodman: That's surprising -- it seems like we only do less work by splitting - we can avoid the redundant calculation of rho, dhsml, grad rho etc in the iterations. 

BTW, there is an almost identical DM Hsml solver in winds.c, (for the vdisp dependent winds).

rainwoodman: That's surprising -- it seems like we only do less work by splitting - we can avoid the redundant calculation of rho, dhsml, grad rho etc in the iterations. 

BTW, there is an almost identical DM Hsml solver in winds.c, (for the vdisp dependent winds).

libgadget/density.c
sbird: hooray! :-)

libgadget/density.c
sbird: Yes, it surprised me too! It didn't seem to be tree walk overhead, so I think it had something to do with the caches.

sbird: I'll make a note to try unifying the DM and Gas HSML code at some point - we can't run winds in the same loop as density() because we need to know which are the new stars. But maybe we can reuse some routines. Or maybe move BH Hsml into blackhole accretion.

PR #314, opened 2019-07-10
Locallocks

sbird:
This PR moves the spin locks out of struct particle_data and into their own little file. This has no measureable effect on performance or output, but it tidies up the pthread code and means we can use the same routines in the tree build.


rainwoodman:
I recommend renaming unlock_particle(i, spin) to spinlocks_unlock(locks, i);
Since you are also using it for nodes.

sbird:
ok, will do that

PR #318, opened 2019-07-12
Buildfixes

sbird:
This PR messes with the Makefile to ensure that config.h gets rebuilt whenever the git version changes but not otherwise. Should avoid the situation we have now where the git version is sometimes out of date.


Makefile.rules
```diff
@@ -5,19 +5,22 @@ LOW_PRECISION ?= double
...
 OPTIONS = $(OPTIMIZE) $(OPT)
+GADGET_TESTDATA_ROOT = $(CURDIR)/../
 
 CFLAGS = $(OPTIONS) $(GSL_INCL)
 CFLAGS += -I../depends/include
 CFLAGS += -I../
 CFLAGS += "-DLOW_PRECISION=$(LOW_PRECISION)"
 CFLAGS += "-DHIGH_PRECISION=$(HIGH_PRECISION)"
+#For tests
+TCFLAGS = $(CFLAGS) -DGADGET_TESTDATA_ROOT=\"$(GADGET_TESTDATA_ROOT)\"
```
rainwoodman: What about constructing TCFLAGS in the Makefile that uses it? passing along GADGET_TESTDATA_ROOT is sufficient.

tests/stub.c
```diff
@@ -33,7 +32,6 @@ _cmocka_run_group_tests_mpi(const char * name, const struct CMUnitTest tests[],
     allocator_init(A_MAIN, "MAIN", 256 * 1024 * 1024, 1, NULL);
     allocator_init(A_TEMP, "TEMP", 8 * 1024 * 1024, 1, A_MAIN);
 
-    message(0, "GADGET_VERSION : %s\n", GADGET_VERSION);
```
rainwoodman: Why avoid rebuilding stub.c on version changes? Too many relinks taking a long time?

rainwoodman:
config.h is typically a dependence of all files. What about writing the version to config.c?Then only that file is rebuilt when version is changed -- and the rest is a relinking (which is unavoidable unless we want to modify the binaries after linking).

tests/stub.c
sbird: It means that any commit will rebuild all the tests, not just relink.

Makefile.rules
sbird: We could but that would mean duplicating this line and I am allergic to duplications

sbird:
Yes! That is a good idea.

sbird:
Done, nice.

rainwoodman:
We shall probably print the version and compilation time at library / app intialization ( I thought we did; but then why is petaio.c the only one that requires config.h?)

sbird:
We do print at startup but that is in gadget/main.c, which requires libgadget.a anyway

PR #320, opened 2019-07-16
Remove setting Evaluated from the treewalk

sbird:
Remove setting Evaluated from the treewalk.

Replace it with inline touching up of the export list. As long as a full export buffer is rare, this speeds up the tree walk substantially. If a full export buffer is common it does nothing.

### DIFF
--- a/libgadget/treewalk.c
+++ b/libgadget/treewalk.c
@@ -529,7 +529,7 @@ int treewalk_export_particle(LocalTreeWalk * lv, int no) {
             tw->BufferFullFlag = 1;
             /* This reduces the time until the other threads see the buffer is full and the loop can exit.
              * Since it is a pure optimization, no need for a full atomic.*/
-#pragma omp flush
+            #pragma omp flush (tw)
             /* Touch up the DataIndexTable, so that exports associated with the current particle
              * won't be exported. This is expensive but rare. */
             int i;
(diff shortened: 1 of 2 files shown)


libgadget/treewalk.c
```diff
@@ -560,25 +522,35 @@ int treewalk_export_particle(LocalTreeWalk * lv, int no) {
...
         if(nexp >= tw->BunchSize) {
-            /* out of buffer space. Need to discard work for this particle and interrupt */
             tw->BufferFullFlag = 1;
+            /* This reduces the time until the other threads see the buffer is full and the loop can exit.
+             * Since it is a pure optimization, no need for a full atomic.*/
 #pragma omp flush
+            /* Touch up the DataIndexTable, so that exports associated with the current particle
+             * won't be exported. This is expensive but rare. */
+            int i;
+            for(i=0; i < tw->BunchSize; i++) {
```
rainwoodman: Looks like you are disgarding the entire bunch instead of the last particle?

sbird: I don't think so - isn't lv->target set to the current particle? So I am discarding only elements of the Bunch which are related to the current particle.

sbird: If you confirm that I haven't misunderstood the code, I will add a comment.

rainwoodman: Yes. I agree with you on this. target is the only item that we have failed to export from this thread, thus this search is sufficient to revert the partial export of target.

sbird: Thanks, in that case I will merge this.

PR #321, opened 2019-07-19
Newtreebuild

sbird:
This modifies the tree build algorithm to both make the tree shallower and the tree build faster (prompted because the walking has gotten so much faster recently that build is noticeable again).

On a small run on my desktop this halves the treebuild time and speeds up density and hydro substantially (maybe 20-30%).

Instead of keeping adding each particle to a specific quadrant and splitting the quadrant when there are two such particles, this just adds each particle to an unordered list of NCHILD particles under the node. When the list is full, we create a layer of 8 child nodes and re-add the particles. This ensures that child nodes are close together, and it means that tree moment building doesn't have a hard-to-predict branch in it (each node either has nodes as children or particles). It also ensures that the relatively common case of two close particles doesn't cause the tree to balloon. It will balloon if we have more than 11 particles in the same place, but in that case we probably have bigger problems.

Once this is merged I might alter the treewalk a bit to make use of the fact that nodes either have node children or particle children and thus allow vectorization.

### DIFF
--- a/libgadget/forcetree.c
+++ b/libgadget/forcetree.c
@@ -226,13 +226,15 @@ static void init_internal_node(struct NODE *nfreep, struct NODE *parent, int sub
         const int sign = (subnode & (1 << j)) ? 1 : -1;
         nfreep->center[j] = parent->center[j] + sign*lenhalf;
     }
-    for(j = 0; j < 8; j++)
-        nfreep->u.suns[j] = -1;
+    for(j = 0; j < NMAXCHILD; j++)
+        nfreep->u.s.suns[j] = -1;
+    nfreep->u.s.noccupied = 0;
 }
 
 /* Size of the free Node thread cache.
- * 100 was found to be optimal for an Intel skylake with 4 threads.*/
-#define NODECACHE_SIZE 100
+ * 12 8-node rows (works out at 8kB) was found
+ * to be optimal for an Intel skylake with 4 threads.*/
+#define NODECACHE_SIZE (8*12)
 
 /*Structure containing thread-local parameters of the tree build*/
 struct NodeCache {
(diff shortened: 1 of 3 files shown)


libgadget/forcetree.c
```diff
@@ -226,13 +227,14 @@ static void init_internal_node(struct NODE *nfreep, struct NODE *parent, int sub
...
-        nfreep->u.suns[j] = -1;
+    for(j = 0; j < NCHILD; j++)
+        nfreep->u.s.suns[j] = -1;
+    nfreep->u.s.noccupied = 0;
 }
 
 /* Size of the free Node thread cache.
- * 100 was found to be optimal for an Intel skylake with 4 threads.*/
-#define NODECACHE_SIZE 100
+ * 9 was found to be optimal for an Intel skylake with 4 threads.*/
```
rainwoodman: Inconsistent with the value.

libgadget/forcetree.c
```diff
@@ -402,52 +387,76 @@ int force_tree_create_nodes(const ForceTree tb, const int npart, DomainDecomp *
...
         }
         while(child >= tb.firstnode);
 
         /*Now lock this node.*/
         lock_spinlock(this-tb.firstnode, spin);
+        /* We have a guaranteed spot.*/
+        nocc = atomic_fetch_and_add(&tb.Nodes[this].u.s.noccupied, 1);
+
+        /* Check whether there is now a new layer of nodes and if so walk down until there isn't.*/
+        if(nocc >= (1<<16)) {
```
rainwoodman: Is it possible to store the flag of 'have_node_children' and 'have_particle_children' to a more explicit variable? I think adding new member to the node struct is OK now as you are significiantlly reducing the number of nodes by allowing more than 1 particle in a node.

libgadget/forcetree.c
```diff
@@ -715,50 +727,52 @@ static int
...
-        if(suns[j] >= tree->firstnode && suns[j] < tree->lastnode)
-            chldcnt++;
+    int noccupied = tree->Nodes[no].u.s.noccupied;
+    for(j = 0; j < NCHILD; j++) {
+        suns[j] = tree->Nodes[no].u.s.suns[j];
     }
 
+    int limit = noccupied;
+    int childcnt = 0;
+    if(noccupied >= 1<<16) {
```
rainwoodman: This is for nodes with node children?

libgadget/forcetree.c
```diff
@@ -715,50 +727,52 @@ static int
...
+            }
+    }
     /*First do the children*/
-    for(j = 0; j < 8; j++)
+    for(j = 0; j < limit; j++)
     {
         int p = suns[j];
-        /*Empty slot*/
-        if(p < 0)
+        if(suns[j] < 0)
```
rainwoodman: why?

libgadget/forcetree.c
```diff
@@ -715,50 +727,52 @@ static int
...
             continue;
         /* For particles and pseudo particles we have nothing to update; */
         /* But the new tail is the last particle in the linked list. */
         if(p < tree->firstnode || p >= tree->lastnode) {
```
rainwoodman: Why is it no longer necessasry to find the correct tail?

libgadget/forcetree.h
```diff
@@ -8,6 +8,8 @@
  * ------------------
  */
 
+#define NCHILD 11
```
rainwoodman: NMAXCHILD ?

libgadget/forcetree.h
```diff
@@ -23,7 +25,11 @@ struct NODE
     } f;
     union
     {
-        int suns[8];		/*!< temporary pointers to daughter nodes */
+        struct
+        {
+            int suns[NCHILD];		/*!< temporary pointers to daughter nodes */
```
rainwoodman: Is it also used to store particle indices?

libgadget/forcetree.c
```diff
@@ -254,88 +256,72 @@ int get_freenode(int * nnext, struct NodeCache *nc)
...
+    /*Initialize the remaining entries to empty*/
+    for(i=8; i<NCHILD;i++)
+        nprnt->u.s.suns[i] = -1;
 
-    if (ninsert != parent) {
-        /* A new node is creatd. Added to the parent: this goes last
-         * so that we don't access the child before it is constructed.*/
-        #pragma omp atomic write
-        tb.Nodes[parent].u.suns[subnode] = ninsert;
+    for(i=0; i < NCHILD+1; i++) {
```
rainwoodman: Why +1?

libgadget/forcetree.c
```diff
@@ -254,88 +256,72 @@ int get_freenode(int * nnext, struct NodeCache *nc)
...
-        /* The new leaf will replace p_child in the parent (done before return)
-         * Re-attach that particle to the new leaf.*/
-        child_subnode = get_subnode(nfreep, p_child);
-        new_subnode = get_subnode(nfreep, p_toplace);
-
-        tb.Father[p_child] = ninsert;
-        nfreep->u.suns[child_subnode] = p_child;
-    }
-
-    /* If these target slot is empty or if the new node is too small.
```
rainwoodman: Removing this is cool!

libgadget/forcetree.c
```diff
@@ -254,88 +256,72 @@ int get_freenode(int * nnext, struct NodeCache *nc)
...
+create_new_node_layer(int parent, int p_toplace,
+        const ForceTree tb, int *nnext, struct NodeCache *nc)
+{
+    int i;
+    int oldsuns[NCHILD+1];
+    struct NODE *nprnt = &tb.Nodes[parent];
+
+    /* Copy the old particles and a new one into a temporary array*/
+    memcpy(oldsuns, nprnt->u.s.suns, NCHILD * sizeof(int));
+    oldsuns[NCHILD] = p_toplace;
```
rainwoodman: Reusing the last element of oldsuns doesn't smell nice, but maybe there is no better way?

rainwoodman:
See if I understand this PR correctly:

- Gadget-2 Tree insists leaf node either contains a single particle, or is a pseudo node that represents regions on a different process.

- This PR relaxes that constraint to allowing leaf nodes to contain several particles. The max number is NCHILD (NMAXCHILD). There fore we now have three types of nodes:
  - pseudo: represents a region on a different process
  - non-leaf, concrete : OCT tree internal nodes with 8 children nodes
  - leaf, concrete : OCT tree leaf nodes with NCHILD children particles.

If this sounds correct, then the data structure could probably be slighty rearranged to represent these in a flag / type variable. The current form this critical 'type of node' information appears to be inferred.

sbird:
That's right, and a good idea. I will add a flag to the struct. I will likely only use it after the tree is built - the current system makes the threading easier, because I can get a 'free' slot even if it later turns out that the oct node contains 8 child nodes, not particles.

libgadget/forcetree.c
sbird: I will do this, but I will leave the nocc logic alone as well: it helps simplify the threading.

libgadget/forcetree.c
sbird: merge error removed

libgadget/forcetree.c
sbird: We are still finding the correct tail: the logic is to ensure the tails are not confused by the presence of empty subnodes

libgadget/forcetree.c
sbird: You are right - there is a better way!

libgadget/forcetree.c
sbird: The loop is not needed anymore because we don't have a linked list of sub particles

libgadget/forcetree.h
```diff
@@ -20,10 +22,19 @@ struct NODE
         unsigned int DependsOnLocalMass :1;  /* Intersects with local mass */
         unsigned int MixedSofteningsInNode:1;  /* Softening is mixed, need to open the node */
         unsigned int NodeIsDirty :1; /*Node is a toplevel node containing local mass, and its moments need updating*/
+        unsigned int InternalNode :1; /* If true, node contains only other nodes.
```
rainwoodman: Do we already use the word Internal for something else? Is there chance for confusion?

libgadget/forcetree.h
```diff
@@ -20,10 +22,19 @@ struct NODE
...
     union
     {
-        int suns[8];		/*!< temporary pointers to daughter nodes */
+        struct
+        {
+            /*!< temporary pointers to daughter nodes or daughter particles. */
+            int suns[NMAXCHILD];
+            /* Number of daughter particles if node contains particles.
+             * During treebuild >= (1<<16) if node contains nodes.*/
+            int noccupied;
```
rainwoodman: What about splitting it to two short ints, one for noccupried, and one for InternalNode?

rainwoodman: Or hide 1<<16 to a macro ?

libgadget/forcetree.c
rainwoodman: Are you confident the old loop was only looping through particles inside the same node? get_next_node doesn't guarantee ending at a node boundary -- it would go through any partially created traversal link-list..

libgadget/forcetree.h
sbird: We have InternalTopNode - this might be confusing. It was my original intention a few hours ago to merge them, but that isn't going to work, I think. But I can't think of a better name, do you have any ideas?

libgadget/forcetree.h
sbird: I want to keep InternalNode around so I can use it in the future in the treewalk code to reduce branches (not sure how yet). Currently 1<<16 is only used during the first phase of the treebuild to make it one less atomic read. It's a little ugly but it really means "a number much larger than NCHILDMAX" and I think that's ok. All the macros I can think of are excessively magic because of the atomic-ness.

sbird:
Uh, Yu, we may have a bigger problem...the travis build fails during insert_pseudo_particles() with this error:
[ 000359.48 ] Allocated 69.2105 MByte for BH-tree, (presently allocated 170.855 MB)
[ 000359.56 ] Task 1: Error, adding a pseudo particle 891346 (task 0) to already full node 524289 with 11 children
which happens when we try to add a pseudo particle to a node already full of real particles. This is not really a problem to fix because I can just have the code create an array of child nodes beneath the full node and add the pseudo particle the same way I would add a real particle.

However - I don't understand why this is happening. That code hasn't changed much with this PR and domain_maintain() should be enforcing that nodes with pseudo particles contain zero real particles.

So I instrumented insert_peudo_particles() in master as so:
static void
force_insert_pseudo_particles(const ForceTree * tree, const DomainDecomp * ddecomp)
{
    int i, index;
    const int firstpseudo = tree->lastnode;
    int ThisTask;
    MPI_Comm_rank(MPI_COMM_WORLD, &ThisTask);

    for(i = 0; i < ddecomp->NTopLeaves; i++)
    {
        index = ddecomp->TopLeaves[i].treenode;

        if(ddecomp->TopLeaves[i].Task != ThisTask) {
            if(tree->Nodes[index].u.suns[0] > -1)
                endrun(1, "In node %d, want to overwrite child %d with pseudo particle %d\n", index, tree->Nodes[index].u.suns[0], i);
            tree->Nodes[index].u.suns[0] = firstpseudo + i;
            force_set_next_node(firstpseudo + i, -1, tree);
        }
    }
}

and it died on step 9 in the travis run:
[ 000173.34 ] Task 1: In node 524342, want to overwrite child 38591 with pseudo particle 22

which I think means our treewalk is occasionally losing particles an may be buggy. I am worried.

sbird:
Aha! It is a garbage particle!

sbird:
See https://github.com/MP-Gadget/MP-Gadget/pull/322

sbird:
For exchanged particles this would not matter, because exchanged garbage will always by definition be underneath a pseudo-particle and the mass moment of a pseudoparticle is over-written during exchange_pseudodata(). But in the presence of garbage from black holes or star formation (not Lyman-alpha) we would double-count the mass of newly formed stars on short timesteps! It is good we fix that.

libgadget/forcetree.c
sbird: That is true, but the old code (somewhat obscurely) set nextnode = -1 by hand on every particle added to the tree, so the lists were always of length 0 unless the node was too small. The code is:

        /*If the node is too small we prepend the particle to a short linked list.*/
        force_set_next_node(p_toplace, too_small ? p_child : -1, &tb);
So before it was indeed only looping through particles inside the same node.

libgadget/forcetree.h
sbird: How about a two-bit bitfield called ChildType which can be enum ChildType{PARTICLE_TYPE=0, NODE_TYPE = 1, PSEUDO_TYPE = 2} ? That way we can avoid some of the messing about with pseudo particle checks as well. The enums can be defines if enum bitfields are not allowed.

sbird:
What do you think of that? Can you confirm that C enums and C bitfields do combine in the way I want them to so one can have a 2-bit wide enum? Someone on stack overflow suggests that you cannot in C, but it compiles without a warning here. https://stackoverflow.com/questions/11983231/is-it-safe-to-use-an-enum-in-a-bit-field.

I like that this approach cleanly separates pseudo-particle nodes, nodes of nodes and nodes of particles.

rainwoodman:
unfortunately bitfield and enum is not standardized..
 What about interger bitfields and macros?

rainwoodman:
another concern about pseudo particles.

I worry that each pseudo particle must use a single node because the
information about the hosting process is stored in the node, not in the
particle.

sbird:
That is very true and I switched to enforcing one pseudo per node in the last push. I will also change to bitfields and defines soon.

sbird:
How do you like that? I used that we now have well-defined types of node to split the node updating into separate functions depending on what is being updated, which is much cleaner. It also makes the threading much better because we only spawn new tasks for nodes with child nodes which have enough to do that the extra task is worth the overhead.

sbird:
Once this is tested and merged I will try to do something similar to the tree walking. That will need me to keep the suns array around somewhere, but might be worth it if it allows vectorizing the tree loop.

sbird:
Passes a hydro test...of course the parictle distribution is not identical because the tree opening has changed, but it is similar enough. Now running a large-scale lyman-alpha test on frontera

sbird:
Passes the lyman alpha test and is about 10% faster overall on the 512^3 box on frontera. This may have been the smoothest major change I ever made to the code...

sbird:
Ready to merge I think.

rainwoodman:
Go for it!

PR #323, opened 2019-07-23
Nopthread

sbird:
This PR is a pure code cleanup which changes the calls to the pthread spinlock API into calls to OpenMP locks, since we require openmp 3 anyway now. This should improve mac compatibility but is mostly just a code cleanup. I haven't tested yet but I would be surprised if it had any effect on the code. The omp flush calls are not necessary anymore because they are implicit after an omp_unlock().


rainwoodman:
Looks good -- what's up with travis?

sbird:
Travis runs in no-thread mode so I think just a glitch...

sbird:
This passes a threaded test here.

PR #324, opened 2019-07-30
Revert "Merge pull request #323 from sbird/nopthread"

sbird:
I found that using OMP locks slows down the tree build by a factor of 200 (!) so that it now takes 85% of the simulation runtime. This only seems to happen on intel, so it affects stampede and frontera. About 60% of the time was spent in init_spinlocks, 30% spent in creating the nodes. This might be a bug with intel's openmp implementation, or it might be because it is using mutexes instead of spinlocks. godbolt suggests that icc does produce much longer code for init_spinlocks than gcc 

In any case, revert this because it is not worth the effort to root cause.

@rainwoodman : Any better ideas for this?

### DIFF
--- a/Options.mk.example
+++ b/Options.mk.example
@@ -15,6 +15,8 @@ GSL_LIBS = $(shell pkg-config --libs gsl)
 #OPT += -DLIGHTCONE                       # write a lightcone on the fly; in development
 #OPT += VALGRIND     # allow debugging with valgrind, disable the GADGET memory allocator.
 #OPT += -DDEBUG      # print a lot of debugging messages
+#Disable openmp locking. This means no threading.
+#OPT += -DNO_OPENMP_SPINLOCK
 
 #-------------------------------------------- Things for special behaviour
 #OPT	+=  -DNO_ISEND_IRECV_IN_DOMAIN     #sparse MPI_Alltoallv do not use ISEND IRECV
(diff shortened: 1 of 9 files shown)


rainwoodman:
Is there a magic cmdline parameter or env that instructs intel to use the
spinlocks? IMPI tends to do this. Do you have the hyperlink to godbolts?

sbird:
Here is the godbolt for icc: https://godbolt.org/z/8ZUA9s
You can see the gcc variant is much simpler. 

I can't find an environment variable for this: omp locks seem fairly unloved. openmp 5 allows for lock hints that can be used to suggest a mutex or spinlock but we can't require that yet.

sbird:
There is also supposedly an implicit flush after the spinlock is unlocked, but I could not find a way to control what it applied to, which seems like it would be a natural api extension.

sbird:
One thing I wondered is if intel flushes all memory at the end of the init_spinlock() call: since this is done in parallel (per the example in the openmp standard document) that would create crazy cache line traffic and contention, which would explain the results.

rainwoodman:
Looks like the main difference is in iomp and gomp library?

Is it possible to compile with icc but link with gomp? The complication in
_init shouldn't have mattered much since it is called only once.

I suppose the flush should be fine if it is only flushing the addresses
around the lock -- if not it will be very bad.

- Yu

sbird:
Yes, I think so. I suspect linking to gomp while compiling with icc would resolve this, but would be perilous and lead to weird errors. But also - if we are doing that, why not just link to pthread? It is available everywhere gomp is.

PR #326, opened 2019-08-03
Earlytamalloc

sbird:
This is a series of tiny memory changes done in an attempt to improve stability on frontera by changing the remaining mallocs to be on our own heap. They help somewhat. The main change is fastpm_strdup_printf which is now done on the temp heap, which is allocated first, before parsing the parameter file. Several static buffers were changed to use fastpm_strdup_printf. The pidfile printing was fixed to not open an arbitrary file and print to it!

I was also nazi about string copies in case the compiler was over-optimizing, but I don't think that did anything.

### DIFF
--- a/gadget/params.c
+++ b/gadget/params.c
@@ -40,11 +40,6 @@ StarformationCriterionAction(ParameterSet * ps, char * name, void * data)
     return 0;
 }
 
-int cmp_double(const void * a, const void * b)
-{
-    return ( *(double*)a - *(double*)b );
-}
-
 /*! This function parses a string containing a comma-separated list of variables,
  *  each of which is interpreted as a double.
  *  The purpose is to read an array of output times into the code.
@@ -95,7 +90,6 @@ OutputListAction(ParameterSet * ps, char * name, void * data)
 /*         message(1, "Output at: %g\n", All.OutputListTimes[count]); */
     }
     myfree(strtmp);
-    qsort(All.OutputListTimes, All.OutputListLength, sizeof(double), cmp_double);
     return 0;
 }
 
--- a/gadget/params.c
+++ b/gadget/params.c
@@ -54,7 +54,7 @@ static int
 OutputListAction(ParameterSet * ps, char * name, void * data)
 {
     char * outputlist = param_get_string(ps, name);
-    char * strtmp=fastpm_strdup(outputlist);
+    char * strtmp = fastpm_strdup(outputlist);
     char * token;
     int count;
 
--- a/gadget/params.c
+++ b/gadget/params.c
@@ -59,7 +59,7 @@ static int
 OutputListAction(ParameterSet * ps, char * name, void * data)
 {
     char * outputlist = param_get_string(ps, name);
-    char * strtmp=strdup(outputlist);
+    char * strtmp=fastpm_strdup(outputlist);
     char * token;
     int count;
 
@@ -94,7 +94,7 @@ OutputListAction(ParameterSet * ps, char * name, void * data)
         All.OutputListTimes[count] = a;
 /*         message(1, "Output at: %g\n", All.OutputListTimes[count]); */
     }
-    free(strtmp);
+    myfree(strtmp);
     qsort(All.OutputListTimes, All.OutputListLength, sizeof(double), cmp_double);
     return 0;
 }
--- a/gadget/params.c
+++ b/gadget/params.c
@@ -370,13 +370,13 @@ void read_parameter_file(char *fname)
         All.NumThreads = omp_get_max_threads();
 
     /* Start reading the values */
-        param_get_string2(ps, "InitCondFile", All.InitCondFile);
-        param_get_string2(ps, "OutputDir", All.OutputDir);
-        param_get_string2(ps, "SnapshotFileBase", All.SnapshotFileBase);
-        param_get_string2(ps, "FOFFileBase", All.FOFFileBase);
-        param_get_string2(ps, "EnergyFile", All.EnergyFile);
+        param_get_string2(ps, "InitCondFile", All.InitCondFile, sizeof(All.InitCondFile));
+        param_get_string2(ps, "OutputDir", All.OutputDir, sizeof(All.OutputDir));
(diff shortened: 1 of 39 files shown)


gadget/params.c
```diff
@@ -59,7 +54,7 @@ static int
 OutputListAction(ParameterSet * ps, char * name, void * data)
 {
     char * outputlist = param_get_string(ps, name);
-    char * strtmp=strdup(outputlist);
+    char * strtmp=fastpm_strdup(outputlist);
```
rainwoodman: " = ", now that we are touching this line.

libgadget/utils/mymalloc.c
```diff
@@ -24,16 +25,36 @@ Allocator A_TEMP[1];
...
 
+void
+tamalloc_init(void)
+{
+    int Nt = omp_get_max_threads();
+    int NTask;
+    MPI_Comm_size(MPI_COMM_WORLD, &NTask);
+
+    /* reserve 128 bytes per task and 128 bytes per thread for TEMP storage.*/
+    size_t n = 4096 * 1024 + 128 * NTask + 128 * Nt * NTask;
```
rainwoodman: reserve 4MB and ...

libgadget/utils/mymalloc.c
```diff
@@ -24,16 +25,36 @@ Allocator A_TEMP[1];
...
+
+    if (MPIU_Any(ALLOC_ENOMEMORY == allocator_init(A_TEMP, "TEMP", n, 1, NULL), MPI_COMM_WORLD)) {
+        endrun(0, "Insufficient memory for the TEMP allocator on at least one nodes."
+                  "Requestion %td bytes. Try reducing MaxMemSizePerNode. Also check the node health status.\n", n);
+
+    }
+}
+
 void
 mymalloc_init(double MaxMemSizePerNode)
```
rainwoodman: Is it worth it to change the interface to
mymalloc_init() {
ta_malloc_init() }

mymalloc_set_size(....) {
   ma_malloc_init(...)
}

and replace all mymalloc to ma_malloc().

libgadget/utils/paramset.c
```diff
@@ -351,13 +363,14 @@ param_get_string(ParameterSet * ps, char * name)
...
 void
-param_get_string2(ParameterSet * ps, char * name, char * dst)
+param_get_string2(ParameterSet * ps, char * name, char * dst, size_t len)
 {
     ParameterSchema * p = param_get_schema(ps, name);
     if (param_is_nil(ps, name)) {
         printf("Accessing an undefined parameter `%s`.\n", p->name);
     }
-    strcpy(dst, ps->value[p->index].s);
+    strncpy(dst, ps->value[p->index].s, len);
```
rainwoodman: I think we shall die instead of truncation when the receiver is not large enough.

libgadget/utils/system.c
```diff
@@ -458,66 +458,41 @@ int MPI_Alltoallv_sparse(void *sendbuf, int *sendcnts, int *sdispls,
...
-            nid[i] = nid[i - 1];
-        }
-    }
-    for(i = 0; i < NTask; i ++) {
-        if(!strcmp(hostname, buffer + i * ml)) {
-            break;
+
+    /* Size is set by the minimal size of the temp heap*/
+    const int bufsz = 256;
+    char * buffer = ta_malloc("buffer", char, bufsz * NTask);
```
rainwoodman: This almost overflows ta_malloc's limit. Perhaps add a comment about the crossref between bufsz and ta_malloc_init.

libgadget/utils/system.c
```diff
@@ -609,33 +584,29 @@ MPIU_write_pids(char * filename)
...
-    my_pid = getpid();
-
-    strcpy(mode, "a+");
 
-    for(i = 0; i < NTask; i++)
+    int my_pid = getpid();
+    int * pids = ta_malloc("pids", int, NTask);
+    char * hosts = ta_malloc("hosts", char, NTask * 32);
+    char host[32]={"NULL"};
+    char * genv = getenv("HOST");
```
rainwoodman: Use get_hostname rather than HOST.

Also the max size limit here is 32, but the one above was (bufsz) 256. I think 64 is a pretty safe value -- perhaps use that instead?

libgadget/walltime.c
```diff
@@ -27,10 +27,10 @@ void walltime_init(struct ClockTable * ct) {
 }
 
 static void walltime_summary_clocks(struct Clock * C, int N, int root, MPI_Comm comm) {
-    double t[N];
-    double min[N];
-    double max[N];
-    double sum[N];
+    double * t = ta_malloc("clocks",double, 4 * N);
```
rainwoodman: ", "

libgadget/walltime.c
```diff
@@ -82,8 +83,8 @@ static int clockcmp(const void * c1, const void * c2) {
 static void walltime_clock_insert(char * name) {
     if(name[0] != '/') abort();
     if(strlen(name) > 1) {
-        char tmp[80];
-        strcpy(tmp, name);
+        char tmp[80] ={0};
```
rainwoodman: "= "

libgadget/utils/mymalloc.c
sbird: hmm, that seems a lot of effort. I don't think it is worth it.

sbird: The point is after all to do ta_malloc before ma_malloc so the wrapper is not useful.

libgadget/utils/system.c
sbird: Oh nice catch! I'll change the pidfile to match cluster_get_num_hosts. 

in cluster_get_num_hosts, the extra buffer size costs us nothing and could potentially help us some day when someone runs on a cluster with really long node names. But in pidfile an overflow is benign but running out of memory is a potential issue.

sbird:
While you are here...it is persistently hanging on frontera just after "Probing Header of snapshot file:"
 inside big_file_mpi_open. I'm having trouble making a reproducer because it seems that the hang, while easy to trigger with the default code, easily goes away if I remove code. It requires > 3000 processors to trigger, and seems to need the root task to be somewhat behind the others. It also needs 64 bit alignment specified in the compiler. Any ideas?

PR #327, opened 2019-08-05
Drift BH to halo center more aggressively.

sbird:
Disable the sound speed check for black hole drift, as is done in bluetides-ii. We suspect it is causing the BH to drift away from the halo center.

### DIFF
--- a/libgadget/blackhole.c
+++ b/libgadget/blackhole.c
@@ -45,7 +45,6 @@ typedef struct {
 typedef struct {
     TreeWalkResultBase base;
     MyFloat BH_MinPotPos[3];
-    MyFloat BH_MinPotVel[3];
     MyFloat BH_MinPot;
 
     int BH_TimeBinLimit;
@@ -89,7 +88,6 @@ struct BHPriv {
     /*Temporary array to store the IDs of the swallowing black hole for gas*/
     MyIDType * SPH_SwallowID;
     double  (* MinPotPos) [3];
-    MyFloat (* MinPotVel) [3];
     MyFloat * MinPot;
 
     /* Particle SpinLocks*/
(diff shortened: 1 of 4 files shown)


rainwoodman:
I don't think we shall drift BH by the velocity in drift.c, neither.

At one point I was convinced to add that, but last Wedensday talking to Yueyin I realized the velocity based drift  is essentially drifting with the velocity of a random particle in the halo -- and it means the BH is moving thermally in the halo instead of drifting to the center.

sbird:
That makes sense - I think you are right about the random drifting. There is the added complication that the accelerations for the BH are computed before the BH is repositioned, so it is also receiving essentially a random kick to its velocity. We should probably reposition and set the velocity to the gradient of the potential or something.

sbird:
Also by moving the BH in blackhole.c we ensure that the bh feedback actually applies to the center of the halo.

rainwoodman:
except we shall not move any particles during tree walk...

Yu

rainwoodman:
Let's merge this!

sbird:
It works, but I would like to tidy it up a bit before merging. I will add a comment about the tree breaking, although it is safe in this specific case because BH do not participate in the wind treewalk.

rainwoodman:
I see. Becasue this is the last time we use the tree on BH, it is OK to
break their tree invariance here.

We can probably remove the drifting logic from drift.c then.

sbird:
Any preferences about whether or not we keep the black hole sound speed check? Yueying says it works fine without it and it is an extra model complexity so I am tempted not to.

rainwoodman:
I didnt think it makes sense to check for sound speed with DM particles.

A bounding energy check is more reasonable if worried about pathelogical
cases.

So I vote for removal.

sbird:
Passed unanimously!

sbird:
Aha I see why the old code was not working at all! It is more simple than we thought. The black hole code should have been doing:

treewalk_preprocess() {
MinPotPos = Pos
}

treewalk_reduce(){
if(MinPot > Potential(neighbours)
MinPot = Potential(neighbours)
MinPotPos = Position(neighbours)
}

but it was actually doing:
treewalk_reduce(){
if(MinPot > Potential(neighbours)
MinPot = Potential(neighbours)
MinPotPos = Position(neighbours)
}

treewalk_preprocess() {
MinPotPos = Pos
}
because the preprocess function is treewalk_feedback_preprocess instead of treewalk_accretion_preprocess! So MinPotPos was always being over-written with the current position! 

hoho. We could reposition in drift.c if we wanted to with this fixed.

Should we?

sbird:
See this PR https://github.com/MP-Gadget/MP-Gadget/pull/328 which is the root cause.

rainwoodman:
Aha! That was it!

Yes drifting in drifting.c is way cleaner than drifting in treewalk.

sbird:
It is, but I am concerned because it *is* a BH model change. If we are to do it for real I would like to remove the "induced star formation" stuff at the same time and just check the star formation rates are roughly correct at z=0 once.

PR #328, opened 2019-08-08
Minimalfixbhdrift

sbird:
This is a minimal fix for the black hole drifting (and a couple of other memory bugs!). The drifting was broken because MinPotPos was being over-written with the current position after being set to the potential minimum. This PR should definitely be merged.

### DIFF
--- a/libgadget/timebinmgr.c
+++ b/libgadget/timebinmgr.c
@@ -36,8 +36,7 @@ setup_sync_points(double TimeIC, double no_snapshot_until_time)
 
     if(NSyncPoints > 0)
         myfree(SyncPoints);
-    SyncPoints = mymalloc("SyncPoints", sizeof(SyncPoints) * (All.OutputListLength+2));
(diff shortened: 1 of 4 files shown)


libgadget/timebinmgr.c
```diff
@@ -36,8 +36,7 @@ setup_sync_points(double TimeIC, double no_snapshot_until_time)
 
     if(NSyncPoints > 0)
         myfree(SyncPoints);
-    SyncPoints = mymalloc("SyncPoints", sizeof(SyncPoints) * (All.OutputListLength+2));
-    memset(&SyncPoints[0], -1, sizeof(SyncPoints[0]) * All.OutputListLength);
+    SyncPoints = mymalloc("SyncPoints", sizeof(SyncPoint) * (All.OutputListLength+2));
```
rainwoodman: Oops. Was the 0xff initialization used for aspecial default True value later?

sbird: No - in fact it was just hiding any potential uninitialized errors!

PR #331, opened 2019-08-15
Nodelist2

sbird:
This enables the NodeList for hydro as well as the others. Since the NodeList is then always on, clean up the related option. Also make the counters a bit more useful.

### DIFF
--- a/libgadget/gravpm.c
+++ b/libgadget/gravpm.c
@@ -193,39 +193,26 @@ static int pm_mark_region_for_node(int startno, int rid, const ForceTree * tree)
         if(node_is_particle(no, tree))	/* single particle */
         {
             int p = no;
+            P[p].RegionInd = rid;
+#ifdef DEBUG
             /* when we are in PM, all particles must have been synced. */
             if (P[p].Ti_drift != All.Ti_Current) {
                 abort();
             }
-
-            P[p].RegionInd = rid;
-            /*
-             *
-             * Enlarge the startno so that it encloses all particles
-             * this happens if a BH particle is relocated to a PotMin
-             * out-side the (enlarged )drifted node.
-             * because the POTMIN relocation is unphysical, this can
-             * happen immediately after a BH is seeded at the dense-most
-             * gas particle. rare rare event!
-             *
-             * */
+            /* Check for particles outside of the node. This should never happen,
+             * unless there is a bug in tree build, or the particles are being moved.*/
             int k;
             for(k = 0; k < 3; k ++) {
-                double l = P[p].Pos[k] - tree->Nodes[startno].center[k];
-                if (l < - 0.5 * All.BoxSize) {
-                    l += All.BoxSize;
-                }
-                if (l > 0.5 * All.BoxSize) {
-                    l -= All.BoxSize;
-                }
+                double l = NEAREST(P[p].Pos[k] - tree->Nodes[startno].center[k]);
                 l = fabs(l * 2);
                 if (l > tree->Nodes[startno].len) {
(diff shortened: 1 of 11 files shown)


libgadget/gravpm.c
```diff
@@ -193,39 +193,26 @@ static int pm_mark_region_for_node(int startno, int rid, const ForceTree * tree)
...
-                }
-                if (l > 0.5 * All.BoxSize) {
-                    l -= All.BoxSize;
-                }
+                double l = NEAREST(P[p].Pos[k] - tree->Nodes[startno].center[k]);
                 l = fabs(l * 2);
                 if (l > tree->Nodes[startno].len) {
                     if(l > tree->Nodes[startno].len * (1+ 1e-7))
-                    message(1, "enlarging node size from %g to %g, due to particle of type %d at %g %g %g id=%ld\n",
+                    endrun(1, "enlarging node size from %g to %g, due to particle of type %d at %g %g %g id=%ld\n",
```
rainwoodman: Does the blackhole drifting trigger this?

libgadget/treewalk.c
```diff
@@ -931,7 +920,7 @@ cull_node(const TreeWalkQueryBase * const I, const TreeWalkNgbIterBase * const i
  * max(P[other].Hsml, iter->Hsml).
  *
  * Particle that intersects with other domains are marked for export.
- * The hosting nodes are exported as well, if tw->UseNodeList is True.
+ * The hosting nodes are exported as well.
```
rainwoodman: Are exported nodes always the leaves of the global tree?

libgadget/treewalk.h
```diff
@@ -96,7 +96,7 @@ struct TreeWalk {
     int NTask; /*Number of MPI tasks*/
     int NThread; /*Number of OpenMP threads*/
 
-    int UseNodeList;      /* Send tree branches or use the entire tree for ghost particles */
+    /* When exporting we now always send tree branches.*/
```
rainwoodman: "Unlike Gadget-3."

libgadget/treewalk.h
```diff
@@ -51,6 +51,7 @@ typedef struct {
     int * ngblist;
     int64_t Ninteractions;
     int64_t Nnodesinlist;
+    int64_t Nlist;
```
rainwoodman: Since we are at this, could you also add a comment about the meaning of each counter?

libgadget/treewalk.c
```diff
@@ -876,7 +889,10 @@ int treewalk_visit_ngbiter(TreeWalkQueryBase * I,
     }
 
     lv->Ninteractions += ninteractions;
-    lv->Nnodesinlist += inode;
+    if(lv->mode == 1) {
+        lv->Nnodesinlist += inode;
+        lv->Nlist += 1;
```
rainwoodman: Is Nlist the number of nodelists used for this ghost particle, or used for this particle (sum of all ghost particles of the real particle)?

rainwoodman:
Looks good. I recommended a few comments for clarity of the code base while the memory is still fresh on the issue.

libgadget/gravpm.c
sbird: Not any more, because the black holes are moved at drift time and the tree is rebuilt afterwards anyway.

libgadget/treewalk.c
sbird: Yes, I think so.

libgadget/treewalk.c
sbird: The latter

sbird:
Thanks, comments added, merging!

PR #332, opened 2019-08-15
Fix the minimum energy floor

sbird:
The minimum energy floor was being applied inconsistently. This PR fixes that and cleans up some aspects of the cooling code.

### DIFF
--- a/libgadget/cooling.c
+++ b/libgadget/cooling.c
@@ -96,8 +96,8 @@ double DoCooling(double redshift, double u_old, double rho, double dt, struct UV
         /* If we know that the new energy
          * is below the minimum gas internal energy, we are done here.*/
         if(u_upper <= MinEgySpec) {
-                u = MinEgySpec;
-                break;
+            u = MinEgySpec;
+            break;
         }
 
         LambdaNet = get_heatingcooling_rate(rho, u, 1 - HYDROGEN_MASSFRAC, redshift, Z, uvbg, ne_guess);
--- a/libgadget/cooling.c
+++ b/libgadget/cooling.c
@@ -47,7 +47,7 @@ void init_cooling(char * TreeCoolFile, char * MetalCoolFile, char * UVFluctuatio
 /* returns new internal energy per unit mass.
  * Arguments are passed in code units, density is proper density.
  */
-double DoCooling(double redshift, double u_old, double rho, double dt, struct UVBG * uvbg, double *ne_guess, double Z)
+double DoCooling(double redshift, double u_old, double rho, double dt, struct UVBG * uvbg, double *ne_guess, double Z, double MinEgySpec)
 {
     if(!coolunits.CoolingOn) return 0;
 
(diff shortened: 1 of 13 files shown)


libgadget/cooling.c
```diff
@@ -67,44 +69,42 @@ double DoCooling(double redshift, double u_old, double rho, double dt, struct UV
     LambdaNet = get_heatingcooling_rate(rho, u, 1 - HYDROGEN_MASSFRAC, redshift, Z, uvbg, ne_guess);
 
     /* bracketing */
-
-    if(u - u_old - LambdaNet * dt < 0)	/* heating */
+    if(LambdaNet * dt > 0)	/* heating */
```
rainwoodman: No predict any more?

libgadget/run.c
```diff
@@ -500,7 +500,7 @@ set_units(void)
     init_cosmology(&All.CP, All.TimeIC);
 
     if(All.InitGasTemp < 0)
-        All.InitGasTemp = DMAX(All.MinGasTemp, All.CP.CMBTemperature / All.TimeInit);
+        All.InitGasTemp = All.CP.CMBTemperature / All.TimeInit;
```
rainwoodman: One concern was at high z this would be too low and cause over-cooling problems to surface.

sbird: Quite so! I'll revert the change.

libgadget/cooling.c
sbird: We have u = u_old above this so for this initial step the removed code did nothing. I can put it back though if you like.

libgadget/run.c
sbird: This is now applied in init.c

libgadget/cooling.c
```diff
@@ -67,33 +70,35 @@ double DoCooling(double redshift, double u_old, double rho, double dt, struct UV
...
+        } while(u_lower - u_old - get_heatingcooling_rate(rho, u_lower, 1 - HYDROGEN_MASSFRAC, redshift, Z, uvbg, ne_guess) * dt > 0);
     }
 
     do
     {
         u = 0.5 * (u_lower + u_upper);
+        /* If we know that the new energy
+         * is below the minimum gas internal energy, we are done here.*/
+        if(u_upper <= MinEgySpec) {
+                u = MinEgySpec;
```
rainwoodman: The indentation looks strange.

libgadget/density.c
```diff
@@ -27,6 +27,12 @@ SPH_EntVarPred(int i)
         /*Entropy limiter for the predicted entropy: makes sure entropy stays positive. */
         if(dloga > 0 && EntVarPred < 0.5*SPHP(i).Entropy)
             EntVarPred = 0.5 * SPHP(i).Entropy;
+        /* Ensures that entropy is always positive, even if we read a bad value from disc.*/
+        if(SPHP(i).Density > 0) {
```
rainwoodman: What about flag this and issue a warning if number of negatives is nonzero?

sbird:
I will merge after testing

PR #335, opened 2019-08-21
Black hole cleanups

sbird:
This removes some unloved parts of the black hole code. The major change is to the black hole timestepping: instead of the BH timestep being <= the smallest SPH timestep surrounding it, it is now = the smallest SPH timestep. This avoids situations where the smallest timesteps contain exclusively black holes.

### DIFF
--- a/libgadget/blackhole.c
+++ b/libgadget/blackhole.c
@@ -51,7 +51,6 @@ typedef struct {
 
     MyFloat Rho;
     MyFloat SmoothedEntropy;
-    MyFloat SmoothedPressure;
     MyFloat GasVel[3];
 } TreeWalkResultBHAccretion;
 
@@ -314,8 +313,6 @@ blackhole_accretion_postprocess(int i, TreeWalk * tw)
     if(BHP(i).Density > 0)
     {
         BHP(i).Entropy /= BHP(i).Density;
-        BHP(i).Pressure /= BHP(i).Density;
-
         for(k = 0; k < 3; k++)
             BH_GET_PRIV(tw)->BH_SurroundingGasVel[PI][k] /= BHP(i).Density;
     }
@@ -475,7 +472,6 @@ blackhole_accretion_ngbiter(TreeWalkQueryBHAccretion * I,
             /* FIXME: volume correction doesn't work on BH yet. */
             O->Rho += (mass_j * wk);
 
-            O->SmoothedPressure += (mass_j * wk * PressurePred(P[other].PI));
             O->SmoothedEntropy += (mass_j * wk * SphP_scratch->EntVarPred[P[other].PI]);
             O->GasVel[0] += (mass_j * wk * SphP_scratch->VelPred[3 * P[other].PI]);
             O->GasVel[1] += (mass_j * wk * SphP_scratch->VelPred[3 * P[other].PI+1]);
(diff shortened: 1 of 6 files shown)


rainwoodman:
The shorter BH step indeed did not bring much of an improvement in stability. I think I added it when trying to track down crazy growth of BH near BHMass ~ 1.

libgadget/blackhole.c
```diff
@@ -47,12 +46,11 @@ typedef struct {
     MyFloat BH_MinPotPos[3];
     MyFloat BH_MinPot;
 
-    int BH_TimeBinLimit;
+    short int BH_minTimeBin;
     MyFloat FeedbackWeightSum;
 
     MyFloat Rho;
     MyFloat SmoothedEntropy;
-    MyFloat SmoothedPressure;
```
rainwoodman: I recall pressure was used to testing if pressure makes more sense with pressure-entropy. The verdict was it didn't matter. So yes for removal.

libgadget/blackhole.c
```diff
@@ -263,7 +273,15 @@ blackhole(ForceTree * tree)
     SphP_scratch->Injected_BH_Energy = mymalloc2("Injected_BH_Energy", SlotsManager->info[0].size * sizeof(MyFloat));
```
rainwoodman: Does it have to be mymalloc2?

sbird: It does, unfortunately. Because we need to free SwallowID around it.

rainwoodman: nevermind. you didn't change this line.

libgadget/slotsmanager.h
```diff
@@ -41,24 +41,18 @@ struct bh_particle_data {
...
     int JumpToMinPot;
     double  MinPotPos[3];
 
     MyIDType SwallowID; /* Allows marking of a merging particle. Used only in blackhole.c.
                            Set to -1 in init.c and only reinitialised if a merger takes place.*/
 
-    short int TimeBinLimit;
+    /* Stores the minimum timebins of all black hole neighbours.
+     * The black hole timebin is then set to this.*/
+    short int minTimeBin;
```
rainwoodman: Min? since we have MinPotPos.

libgadget/slotsmanager.h
```diff
@@ -41,24 +41,18 @@ struct bh_particle_data {
...
-    MyFloat Pressure;
-    MyFloat SurroundingGasVel[3];
-    MyFloat FormationTime;		/*!< formation time of black hole. */
-
-    MyFloat accreted_Mass;
-    MyFloat accreted_BHMass;
-    MyFloat accreted_momentum[3];
+    MyFloat FormationTime;  /*!< formation time of black hole. */
 
     int JumpToMinPot;
```
rainwoodman: Is this still used?

libgadget/blackhole.c
```diff
@@ -371,11 +389,11 @@ blackhole_accretion_preprocess(int n, TreeWalk * tw)
 static void
 blackhole_feedback_postprocess(int n, TreeWalk * tw)
 {
-    if(BHP(n).accreted_Mass > 0)
+    int PI = P[n].PI;
+    if(BH_GET_PRIV(tw)->BH_accreted_Mass[PI] > 0)
     {
-        P[n].Mass += BHP(n).accreted_Mass;
-        BHP(n).Mass += BHP(n).accreted_BHMass;
-        BHP(n).accreted_Mass = 0;
```
rainwoodman: No longer need this because we reallocate accreted_Mass every step the BH is active.

However, I am a bit worried if we do accumulate this variable cross steps in corner cases.
So far it looks like that's impossible to happen to me. Did you think this through?

rainwoodman:
Looks good over all. A few minor comments. Thanks!

libgadget/slotsmanager.h
sbird: We do - but some of the other timebin variables have a lower case m, I think.

libgadget/blackhole.c
sbird: Ok!

libgadget/slotsmanager.h
sbird: It is never actually used, but it is written to disc - could be useful for some post-processing analyses.

libgadget/blackhole.c
sbird: I think it is impossible to accumulate it - TREEWALK_REDUCE erases the previous value, so any time the particle is active it will be accumulated anew.

sbird:
Thanks for the quick review! I will merge this one.

PR #336, opened 2019-08-22
Fix black holes and the tree

sbird:
This PR: 
1) Makes it easier to debug a failure of the force tree to build
2) Makes the first stage of the force tree building non-recursive, so that we don't have a stack overflow if multiple particles are in the same spot.
3) Removes the sound speed check for the black hole mergers. This is actually a very bad idea, but we never noticed before because with repositioning broken there were no mergers anyway!
What happens goes like this:

1. BH moves to halo center
2. Merger happens
3. N+1 BHs move to halo
4. These BHs have different velocities so fail the sound speed check and do not merge. They are not dissipative so have no way for the velocities to reduce.
if N < 11:
    Goto 1.
if N > 11:
    Too many particles in one place for the tree to build!
    goto 5.

5. The function which creates new tree nodes recurses infinitely and overflows the stack / crashes with out of memory! Often the crash dump is never reached so you do not even know why.
    
This may also be affecting bt-ii, making it really really slow: it creates very deep trees with clumps of identically positioned black holes.

4) Fixes an extremely rare bug when building the force tree: pseudo-particle hosts could be removed which caused problems.

Please review! I would like to merge quickly, since the bugs are serious.

### DIFF
--- a/libgadget/forcetree.c
+++ b/libgadget/forcetree.c
@@ -285,8 +285,16 @@ create_new_node_layer(int parent, int p_toplace,
         /* Get memory for an extra node from our cache.*/
         nprnt->u.s.suns[i] = get_freenode(nnext, nc);
         /*If we already have too many nodes, exit loop.*/
-        if(nc->nnext_thread >= tb.lastnode)
+        if(nc->nnext_thread >= tb.lastnode) {
+            /* This means that we have > NMAXCHILD particles in the same place,
+             * which usually indicates a bug in the particle evolution. Print some helpful debug information.*/
+            message(1, "Failed placing %d at %g %g %g, type %d, ID %ld. Others were %d (%g %g %g, t %d ID %ld) and %d (%g %g %g, t %d ID %ld).\n",
+                p_toplace, P[p_toplace].Pos[0], P[p_toplace].Pos[1], P[p_toplace].Pos[2], P[p_toplace].Type, P[p_toplace].ID,
+                oldsuns[0], P[oldsuns[0]].Pos[0], P[oldsuns[0]].Pos[1], P[oldsuns[0]].Pos[2], P[oldsuns[0]].Type, P[oldsuns[0]].ID,
+                oldsuns[1], P[oldsuns[1]].Pos[0], P[oldsuns[1]].Pos[1], P[oldsuns[1]].Pos[2], P[oldsuns[1]].Type, P[oldsuns[1]].ID
+            );
             return 1;
+        }
         struct NODE *nfreep = &tb.Nodes[nprnt->u.s.suns[i]];
         /* We create a new leaf node.*/
         init_internal_node(nfreep, nprnt, i);
--- a/libgadget/forcetree.c
+++ b/libgadget/forcetree.c
@@ -791,6 +791,11 @@ force_update_particle_node(int no, int sib, const ForceTree * tree, const int Hy
 static int
 force_update_node_recursive(int no, int sib, int level, const ForceTree * tree, const int HybridNuGrav)
 {
+#ifdef DEBUG
+    if(tree->Nodes[no].f.ChildType != NODE_NODE_TYPE)
+        endrun(3, "force_update_node_recursive called on node %d of type %d != %d!\n", no, tree->Nodes[no].f.ChildType, NODE_NODE_TYPE);
+#endif
+
     /*Last value of tails is the return value of this function*/
     int j, suns[8], tails[8];
 
@@ -1042,6 +1047,11 @@ void force_treeupdate_pseudos(int no, const ForceTree * tree)
         /*This may not happen as we are an internal top level node*/
         if(p < tree->firstnode || p >= tree->lastnode)
             endrun(6767, "Updating pseudos: %d -> %d which is not an internal node between %d and %d.",no, p, tree->firstnode, tree->lastnode);
+#ifdef DEBUG
+        /* Check we don't move to another part of the tree*/
+        if(tree->Nodes[p].father != no)
+            endrun(6767, "Tried to update toplevel node %d with parent %d != expected %d\n", p, tree->Nodes[p].father, no);
+#endif
 
         if(tree->Nodes[p].f.InternalTopLevel)
             force_treeupdate_pseudos(p, tree);
--- a/libgadget/forcetree.c
+++ b/libgadget/forcetree.c
@@ -145,7 +145,7 @@ ForceTree force_tree_build(int npart, DomainDecomp * ddecomp, const double BoxSi
             force_tree_free(&tree);
             message(1, "TreeAllocFactor from %g to %g\n", ForceTreeParams.TreeAllocFactor, ForceTreeParams.TreeAllocFactor*1.15);
             ForceTreeParams.TreeAllocFactor *= 1.15;
-            if(ForceTreeParams.TreeAllocFactor > 5.0) {
+            if(ForceTreeParams.TreeAllocFactor > 3.0) {
                 TooManyNodes = 1;
                 break;
             }
--- a/libgadget/forcetree.c
+++ b/libgadget/forcetree.c
@@ -808,10 +808,18 @@ force_update_node_recursive(int no, int sib, int level, const ForceTree * tree, 
      * This sharply reduces the size of the tree.
      * Also count the node children for thread balancing.*/
     for(j=0; j < 8; j++) {
-        /* Pseudo nodes may have zero occupation*/
-        if(tree->Nodes[suns[j]].f.ChildType == PARTICLE_NODE_TYPE &&
-            tree->Nodes[suns[j]].u.s.noccupied == 0)
+        /* Never remove empty top-level nodes so we don't
+         * mess up the pseudo-data exchange.
+         * This may happen for a pseudo particle host or, in very rare cases,
+         * when one of the local domains is empty. */
+        if(!tree->Nodes[suns[j]].f.TopLevel &&
+            tree->Nodes[suns[j]].f.ChildType == PARTICLE_NODE_TYPE &&
+            tree->Nodes[suns[j]].u.s.noccupied == 0) {
+                /* In principle the removed node will be
+                 * disconnected and never used, but zero it just in case.*/
+                force_zero_union(&tree->Nodes[suns[j]]);
                 suns[j] = -1;
+        }
         else if(tree->Nodes[suns[j]].f.ChildType == NODE_NODE_TYPE)
             childcnt++;
     }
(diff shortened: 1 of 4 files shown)


libgadget/forcetree.c
```diff
@@ -268,73 +268,85 @@ modify_internal_node(int parent, int subnode, int p_toplace, const ForceTree tb)
...
-                oldsuns[1], P[oldsuns[1]].Pos[0], P[oldsuns[1]].Pos[1], P[oldsuns[1]].Pos[2], P[oldsuns[1]].Type, P[oldsuns[1]].ID
-            );
-            return 1;
+    /* This is so we can defer changing
+     * the type of the existing node until the end.*/
+    int parent = firstparent;
+
+    do {
+        int i;
+        int oldsuns[NMAXCHILD];
```
rainwoodman: This is still allocated on the stack -- can you lift it out?

rainwoodman: Nevermind! Actually it is freed at the end of the loop.

libgadget/forcetree.c
```diff
@@ -268,73 +268,85 @@ modify_internal_node(int parent, int subnode, int p_toplace, const ForceTree tb)
...
-    if(nchild->u.s.noccupied < NMAXCHILD) {
-        modify_internal_node(child, nchild->u.s.noccupied, p_toplace, tb);
-        nchild->u.s.noccupied++;
-    }
-    else {
-        int ret = create_new_node_layer(child, p_toplace, tb, nnext, nc);
-        if(ret > 0)
-            return ret;
-    }
+        if(nchild->u.s.noccupied < NMAXCHILD) {
```
rainwoodman: I don't know if you want to mess with it more -- the current code does what it says. but
This logic is somewhat duplicated with line 480, which makes me wonder if the assertion on creating a new layer of nodes was only required to ease the recursion.

A simpler invariance is to assert that after a function call particle at to_place is put into the subtree starting at first_parent, and the new starting point of the insertion is returned.

libgadget/forcetree.c
```diff
@@ -808,10 +808,18 @@ force_update_node_recursive(int no, int sib, int level, const ForceTree * tree,
...
+        /* Never remove empty top-level nodes so we don't
+         * mess up the pseudo-data exchange.
+         * This may happen for a pseudo particle host or, in very rare cases,
+         * when one of the local domains is empty. */
+        if(!tree->Nodes[suns[j]].f.TopLevel &&
+            tree->Nodes[suns[j]].f.ChildType == PARTICLE_NODE_TYPE &&
+            tree->Nodes[suns[j]].u.s.noccupied == 0) {
+                /* In principle the removed node will be
+                 * disconnected and never used, but zero it just in case.*/
+                force_zero_union(&tree->Nodes[suns[j]]);
```
rainwoodman: Zero it or fill it with strange numbers?

sbird: Zero is better because strange numbers can affect other tree moments without it being obvious where they come from.

libgadget/forcetree.c
sbird: I see what you mean, but 480 is not necessarily a duplicate, because different threads may be walking the same subtree. The loop is useful because it takes place before the parent node is marked and unlocked, so it can happen without atomic accesses.

sbird:
A test using the hydro example with 128^3 particles found that at  z ~ 4 the number of black holes were about doubled: every single one of them was having a prevented merger. The csound check almost always fails. This was also true for the repositioning, as Yueying found: if the csound check applies then the black holes drift from the center. I think that the gas around the black hole is not a hot blob of gas but a cold star forming blob of gas: recall that star-forming gas is not heated by AGN feedback (it is in fact cooled). There is no sound speed check when accreting gas particles and never has been: a high gas velocity reduces the eddington limited accretion of BHMass, but it does not affect the probability of a gas particle being absorbed. See line 501 of blackhole.c

sbird:
It is hard to check the black hole mass function at z~0 without this change because the code never gets that far...

libgadget/forcetree.c
rainwoodman: Loop at 422? Yes I think we talked about that loop -- it indeed happens before locking. 

But line 480 we already have the lock on the tree branch rooting at `this`, so I think merging the logic from 480 to 486 into the function shouldn't break threading?

rainwoodman:
Oh. I was about the catch up with the email thread. It is a bit worrying that the gas near the BH is so cold.

Recall that at one point we were looking at the artificial cap on gas particle temperature due to BH feedback? When the BH has grown, the temperature must have been very large. But in less massive BH the feedback from both BH and SFR are probably not caught up with cooling, and thus the sound speed check fails.

But why do we have two BHs in small halos? Is it really that often that we have halo mergers, (strange) or is it because BHs drift in and out of the least massive FOF halos too frequently (since we seed a new BH if the previous one is no longer in the FOF).

Now relate this to the weirdness of the FOF finder that I have been suspecting -- A bug in FOF that misses BH randomly from halos can surely create many BHs in the same halo. And then it trickles down to this problem. If before writing snapshots we run FOF without seeding BH, then look for the fraction of halos without BH, it could be a useful indicator.

In any case yes, disabling the sound speed check will improve the stability of less massive halos, but at the cost of may artificially merge the BH of a bullet galaxy, when the bullet is shooting directly into the center of the more massive system -- which can be a very valid case if you simulate isolated system of two galaxies colliding each other, but I suppose it not going to happen too frequently in cosmology runs?

Tagging @yueyingn

yueyingn:
The "two BHs in small halo" issue is indeed caused by the failure of soundspeed check for BH merger. I did a test run by disabling the cs check for BH merger in bhdrift branch and got rid of those overlap BHs.

The merger event is indeed rare --- we only have 300 such events out of 300000 BH-containing halos in BT at z=7. We spotted this issue in smaller box run at lower redshift (z<5).

sbird:
Thanks @yueyingn! It sounds like we should fix this, but it is probably not causing the catastrophic problems of BT-ii, just because there haven't been enough mergers yet. If we care about the idealised case (at the moment we only support periodic boundary conditions, so we can't simulate them), we should re-implement a different condition. I think the physics of using the sound speed for the 'isolated bullets' check is wrong: black holes merge by GW emission whenever they become gravitationally bound. So we should compare check the relative velocity of the two black holes to a constant, rather than a gas physics related quantity.

I'm out of town again this week so I won't be at the next telecon, but I'll be back the week after.

libgadget/forcetree.c
sbird: Probably not - but I think I'd prefer to leave it in there, if that's ok. It doesn't hurt, and the code is already a bit confusing.

sbird:
Let us merge this master PR and talk more about the similar fix in bt-ii on the telecon. The FOF stuff should definitely be checked (ideally we could have a FOF test) but I don't have time to work on it at the moment, sorry.

sbird:
Also the black holes in a bullet cluster type event would only merge if the potential minima were in the same place. Which either means a direct head-on collision or that they are actually bound. I think no velocity check will be fine in practice in a cosmology simulation.

rainwoodman:
I disagree.

If one of the potential well is sufficiently steep, then the shallower
potential well can be smoothed out, becoming a kink in the deeper well.
Since there is now only one minium, both BH flow towards the 'global'
minimum. As the galaxies part, the shallower potential well moves far away,
and becomes a minimum again.

Currently, the speed of the reposition flow is non-physical (proportional
to local particle density and time step size), it can happen that the bh
originated from the shallower well moves too rapidly, and arrives to the
global minimum faster than the shallower well regains its independence.

Llast Wednesday, Yueying showed us a picture catching this in action in one
of her simulations -- the galaxies penetrated each other, but the bhs are
stuck in the same galaxy, leaving the other galaxy without a bh.

This suggests that if instead we only change the direction of the BH,
keeping the velocity magnitude -- then the reposition flow is a physical
velocity and will track the binding energy better -- then no need to add a
strange sound speed condition?

sbird:
We would still need a way to slow the BH down: if we only change the velocity direction then it will move thermally around the galactic center. This paper: https://arxiv.org/abs/1607.02151 uses an approach with an effective extra force towards the potential minimum, which is substantially more physical but I'm not sure if it would work for our relatively low resolution.

We could also add a velocity cut (probably a probabilistic one) based not on gas velocity but on relative BH velocity.

@yueyingn : Can you post the image here (or email me)? Was it a cosmological simulation?

sbird:
Also see here: https://mtremmel.github.io/research/romulus.html - the dynamical friction is the thing we are interested in.

sbird:
Basically what they do is to add an acceleration towards the center of mass within the smoothing kernel. See section 5.3 of the paper. See also 1501.07609 Calling it fully realistic is perhaps an exaggeration, because the dynamical friction arises from stars *and dark matter* but it is definitely the most physical way of doing this that I know of, and the only one that really fixes the isolated merger problem.

yueyingn:
Hi Simeon,

Here is an event I caught in BT simulation (there must be more, I just
caught this one randomly) of two merging galaxies. When the two center BHs
pass by each other (within smoothing kernel), one got repositioned and
trapped in another's minimum potential while left the host galaxy with no
BH at z=7 (and were not able to merge because of the sound speed check). I
also attached the zoom-in plot at z=7 to show that they are indeed
dynamically separate galaxies at z=7.

sbird:
Hmm, ok. So we do need a further fix. Option a) is to implement dynamical friction,  option b) is a weaker but simpler "is this bound" check.

sbird:
Thinking it over, option b) will not work: it is not the merging that is a problem but the repositioning. If we slow the merger down we will just end up with two black holes in the wrong potential minimum, staring at each other. We need to implement Tremmel's model.

rainwoodman:
Yes. The problem is in repositioning. What if we instead reposition only if 'is this bound" is true? As a more physical version of the sound speed based reposition condition. This is saying in a bullet event where two fluids doesn't fully mix, each BH will only flow with its fluid.

Notice this chart: https://mtremmel.github.io/research/img/dynamics/DFmodel_compare.png

The redline is repositioning. The blue line is dynamical friction. Not sure about the halo mass -- this suggests we need a recalibration the the feedback model -- the BH is now more often 1 Kpc/h away from the center rather than directly sitting at the center.

Finally: on this week's telecon, Tiziana says the non-mergers are rare, and no need to worry too much about them.

sbird:
Yes, that could be a good idea. We would not reposition if the relative velocity between the BH and the potential minimum was above a fixed threshold. Note that in that image the dashed black line is the softening length, so you would not expect the BH to be closer than of order 1 kpc. 

Did Tiziana mean that we should not worry too much about bullet cluster like events and thus the current code is fine? So we should merge the bt-ii equivalent of this PR?

PR #338, opened 2019-09-11
Impifix

sbird:
Some defensive programming here. Don't use MPI_IN_PLACE with arrays and MPI_MAX: intel MPI doesn't like it (it will return 0). Also make the initial allocation larger than 0, just in case.

### DIFF
--- a/libgadget/petaio.c
+++ b/libgadget/petaio.c
@@ -200,7 +200,6 @@ void petaio_read_internal(char * fname, int ic, MPI_Comm Comm) {
         int64_t end = (ThisTask + 1) * NTotal[ptype] / NTask;
         NLocal[ptype] = end - start;
         PartManager->NumPart += NLocal[ptype];
-
     }
 
     /* Allocate enough memory for stars and black holes.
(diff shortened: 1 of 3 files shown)


sbird:
With this one I can finally run on frontera with <=1024 MPI ranks and threads.

libgadget/petaio.c
```diff
@@ -210,21 +209,18 @@ void petaio_read_internal(char * fname, int ic, MPI_Comm Comm) {
...
-            newSlots[ptype] = All.PartAllocFactor * NLocal[ptype];
-        }
-    }
-
     /* Now allocate memory for the secondary particle data arrays.
      * This may be dynamically resized later!*/
 
     /*Ensure all processors have initially the same number of particle slots*/
-    MPI_Allreduce(MPI_IN_PLACE, newSlots, 6, MPI_INT, MPI_MAX, Comm);
+    int newSlots[6] = {0};
```
rainwoodman: Could you add a comment refering to impi version and this ticket, so we don't roll this back accidentally?

sbird: No ticket, but I added a comment with the version

PR #340, opened 2019-09-12
Add a random offset to particle positions, changing every PM step

sbird:
This shifts the periodic box by a smallish random number to avoid inter-timestep correlations in the tree opening condition. The public Arepo release does something similar, but we use a much smaller shift so we don't move transfer all particles between processors every full domain decomp. This appears to actually increase speed slightly (!) by reducing the short range gravity time but I suspect this of being a coincidence in this specific box.

### DIFF
--- a/libgadget/drift.c
+++ b/libgadget/drift.c
@@ -35,7 +35,9 @@ void drift_particle(int i, inttime_t ti1, struct SpinLocks * spin) {
 /* Drifts an individual particle to time ti1, by a drift factor ddrift.
  * The final argument is a random shift vector applied uniformly to all particles before periodic wrapping.
  * The box is periodic, so this does not affect real physics, but it avoids correlated errors
- * in the tree expansion.
+ * in the tree expansion. The shift is relative to the current position. On most timesteps this is zero,
+ * signifying no change in the coordinate frame. On PM steps a random offset is generated, and the routine
+ * receives a shift vector removing the previous random shift and adding a new one.
  * This function also updates the velocity and updates the density according to an adiabatic factor.
  */
 static void real_drift_particle(int i, inttime_t ti1, const double ddrift, const double random_shift[3])
--- a/libgadget/drift.c
+++ b/libgadget/drift.c
@@ -14,24 +14,31 @@
 
 #define MAXHSML 30000.0
 
-
-static void real_drift_particle(int i, inttime_t ti1, const double ddrift);
+static void
+real_drift_particle(int i, inttime_t ti1, const double ddrift, const double random_shift[3]);
 
 /* Updates a single particle to the current drift time*/
 void drift_particle(int i, inttime_t ti1, struct SpinLocks * spin) {
     if(P[i].Ti_drift == ti1) return;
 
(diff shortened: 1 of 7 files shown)


rainwoodman:
Yes and using the global in snapshot getter is scary. The statement thst
there is an global that is set only if .... is also implying hard to find
bug down the road.

A transformation there is like saying we have encoded the particle
positions differently. However anywhere else we actually use this shifted
position without decoding it.

But if there is no better way.. then so be it. Let me poke with this over
the weekend.

sbird:
Right. It is a bit messy. The other thing I considered was adding an offset to the PEANO key instead, since we only really care about the tree. But since we use the Pos directly to build the tree that doesn't work so well.

sbird:
@rainwoodman  What are your thoughts?

libgadget/run.c
```diff
@@ -167,8 +167,36 @@ void run(DomainDecomp * ddecomp)
                 endrun(0, "Human triggered termination.\n");
             }
         }
+
+        double new_random_shift[3] = {0};
```
rainwoodman: rel_random_shift.

libgadget/run.c
```diff
@@ -167,8 +167,36 @@ void run(DomainDecomp * ddecomp)
...
+        if(NumCurrentTiStep > 0 && is_PM  && All.RandomParticleOffset > 0) {
+            int i;
```
rainwoodman: Perhaps move this to a function (173-196).
UpdateRandomOffset(All.RandomParticleOffset, All.CurrentParticleOffset, rel_random_shift);

libgadget/drift.c
```diff
@@ -14,24 +14,31 @@
...
 }
 
-static void real_drift_particle(int i, inttime_t ti1, const double ddrift)
+/* Drifts an individual particle to time ti1, by a drift factor ddrift.
+ * The final argument is a random shift vector applied uniformly to all particles before periodic wrapping.
+ * The box is periodic, so this does not affect real physics, but it avoids correlated errors
+ * in the tree expansion.
+ * This function also updates the velocity and updates the density according to an adiabatic factor.
+ */
+static void real_drift_particle(int i, inttime_t ti1, const double ddrift, const double random_shift[3])
```
rainwoodman: The drift function is an unfortunate mix of several physical processes. 

The reason random_shift has to be plumbed into this function is because we update PEANO key in this function (presumably for better cache coherency), and also because we do the BH reposition in this function. 

Could you add a comment that describes random_shift [3], e.g. 

'In certain steps (currently PM steps) we can add a small random_shift to the position displacement to enhance the accuracy of tree force calculation. The shift is a relative shift between the desired shift in the last step and that of the current shift.' 
my language may be a bit loose, but something like this?

rainwoodman:
I don't think there is a better way. drift couples several processes into the same function, and it has to be that way. 

I added a few new requests about describing how this works. Looks like real_drift_particles expects a relative offset but the variable in All stores an absolute offset?

sbird:
ok, I will merge when travis goes green

PR #342, opened 2019-09-19
Two small fixes

sbird:
Two small fixes.

### DIFF
--- a/libgadget/blackhole.c
+++ b/libgadget/blackhole.c
@@ -291,6 +291,7 @@ blackhole(ForceTree * tree)
     message(0, "Accretion done: %d gas particles swallowed, %d BH particles swallowed\n",
                 Ntot_gas_swallowed, Ntot_BH_swallowed);
 
+    int total_bh;
     double total_mdoteddington;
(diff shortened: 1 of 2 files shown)


libgadget/blackhole.c
```diff
@@ -320,7 +322,7 @@ blackhole(ForceTree * tree)
                     (0.1 * LIGHTCGS * LIGHTCGS * THOMPSON)) * All.UnitTime_in_s);
 
         fprintf(FdBlackHoles, "%g %d %g %g %g %g\n",
-                All.Time, SlotsManager->info[5].size, total_mass_holes, total_mdot, mdot_in_msun_per_year, total_mdoteddington);
+                All.Time, total_bh, total_mass_holes, total_mdot, mdot_in_msun_per_year, total_mdoteddington);
```
rainwoodman: It would be great if we can add a file header every now and then  -- (numpy access # comment lines). A file with just numbers is a bit confusing.. But it may choke other downstream parsers.

sbird: hah, yes, a good idea. I don't know of any downstream parsers for these files

PR #344, opened 2019-09-25
Longtimeline

sbird:
We ran out of timebins. Make the inttime variable 64 bit so this doesn't happen.

### DIFF
--- a/libgadget/timebinmgr.c
+++ b/libgadget/timebinmgr.c
@@ -166,8 +166,7 @@ loga_from_ti(inttime_t ti)
 inttime_t
 ti_from_loga(double loga)
 {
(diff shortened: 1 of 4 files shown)


rainwoodman:
Not sure if the travis failure is related. But I would strongly suggest adding a test cases for bin size of 1 in timebinmgr -- will it underflow doubles or cause numerical issues with log?

sbird:
I could not debug this sufficiently, closing for now.

rainwoodman:
If we need more than 100K time steps in a sim, then it will unlikely ever
finish anyways.

sbird:
It is true.

PR #345, opened 2019-09-26
Debugoutput

sbird:
This PR saves a variety of extra fields (all fields in P and SPHP) to the crash dump snapshot, and optionally to the main snapshot. It also saves and restores DelayTime, which we weren't doing before. This should be enough to debug future "bad timestep" crashes.

### DIFF
--- a/libgadget/petaio.c
+++ b/libgadget/petaio.c
@@ -809,55 +809,55 @@ void register_io_blocks(struct IOTable * IOTable) {
     IOTable->ent = mymalloc("IOTable", IOTable->allocated* sizeof(IOTableEntry));
     /* Bare Bone Gravity*/
     for(i = 0; i < 6; i ++) {
-        IO_REG(Position, "f8", 3, i);
-        IO_REG(Velocity, "f4", 3, i);
-        IO_REG(Mass,     "f4", 1, i);
-        IO_REG(ID,       "u8", 1, i);
+        IO_REG(Position, "f8", 3, i, IOTable);
+        IO_REG(Velocity, "f4", 3, i, IOTable);
+        IO_REG(Mass,     "f4", 1, i, IOTable);
+        IO_REG(ID,       "u8", 1, i, IOTable);
         if(All.OutputPotential)
-            IO_REG_WRONLY(Potential, "f4", 1, i);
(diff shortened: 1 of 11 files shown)


libgadget/petaio.c
```diff
@@ -856,10 +859,47 @@ static void register_io_blocks() {
...
+    IO_REG_WRONLY(DtEntropy,       "f4", 1, 0);
+    IO_REG_WRONLY(DhsmlEgyDensityFactor,       "f4", 1, 0);
+    IO_REG_WRONLY(DivVel,       "f4", 1, 0);
+    IO_REG_WRONLY(CurlVel,       "f4", 1, 0);
     /*Sort IO blocks so similar types are together; then ordered by the sequence they are declared. */
-    qsort_openmp(IOTable.ent, IOTable.used, sizeof(struct IOTableEntry), order_by_type);
+    qsort_openmp(IOTable->ent, IOTable->used, sizeof(struct IOTableEntry), order_by_type);
+}
+
+void free_io_blocks(struct IOTable * IOTable) {
```
rainwoodman: destroy_io_table(); or perhaps purge_io_table().

destroy: frees the members.
free: frees the object and the member.
purge: nothing to do with life cycle of the object.

libgadget/petaio.h
```diff
@@ -55,20 +67,21 @@ petaio_build_selection(int * selection,
  * IO_REG_WRONLY declares an io block which is written, but is not read on snapshot load.
  * */
 #define IO_REG(name, dtype, items, ptype) \
```
rainwoodman: Perhaps worth it to plumb IOTable up through these macros -- a bit of search and replacement, but avoids an hidden argument.

libgadget/petaio.c
sbird: Done!

libgadget/petaio.h
sbird: Yes, but it is a fair bit of careful search and replace. I don't really want to just now.

sbird:
@yueyingn : Can you re-run the crashing simulation with this PR included and find the low-timebin particles in the output?

libgadget/petaio.h
sbird: I did it anyway :-)

Issue #315, opened 2019-07-11
GenIC shall write the initial position to the file.


rainwoodman:

sbird:
The idea with this one is to write the initial, pre-displacement, particle positions to a file. The file would contain a homogeneous gas of particles. This is mostly useful with glassfiles: grids already contain the initial position encoded in the particle ID.

PR #347, opened 2019-10-01
Issue #315

maamari:
Added "PrePosition" to the ICP struct. PrePosition is set prior to displacement and saved into the bigfile (along with the position, velocity, etc.) when write_particle_data(...) is called.


genic/main.c
```diff
@@ -129,7 +129,12 @@ int main(int argc, char **argv)
     if(All2.MakeGlassGas || All2.MakeGlassCDM)
         glass_evolve(pm, 14, "powerspectrum-glass-tot", ICP, NumPartCDM+NumPartGas);
   }
-
+  
+  /*Write initial positions into ICP struct*/
+  for(int j=0; j<NumPartCDM; j++)
+      for(int k=0; k<3; k++)
```
sbird: House style is to avoid C99 inline for loop declarations, could you move the int j outside?

genic/main.c
```diff
@@ -129,7 +129,12 @@ int main(int argc, char **argv)
     if(All2.MakeGlassGas || All2.MakeGlassCDM)
         glass_evolve(pm, 14, "powerspectrum-glass-tot", ICP, NumPartCDM+NumPartGas);
   }
-
+  
+  /*Write initial positions into ICP struct*/
+  for(int j=0; j<NumPartCDM; j++)
```
sbird: I think you want NumPartCDM + NumPartGas here? Also does this need to be set for neutrino particles as well?

sbird:
Look nice, thanks!

PR #348, opened 2019-10-07
Gravity cleanups

sbird:
This PR does numerous cleanups and refactorings in the gravity code to support using runtests as a force accuracy test, and for an upcoming unit test for the gravity force. Should not change output, but I will test. Please review.


sbird:
This version is tested to give the same answer to within floating point accuracy. It gives exactly the same answer apart for commit 5f680b5202baaa7ca14ea755a3b67f69a24f52b4 which changes the order of floating point operations. The force-push just reorders things so that all intermediate commits compile properly.

libgadget/domain.h
```diff
@@ -43,8 +43,25 @@ typedef struct DomainDecomp {
...
+    /** Initial number of Top level tree nodes as a fraction of particles */
+    double TopNodeAllocFactor;
+    /** Fraction of local particle slots to leave free for, eg, star formation*/
+    double SetAsideFactor;
+};
+
 /*Set the parameters of the domain module*/
 void set_domain_params(ParameterSet * ps);
+/* Test helper*/
+void set_domain_par(struct DomainParams dp);
```
rainwoodman: If a struct is in the public header, suggest typedef it.

libgadget/gravpm.c
```diff
@@ -43,34 +39,40 @@ static PetaPMFunctions functions [] =
 
 static PetaPMGlobalFunctions global_functions = {NULL, NULL, potential_transfer};
 
-static PetaPM pm[1];
 static PetaPMRegion * _prepare(PetaPM * pm, void * userdata, int * Nregions);
 
-void gravpm_init_periodic() {
-    petapm_init(pm, All.BoxSize, All.Nmesh, MPI_COMM_WORLD);
+PetaPM
+gravpm_init_periodic(double BoxSize, double Asmth, int Nmesh) {
```
rainwoodman: void gravpm_init_periodic(PetaPM *, ...)

Instead of returning a copy of PetaPM object -- consistency with existing calling conventions.

libgadget/gravpm.c
```diff
@@ -43,34 +39,40 @@ static PetaPMFunctions functions [] =
...
+void
+gravpm_force(PetaPM * pm, ForceTree * tree) {
     PetaPMParticleStruct pstruct = {
         P,
         sizeof(P[0]),
         (char*) &P[0].Pos[0]  - (char*) P,
         (char*) &P[0].Mass  - (char*) P,
         (char*) &P[0].RegionInd - (char*) P,
-        (All.HybridNeutrinosOn ? &hybrid_nu_gravpm_is_active : NULL),
+        NULL,
```
rainwoodman: Add a comment to indicate this field is `set below`.

libgadget/gravshort-tree.c
```diff
@@ -27,29 +26,71 @@
...
+void
+grav_short_tree(const ActiveParticles * act, PetaPM * pm, ForceTree * tree, double G, double rho0, int NeutrinoTracer, int FastParticleType)
 {
     double timeall = 0;
     double timetree, timewait, timecomm;
-    if(!All.TreeGravOn)
-        return;
 
     TreeWalk tw[1] = {{0}};
+    struct GravShortPriv priv;
```
rainwoodman: I know this is OK because tw doesn't live outside of this functio.
It still feels there shall be an _init function and an _destroy function that maintain the life cycle of priv. (can probably move these initializers to the _init too?

libgadget/gravshort-tree.c
```diff
@@ -98,6 +140,10 @@ void grav_short_tree(ForceTree * tree)
 
     walltime_add("/Tree/Misc", timeall - (timetree + timewait + timecomm));
 
+    /* TreeUseBH > 1 means use the BH criterion on the initial timestep only,
+     * avoiding the fully open O(N^2) case.*/
+    if(TreeParams.TreeUseBH > 1)
```
rainwoodman: It's a bit confusing to mutate the Param during runtime.

libgadget/gravshort.h
```diff
@@ -31,21 +25,49 @@ typedef struct {
...
+    int FastParticleType;
+    /* Are neutrinos tracers? If so, exclude them from the tree force*/
+    int NeutrinoTracer;
+    /* Newton's constant in internal units*/
+    double G;
+    /* Matter density in internal units.
+     * rho_0 = Omega0 * rho_crit
+     * rho_crit = 3 H^2 /(8 pi G).
+     * This is (rho_0)^(1/3) */
+    double cbrtrho0;
```
rainwoodman: Does this include mass-like neutrino?

libgadget/petaio.c
```diff
@@ -44,10 +44,10 @@ void petaio_init(void) {
 }
 
 /* save a snapshot file */
-static void petaio_save_internal(char * fname, struct IOTable * IOTable);
+static void petaio_save_internal(char * fname, struct IOTable * IOTable, int print);
```
rainwoodman: by print, you mean verbose?

libgadget/powerspectrum.c
```diff
@@ -13,14 +13,15 @@
 /*Power spectrum related functions*/
 
 /*Allocate memory for the power spectrum*/
-void powerspectrum_alloc(struct _powerspectrum * PowerSpectrum, const int nbins, const int nthreads, const int MassiveNuLinResp, const double BoxSize_in_cm)
+void powerspectrum_alloc(Power * PowerSpectrum, const int nbins, const int nthreads, const int MassiveNuLinResp, const double BoxSize_in_cm)
```
rainwoodman: Consider PowerSpectrum -> ps, I've seen you using this in other files!

rainwoodman: Then you actually leave out the name for Power -> PowerSpectrum

libgadget/run.c
```diff
@@ -39,7 +39,7 @@ void init(int snapnum, DomainDecomp * ddecomp); /* init.c only used here */
  * reached, when a `stop' file is found in the output directory, or
  * when the simulation ends because we arrived at TimeMax.
  */
-static void compute_accelerations(int is_PM, int FirstStep, int GasEnabled, int HybridNuGrav, ForceTree * tree, DomainDecomp * ddecomp);
+static void compute_accelerations(const ActiveParticles * act, int is_PM, PetaPM * pm, int FirstStep, int GasEnabled, int HybridNuGrav, ForceTree * tree, DomainDecomp * ddecomp);
```
rainwoodman: If the step is not PM, can we pass in pm == NULL instead?

libgadget/run.c
```diff
@@ -359,11 +363,11 @@ void compute_accelerations(int is_PM, int FirstStep, int GasEnabled, int HybridN
...
-    grav_short_tree(tree);
-    /* TreeUseBH > 1 means use the BH criterion on the initial timestep only,
-     * avoiding the fully open O(N^2) case.*/
-    if(All.TreeUseBH > 1)
-        All.TreeUseBH = 0;
+    const int NeutrinoTracer =  All.HybridNeutrinosOn && (All.Time <= All.HybridNuPartTime);
+    const double rho0 = All.CP.Omega0 * 3 * All.CP.Hubble * All.CP.Hubble / (8 * M_PI * All.G);
+
+    if(All.TreeGravOn)
+        grav_short_tree(act, pm, tree, All.G, rho0, NeutrinoTracer, All.FastParticleType);
```
rainwoodman: grav_short_tree requires pm.. That's a bit surprising, but understandable..

libgadget/runtests.c
```diff
@@ -32,45 +33,162 @@ void register_extra_blocks(struct IOTable * IOTable)
     }
 }
 
-void runtests(DomainDecomp * ddecomp)
+double copy_accns(double (* PairAccn)[3])
```
rainwoodman: The name says copy, but it computes a mean?

libgadget/runtests.c
```diff
@@ -32,45 +33,162 @@ void register_extra_blocks(struct IOTable * IOTable)
...
+
+    treeacc = origtreeacc;
+    set_gravshort_treepar(treeacc);
+    /* Code automatically sets the UseTreeBH parameter.*/
+    grav_short_tree(&Act, &pm, &Tree, All.G, rho0, 0, All.FastParticleType);
+    grav_short_tree(&Act, &pm, &Tree, All.G, rho0, 0, All.FastParticleType);
+
+    petaio_save_snapshot(&IOTable, 0, "%s/PART-tree-%03d", All.OutputDir, RestartSnapNum);
+
+    check_accns(&meanerr,&maxerr,PairAccn, meanacc);
```
rainwoodman: ,

sbird:
By the way, is there a reason why the PetaPM code always gives zero force for all zero mass particles?
I am trying to make a plot like in the Gadget-2 paper, using a single massive particle and a number of massless test particles, but GravPM is always coming out zero.

sbird:
I think it might be discovering that there is no mass in most regions of the box, then compressing the pencils to avoid those cells.

libgadget/timestep.c
```diff
@@ -30,14 +30,11 @@ static inline inttime_t dti_from_timebin(int bin) {
     /*Casts to work around bug in intel compiler 18.0*/
     return bin > 0 ? (1u << (unsigned) bin) : 0;
 }
-/*Flat array containing all active particles*/
-int NumActiveParticle;
-int *ActiveParticle;
 
-static inline int get_active_particle(int pa)
+static inline int get_active_particle(const ActiveParticles * act, int pa)
```
rainwoodman: This is an index in P? Do we have an acronym for particle index in P?

libgenic/pmesh.h
```diff
@@ -16,10 +16,10 @@ typedef struct {
         ptrdiff_t total;
     } ORegion;
     int Nmesh[3];
-} PM;
+} PMField;
```
rainwoodman: What about PMDef or PMDescriptor / PMDesc? -- PMField smells like an array of double that represents a 3d field.

rainwoodman:
I did a `quick` walk through and had a few styling comments. Didn't notice potential issues -- but I didn't look very hard. Most transformations are plumbing global variables to the stack, good move!

An interesting perspective: density, grav_short, .. can be thought of as member functions of the ActiveParticleSet.

libgadget/gravshort-tree.c
sbird: mm, it is overhead for only one use.

libgadget/gravshort-tree.c
sbird: True. But I could not think of a cleaner way to do it.

libgadget/gravshort.h
sbird: Nope, not properly.

libgadget/run.c
sbird: I don't think so, we still need it to tell the short-range gravity what to do.

libgadget/run.c
sbird: I thought about it and it seems actually cleaner: it indicates that the short-range gravity depends on how the PM mesh is set up

libgadget/timestep.c
sbird: This is an index in act, but sometimes when act is p, it is the same thing

PR #349, opened 2019-10-10
Forcetest

sbird:
This includes the long-sought gravity test and increases the default gravitational accuracy of the code substantially. Not for merging just yet.


libgadget/tests/test_gravity.c
```diff
@@ -0,0 +1,364 @@
...
+ * Periodic boundary conditions are included by mirroring the box.*/
+static int force_direct(double ErrTolForceAcc)
+{
+    double * accn = (double *) mymalloc("accelerations", 3*sizeof(double) * PartManager->NumPart);
+    memset(accn, 0, 3 * sizeof(double) * PartManager->NumPart);
+
+    int xx, yy, zz;
+    int repeat = 1;
+    /* (slowly) compute gravitational force, accounting for periodicity by just inventing extra boxes on either side.*/
+    for(xx=-repeat; xx < repeat; xx++)
```
rainwoodman: Isn't this just including -1 and 0 -- is it large enough? The convergence is kind of slow if you are looking at potential, I think.

libgadget/tests/test_gravity.c
```diff
@@ -0,0 +1,364 @@
...
+        *suppmean/= (tot_npart*3.);
+    }
+    MPI_Allreduce(&meanforce, meangrav, 1, MPI_DOUBLE, MPI_SUM, MPI_COMM_WORLD);
+    *meangrav/= (tot_npart*3.);
+}
+
+/* This checks the force on each particle using a direct summation:
+ * very slow, but accurate.
+ * Periodic boundary conditions are included by mirroring the box.*/
+static int force_direct(double ErrTolForceAcc)
```
rainwoodman: There might be some difficulty computing force directly after particles are scattered -- but you can always gather them to the same rank and compute the force.

Do we have a routine that can sort particles in P by ID or alike? If there is a stable ordering, then you can literally compute the residual between direct force and TreePM and report the variance etc?

sbird: What do you mean by that? Isn't that what I do in check_accns?

libgadget/tests/test_gravity.c
sbird: Oops, yes. I changed this a few times, it should be larger.

sbird:
Ok, I just checked with some small simulations. I ran three simulations, with this branch ("forcetest"), with the master branch ("oldforce") and one with pairwise force, Rcut = 9 Asmth = 1.5 and Nmesh = 3xMIP, as an attempt at ground truth ("pairwise"). These were 300, 256^3 simulations, so not that much non-linear structure.

forcetest and pairwise differed by about 10^-3, with pairwise having a larger matter power. They both differed from oldforce by about 10^-2, which is consistent with what we were seeing before at this larger box size (with a smaller box we had a 5% increase in power with pairwise).

The computation time hit was about 25% in the tree, and a factor of 3.15 in the PM force (which is proportional to the increase in FFT size: 3^3 /2^3 ~ 3.25). The tree dominates (actually the SPH dominates) so the total time hit was about 30%. Not too bad. There were maybe 2% fewer timesteps, probably because the noise was reduced.

Overall, I think this is worth it, it looks like the force accuracy is very good with this patch.

sbird:
Oh, and there was no effect on the CDM/baryon relative offset, unfortunately.

libgadget/tests/test_gravity.c
rainwoodman: I see. check_accns computes the residual. 

Two questions remains:
1. how does the direct_force calculation work across ranks? Looks like the loop does not include particles from other ranks.
2. is is possible to rename force_direct to check_force_againast_direct, and lift the direct force calculation loop to a function force_direct?

sbird: 1. It does not - the test is serial only at the moment. This should be fixed, along with the density test (and perhaps the forcetree test), and I'll put that on my todo list. For this one I would indeed probably gather them to one rank.

2. Yes, I will do that!

PR #351, opened 2019-10-13
Nompilor

sbird:
This does a few cleanups and might help out MPI.

### DIFF
--- a/gadget/main.c
+++ b/gadget/main.c
@@ -46,19 +46,15 @@ int main(int argc, char **argv)
 
     if(argc < 2)
     {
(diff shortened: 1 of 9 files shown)


sbird:
Found it! The last commit fixes the hang. But the cleanups should remain.

sbird:
I spend far too much time fixing issues that are really intel's fault

sbird:
Looks like 9d891 just makes it much less likely: but 45fb867 seems to do it so far.

sbird:
Gives bit-identical output, fixes the MPI hang. Merging.

PR #355, opened 2019-10-21
Issue 67

maamari:
This is only wrt BH particles. Wanted to make sure that this is all good before working on the gas particles. 

Mostly minor checks to ensure that swallowed particles are not interacting. The garbage business was not tampered outside of removing slots_mark_garbage() for BH particles. I figured I should leave the garbaging be until swallowed gas particles are also forbidden from interaction.


sbird:
Could you rebase on top of the master branch? You can do this with:

git pull
git rebase -i master
(drop the first four commits if they show up)

libgadget/timestep.c
```diff
@@ -55,7 +55,7 @@ timestep_eh_slots_fork(EIBase * event, void * userdata)
     int parent = ev->parent;
     int child = ev->child;
 
-    if(is_timebin_active(P[parent].TimeBin, All.Ti_Current)) {
+    if(is_timebin_active(P[parent].TimeBin, All.Ti_Current) && !P[parent].Swallowed) {
```
sbird: I think you don't need this: this hook fires if a new particle is created, but a new particle is never already swallowed.

libgadget/timestep.c
```diff
@@ -197,7 +197,7 @@ find_timesteps(inttime_t Ti_Current)
         {
             /* make sure the new step is currently active,
              * so that particles do not miss a step */
-            while(!is_timebin_active(bin, Ti_Current) && bin > binold && bin > 1)
+            while((!is_timebin_active(bin, Ti_Current) && bin > binold && bin > 1) || P[pa].Swallowed)
```
sbird: You don't need this because you already continue; on line 172. Worse, you are indexing with pa, which is an index into ActiveParticle, not P!

maamari:
Closing this and opening a new request for the same issue. Had to delete and refork so figure I'll start the request on a new slate as well.

PR #356, opened 2019-10-22
Fixes

sbird:
This PR fixes force_update_hmax, which was not computing the maximum across MPI tasks correctly. A warning in the fof code is also fixed.

### DIFF
--- a/libgadget/forcetree.c
+++ b/libgadget/forcetree.c
@@ -1109,34 +1109,16 @@ void force_treeupdate_pseudos(int no, const ForceTree * tree)
  *  out just before the hydrodynamical SPH forces are computed, i.e. after
  *  density().
  */
-void force_update_hmax(int * activeset, int size, ForceTree * tree)
+void force_update_hmax(int * activeset, int size, ForceTree * tree, DomainDecomp * ddecomp)
 {
-    int NTask, ThisTask;
+    int NTask, ThisTask, recvTask;
     int i, ta;
-    int *counts, *offsets;
-    struct dirty_node_data {
-        int treenode;
-        MyFloat hmax;
-    } * DirtyTopLevelNodes;
-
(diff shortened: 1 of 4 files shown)


sbird:
I confirm that this bug makes exactly zero difference to output! But I will fix it anyway.

rainwoodman:
Interesting -- In larger runs this might show up - where we run into other domains more frequently.

libgadget/forcetree.c
```diff
@@ -1149,74 +1131,57 @@ void force_update_hmax(int * activeset, int size, ForceTree * tree)
...
                 break;
-            }
-
+            tree->Nodes[no].u.d.hmax = P[p_i].Hsml;
             no = tree->Nodes[no].father;
         }
     }
 
-    /* share the hmax-data of the dirty nodes accross CPUs */
+    double * TopLeafhmax = (double *) mymalloc("TopLeafMoments", ddecomp->NTopLeaves * sizeof(double));
```
rainwoodman: The old algorithm only exchanges top leaves that are 'dirty'. Your new algorithm seems to be exchanging all TopLeaves. 

Do we already have MPI operations per step on the order of NTopLeaves? (We probably do, since the tree is built per step)

libgadget/forcetree.h
```diff
@@ -67,6 +66,8 @@ struct NODE
 typedef struct ForceTree {
     /*Is 1 if the tree is allocated. Only used inside force_tree_allocated() and when allocating.*/
     int tree_allocated_flag;
+    /* Flags that hmax has been computed for this tree*/
+    int hmax_computed_flag;
```
rainwoodman: This is only set but not used to branch operations?

sbird: Yes - I put it in for debugging only. I have some future patches that use it to decide to do the hmax, but they are not useful yet.

PR #357, opened 2019-10-23
Use MPI_Init_thread instead of MPI_Init

sbird:
On the recommendation of one of the Stampede admins

I think this will have little effect, because it seems most implementations are only affected if you use  MPI_THREAD_MULTIPLE, and their support is flaky at best, but it can't hurt.

### DIFF
--- a/gadget/main.c
+++ b/gadget/main.c
@@ -38,9 +38,12 @@ void gsl_handler (const char * reason, const char * file, int line, int gsl_errn
 int main(int argc, char **argv)
 {
     int NTask;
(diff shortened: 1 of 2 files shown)


rainwoodman:
Yes this looks right to me. I do not recall why we were not doing this
before.

Probably also want to assert thread provided in genic, if it wasnt I missed
it on the phone screen.

sbird:
Done - I think I was just lazy before

sbird:
It works, merging!

PR #358, opened 2019-10-24
Avoid FOF double exchange

sbird:
The point of this PR is to avoid the double particle exchange when saving a FOF table. Instead it makes a copy of the subset of the particles in the halos, exchanges them to the right processor, saves and discards them. This has the side-effect that a FOF table save does not affect the overall particle order. To do this I had to add a lot of infrastructure to make the particle and slot tables local to their functions in the IO and exchange routines. This was enough effort that I am probably not making these fully local variables any time soon. There was also a lot of messing about with the semi-generic IO routines, which is a little ugly.

What do you think? The IO routine stuff badly needs cleaning up, but I am unsure how to do that without going full C++.

### DIFF
--- a/libgadget/exchange.c
+++ b/libgadget/exchange.c
@@ -452,6 +452,8 @@ domain_build_plan(ExchangeLayoutFunc layoutfunc, const void * layout_userdata, E
         const int target = layoutfunc(i, layout_userdata);
         plan->layouts[n].ptype = PartManager->Base[i].Type;
         plan->layouts[n].target = target;
+        if(target >= plan->NTask || target < 0)
+            endrun(4, "layoutfunc for %d returned unreasonable %d for %d tasks\n", i, target, plan->NTask);
     }
 
     /*Do the sum*/
--- a/libgadget/exchange.c
+++ b/libgadget/exchange.c
@@ -47,10 +47,10 @@ typedef struct {
  * exchange particles according to layoutfunc.
  * layoutfunc gives the target task of particle p.
 */
-static int domain_exchange_once(ExchangePlan * plan, int do_gc, struct part_manager_type * PartManager, struct slots_manager_type * SlotsManager, MPI_Comm Comm);
-static void domain_build_plan(ExchangeLayoutFunc layoutfunc, const void * layout_userdata, ExchangePlan * plan, struct part_manager_type * PartManager);
-static int domain_find_iter_space(ExchangePlan * plan, const struct part_manager_type * PartManager, const struct slots_manager_type * SlotsManager);
-static void domain_build_exchange_list(ExchangeLayoutFunc layoutfunc, const void * layout_userdata, ExchangePlan * plan, struct part_manager_type * PartManager, MPI_Comm Comm);
+static int domain_exchange_once(ExchangePlan * plan, int do_gc, struct part_manager_type * pman, struct slots_manager_type * sman, MPI_Comm Comm);
+static void domain_build_plan(ExchangeLayoutFunc layoutfunc, const void * layout_userdata, ExchangePlan * plan, struct part_manager_type * pman);
+static int domain_find_iter_space(ExchangePlan * plan, const struct part_manager_type * pman, const struct slots_manager_type * sman);
+static void domain_build_exchange_list(ExchangeLayoutFunc layoutfunc, const void * layout_userdata, ExchangePlan * plan, struct part_manager_type * pman, MPI_Comm Comm);
 
 /* This function builts the count/displ arrays from
  * the rows stored in the entry struct of the plan.
@@ -73,7 +73,7 @@ _transpose_plan_entries(ExchangePlanEntry * entries, int * count, int ptype, int
 }
 
 /*Plan and execute a domain exchange, also performing a garbage collection if requested*/
-int domain_exchange(ExchangeLayoutFunc layoutfunc, const void * layout_userdata, int do_gc, struct part_manager_type * PartManager, struct slots_manager_type * SlotsManager, MPI_Comm Comm) {
+int domain_exchange(ExchangeLayoutFunc layoutfunc, const void * layout_userdata, int do_gc, struct part_manager_type * pman, struct slots_manager_type * sman, MPI_Comm Comm) {
     int64_t sumtogo;
     int failure = 0;
 
@@ -86,7 +86,7 @@ int domain_exchange(ExchangeLayoutFunc layoutfunc, const void * layout_userdata,
     /*Structure for building a list of particles that will be exchanged*/
     ExchangePlan plan;
     plan.last = 0;
-    plan.nexchange = PartManager->NumPart;
+    plan.nexchange = pman->NumPart;
 
     MPI_Comm_size(Comm, &plan.NTask);
     /*! toGo[0][task*NTask + partner] gives the number of particles in task 'task'
@@ -103,7 +103,7 @@ int domain_exchange(ExchangeLayoutFunc layoutfunc, const void * layout_userdata,
     int iter = 0;
 
     do {
-        domain_build_exchange_list(layoutfunc, layout_userdata, &plan, PartManager, Comm);
+        domain_build_exchange_list(layoutfunc, layout_userdata, &plan, pman, Comm);
 
         /*Exit early if nothing to do*/
         if(!MPIU_Any(plan.nexchange > 0, Comm))
@@ -113,8 +113,8 @@ int domain_exchange(ExchangeLayoutFunc layoutfunc, const void * layout_userdata,
         }
 
         /* determine for each rank how many particles have to be shifted to other ranks */
-        plan.last = domain_find_iter_space(&plan, PartManager, SlotsManager);
-        domain_build_plan(layoutfunc, layout_userdata, &plan, PartManager);
+        plan.last = domain_find_iter_space(&plan, pman, sman);
+        domain_build_plan(layoutfunc, layout_userdata, &plan, pman);
         walltime_measure("/Domain/exchange/togo");
 
         sumup_large_ints(1, &plan.toGoSum.base, &sumtogo);
@@ -126,7 +126,7 @@ int domain_exchange(ExchangeLayoutFunc layoutfunc, const void * layout_userdata,
          * and a gc will also be done if we have no space for particles.*/
         int really_do_gc = do_gc || (plan.last < plan.nexchange);
 
-        failure = domain_exchange_once(&plan, really_do_gc, PartManager, SlotsManager, Comm);
+        failure = domain_exchange_once(&plan, really_do_gc, pman, sman, Comm);
 
         myfree(plan.ExchangeList);
 
@@ -147,23 +147,23 @@ int domain_exchange(ExchangeLayoutFunc layoutfunc, const void * layout_userdata,
 /*Function decides whether the GC will compact slots.
  * Sets compact[6]. Is collective.*/
 static void
-shall_we_compact_slots(int * compact, ExchangePlan * plan, const struct slots_manager_type * SlotsManager, MPI_Comm Comm)
+shall_we_compact_slots(int * compact, ExchangePlan * plan, const struct slots_manager_type * sman, MPI_Comm Comm)
 {
     int ptype;
     int lcompact[6] = {0};
     for(ptype = 0; ptype < 6; ptype++) {
         /* gc if we are low on slot memory. */
-        if (SlotsManager->info[ptype].size + plan->toGetSum.slots[ptype] > 0.95 * SlotsManager->info[ptype].maxsize)
+        if (sman->info[ptype].size + plan->toGetSum.slots[ptype] > 0.95 * sman->info[ptype].maxsize)
             lcompact[ptype] = 1;
         /* gc if we had a very large exchange. */
-        if(plan->toGoSum.slots[ptype] > 0.1 * SlotsManager->info[ptype].size)
+        if(plan->toGoSum.slots[ptype] > 0.1 * sman->info[ptype].size)
             lcompact[ptype] = 1;
     }
     /*Make the slot compaction collective*/
     MPI_Allreduce(lcompact, compact, 6, MPI_INT, MPI_LOR, Comm);
 }
 
-static int domain_exchange_once(ExchangePlan * plan, int do_gc, struct part_manager_type * PartManager, struct slots_manager_type * SlotsManager, MPI_Comm Comm)
+static int domain_exchange_once(ExchangePlan * plan, int do_gc, struct part_manager_type * pman, struct slots_manager_type * sman, MPI_Comm Comm)
 {
     int n, ptype;
     struct particle_data *partBuf;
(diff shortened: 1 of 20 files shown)


rainwoodman:
Let me read this closely next few days.

Feels like we are calling for lifiting  particle storage as a class. Then
FOF can create a new storage object.

We probably still need an equivalent mess with c++. the anonymous functions
may help a bit in cleaning up the global function name space.

sbird:
That's right - my concern is twofold: firstly about the mess of the IO functions, which seems barely worth it for the gain, secondly about the whole difficulty of having a PI vs a single structure, which makes anything that may take both a slot and a particle data a bit messy. But nothing to be done about that without major major changes.

rainwoodman:
Are you talking about the IO functions as the Get and Set? We do need to
build the split data sets for each particle type. Then maybe we also want
to transform the column before writing it. Once the column is ready bigfile
takes over. To elimenate these boilerplates we need to use the same in mem
representation as the out mem representation. I think thats going to cause
more pain.

PI makes sure members in P have exactly the same memory size, thus allow
quick random acess of the P stream. subclass wont offer you that. Unless
you use a vfunc to return a pointer to the per type data, but thats the
same as having a PI interpreted by the type. array with variable sized
elements!

sbird:
That is very true! If we had a subclass we would have to store and access P and the other particle types separately, with separate trees, separate arrays, etc, which would be very awkward and a huge amount of work.

I also agree about the Get and Set: setting in mem and out mem representations to be the same is going to be almost impossible, and worse limit the physics changes we can do without changing the snapshots.

I don't have a better way of doing what this does, I was just hoping that you would :-)

sbird:
Now I look at the changes in total they don't look so bad, actually. Just the PetaIO macros got a little ugly.

sbird:
With this patch:
+++ b/libgadget/fofpetaio.c
@@ -255,7 +255,6 @@ static void fof_distribute_particles(struct part_manager_type * halopartmanager,
     /* Copy the target task into a temporary array for all particles*/
     int * targettask = mymalloc2("targettask", sizeof(int) * NpigLocal);
 
-#ifdef DEBUG
     int NTask;
     MPI_Comm_size(Comm, &NTask);
     for(i = 0; i < NpigLocal; i ++) {
@@ -263,19 +262,23 @@ static void fof_distribute_particles(struct part_manager_type * halopartmanager,
         if(pi[i].targetTask >= NTask || pi[i].targetTask < 0)
             endrun(23, "pi %d is impossible %d of %d tasks\n",i,pi[i].targetTask, NTask);
     }
-#endif
     #pragma omp parallel for
     for(i = 0; i < NpigLocal; i ++) {
         int index = pi[i].origin % PartManager->MaxPart;
+        if(targettask[index] >= 0)
+            message(1, "i = %d index=%d pi is %d but tt already set: %d.\n",i, index,pi[i].targetTask, targettask[index]);
         targettask[index] = pi[i].targetTask;
     }
     myfree(pi);
-#ifdef DEBUG
+    int bad = 0;
     for(i = 0; i < NpigLocal; i ++) {
-        if(targettask[i] < 0)
-            endrun(4, "targettask %d not changed %d! neighbours: %d %d\n", i, targettask[i], targettask[i-1], targettask[i+1]);
+        if(targettask[i] < 0) {
+            message(4, "targettask %d not changed %d! neighbours: %d %d\n", i, targettask[i], targettask[i-1], targettask[i+1]);
+            bad = 1;
+        }
     }
-#endif
+if(bad)
+endrun(3, "Bad\n");


I get error output:

[ 000149.32 ] Task 0: [ 000149.32 ] Task 2: [ 000149.32 ] Task 1: i = 956 index=6883 pi is 2 but tt already set:
1083124458.
i = 0 index=1174 pi is 1 but tt already set: 1116191084.
i = 0 index=8481 pi is 2 but tt already set: 1085326320.
[ 000149.32 ] Task 0: [ 000149.32 ] Task 3: [ 000149.32 ] Task 1: [ 000149.32 ] Task 2: i = 957 index=6885 pi is
3 but tt already set: 1072858576.
i = 1 index=8483 pi is 1 but tt already set: 1085340212.
i = 1 index=1175 pi is 1 but tt already set: 1085334632.
i = 985 index=985 pi is 1 but tt already set: 0.
[ 000149.32 ] Task 0: [ 000149.32 ] Task 1: [ 000149.32 ] Task 2: i = 960 index=6891 pi is 2 but tt already set:
...
i = 987 index=987 pi is 1 but tt already set: 0.
[ 000149.32 ] Task 0: [ 000149.32 ] Task 2: [ 000149.32 ] Task 1: targettask 14 not changed -1! neighbours: 3 3
i = 5 index=1179 pi is 1 but tt already set: 1085337813.
i = 4 index=8509 pi is 1 but tt already set: 1085356194.
[ 000149.32 ] Task 0: [ 000149.32 ] Task 2: [ 000149.32 ] Task 1: [ 000149.32 ] Task 3: targettask 31 not changed -1! neighbours: 3 3

which is impossible unless mpsort_mpi is returning duplicate corrupted values! This may be a bug in the existing code.

sbird:
It is also rather unusual because I check that pi < NTask (4 here) in the above line, so I do not understand how tt gets set to 1085337813 > 4.

sbird:
Intel compilers...Yu, I am thinking of putting a warning in the readme suggesting users avoid icc where possible. What do you think? Essentially all the problems we've had over the last year have been icc miscompilations

sbird:
It's not like icc is faster than gcc anyway...it's just the default in a lot of places.

sbird:
(alas... from bench-mpi on stampede)

REQUIRE_GATHER_SORT
NTask = 48
src size = 1569
dest size = 1569
csize = 75312
MPSort total time: 0.0116441
FirstSort: 0.00114989
PmaxPmin: 0.000167847
bisect0001: 4.673e-05
bisect0002: 4.88758e-05
bisect0003: 4.05312e-05
bisect0004: 3.98159e-05
bisect0005: 0.00115204
bisect0006: 0.000440121
bisect0007: 0.000135899
bisect0008: 0.000210047
bisect0009: 0.000255108
bisect0032: 0.00375199
findP: 6.91414e-06
LayDistr: 0.000207424
LaySolve: 2.24113e-05
Exchange: 0.00115657
SecondSort: 0.000762701
global ordering fail
global ordering fail
global ordering fail
global ordering fail
global ordering fail
global ordering fail

sbird:
But I found it: see the PR for mpsort.

sbird:
@rainwoodman : I am having problems with mpsort_mpi using the intel compilers. It seems to be failing to sort properly when Gather sort is forced. This seems to be the case where Groups and Segments do not align properly, so either something is messed up with the communicators or there is a latent bug in MPIU_Gather and MPIU_Scatter that the intel compilers just happen to trigger.

Can you take a look? MPsort is a bit beyond me. Until then I will force-disable the Gather sort in MP-Gadget and default pref off UseGlobalSorting, just in case.

rainwoodman:
But do you really need to run with Intel compilers?

Forcing Gather sort means collecting all data to a single rank and sort. I think you may be hitting an overflow due to that. 

If the call to MPI is not alighted to 8 bytes on any pointer from any rank, then it is very likely a buggy MPI will trigger mysterious problems. This also affected Cray MPI when I was doing full machine runs on Cori with FastPM; the solution was, I made sure all customized data types are 8 types aligned. 

There really isn't an easy path for a workaround in MPSort when the culprit is in MPI_Alltoallv -- to avoid misalignment from MPSort, we will have to make a copy of the data, which is not very desirable. 

We can add a warning in MPSort if it sees anything that is misaligned -- then it is easy to catch such issues?

libgadget/exchange.c
```diff
@@ -47,10 +47,10 @@ typedef struct {
  * exchange particles according to layoutfunc.
  * layoutfunc gives the target task of particle p.
 */
-static int domain_exchange_once(ExchangePlan * plan, int do_gc, MPI_Comm Comm);
-static void domain_build_plan(ExchangeLayoutFunc layoutfunc, const void * layout_userdata, ExchangePlan * plan);
-static int domain_find_iter_space(ExchangePlan * plan);
-static void domain_build_exchange_list(ExchangeLayoutFunc layoutfunc, const void * layout_userdata, ExchangePlan * plan, MPI_Comm Comm);
+static int domain_exchange_once(ExchangePlan * plan, int do_gc, struct part_manager_type * PartManager, struct slots_manager_type * SlotsManager, MPI_Comm Comm);
```
rainwoodman: Perhaps use part_manager as a variable name? Then you can add the typedef. It seems to be more conventional using CamelCase for type names.

I found sed helps in massive replacements like this:
```
sed --in-place 's;\<PartManager\>;part_manager;g' *.[ch]
sed --in-place 's;struct part_manager_type;PartManager;g' *.[ch]
```

libgadget/exchange.c
```diff
@@ -113,8 +113,8 @@ int domain_exchange(ExchangeLayoutFunc layoutfunc, const void * layout_userdata,
         }
 
         /* determine for each rank how many particles have to be shifted to other ranks */
-        plan.last = domain_find_iter_space(&plan);
-        domain_build_plan(layoutfunc, layout_userdata, &plan);
+        plan.last = domain_find_iter_space(&plan, PartManager, SlotsManager);
```
rainwoodman: shall we move part manager and slots manager into plan?

libgadget/exchange.c
rainwoodman: part_mgr if part_manager looks too long?

rainwoodman: or even `pman` and `sman`.

libgadget/fofpetaio.c
```diff
@@ -89,18 +71,16 @@ void fof_save_particles(FOFGroups * fof, int num, int SaveParticles, MPI_Comm Co
     if(SaveParticles) {
         struct IOTable IOTable = {0};
         register_io_blocks(&IOTable);
-        fof_distribute_particles(Comm);
+        struct part_manager_type halopartmanager;
```
rainwoodman: PartManager halo_pman

libgadget/fofpetaio.c
```diff
@@ -169,40 +152,96 @@ static int fof_cmp_origin(const void * c1, const void * c2) {
...
+        }
+    }
+
+    struct PartIndex * pi = mymalloc("PartIndex", sizeof(struct PartIndex) * NpigLocal);
+
+    int GrNrMax = -1;   /* will mark particles that are not in any group */
     int GrNrMaxGlobal = 0;
+    NpigLocal = 0;
+    /* Yu: found it! this shall be int64 */
+    const uint64_t taskfactor = ((uint64_t) ThisTask) * (halopartmanager->NumPart + 1Lu);
```
rainwoodman: Is NumPart the same on all ranks? This looks a bit dangerous. pi[.].origin shall be non-overlapping. But if oen rank has  NumPart 0 then this doesn't seem to be right.

At least need a MPI_MAX reduction for the factor multipied on ThisTask.

rainwoodman: taskfactor => task_origin_offset;

libgadget/petaio.c
```diff
@@ -704,26 +705,69 @@ void io_register_io_block(char * name,
...
+    struct slot_info * info = &(((struct slots_manager_type *) slotptr)->info[ptype]); \
+    slottype * sl = (slottype *) info->ptr; \
+    int k; \
+    for(k = 0; k < items; k ++) { \
+        out[k] = *(&(sl[PI].field) + k); \
+    } \
+}
+
+#define SIMPLE_SETTER_PI(name, field, dtype, items, slottype) \
+static void name(int i, dtype * out, void * baseptr, void * slotptr) { \
```
rainwoodman: slotptr really is slotmanptr.

libgadget/sfr_eff.c
```diff
@@ -621,7 +620,7 @@ starformation(int i, double *localsfr, double * sum_sm)
 
 /* Get the parameters of the basic effective
  * equation of state model for a particle.*/
-static struct sfr_eeqos_data get_sfr_eeqos(int i, double dtime)
+static struct sfr_eeqos_data get_sfr_eeqos(struct particle_data * part, struct sph_particle_data * sph, double dtime)
```
rainwoodman: struct xxx_particle_data is probably the only place where we have intentionally avoided typedefing the struct; I don't know why; might be to preserve the Gadget-ness of the code.

libgadget/utils/openmpsort.c
```diff
@@ -1,6 +1,4 @@
-#ifdef _OPENMP
```
rainwoodman: So we will not compile without -fopenmp. Is this already the case?

rainwoodman:
I may have spotted an issue that can be related to what you are seeing. Looks like the way you assign origin in the FOF sorting was problematic. It may cause a similar symptom to what you have been observing. 

I also recommend avoid adding _type to type names -- if we use CamelCase for type name and snake_case for variable name then there shouldn't be a need. If the snake_case looks too long, it can always be abbreviated in a more or less sensible way.

And yes the changes to petaio does not look too bad. Now you do have the capacity of issuing IO from an arbitrary PartManager + SlotManager pair. 

A lot of types pman and sman show up together. They can probably be merged into the same interface. Do we have a place where we only use one of them?

libgadget/fofpetaio.c
sbird: Yes! A good point, I will fix it!

libgadget/utils/openmpsort.c
sbird: It is already the case, enforced by #error in allvars.h

sbird: The _OPENMP guards were there for the clang static analyzer, but that has long since grown OMP support.

sbird:
> But do you really need to run with Intel compilers?

On stampede I either need an intel compiler or an intel MPI: they don't have gcc + mvapich installed.

> Forcing Gather sort means collecting all data to a single rank and sort. I think you may be hitting an overflow due to that.

Ah yes, maybe! This memory is all stack-allocated. 

> If the call to MPI is not alighted to 8 bytes on any pointer from any rank, then it is very likely a buggy MPI will trigger mysterious problems. This also affected Cray MPI when I was doing full machine runs on Cori with FastPM; the solution was, I made sure all customized data types are 8 types aligned.

Everything should be 8-byte aligned by a compiler flag. I would like to bring MP-Sort in-tree and work on this.

> There really isn't an easy path for a workaround in MPSort when the culprit is in MPI_Alltoallv -- to avoid misalignment from MPSort, we will have to make a copy of the data, which is not very desirable.

What do you think about bringing it in-tree and enforcing that the radix is a uint64_t * array? That would enforce 8-byte alignment, and is all we use in MP-Gadget anyway.

> We can add a warning in MPSort if it sees anything that is misaligned -- then it is easy to catch such issues?

libgadget/sfr_eff.c
sbird: For another cleanup :-)

libgadget/exchange.c
sbird: We could but I somehow don't like the look of it, if that's ok.

rainwoodman: Sure; it can be beneficial to keep interfaces small. On the other hand, if we always use PartMan and SlotMan together then probably will want to merge them into the same interface -- something that models a table with extensible rows (variable number of columns per row) span across different MPI ranks.

rainwoodman:
BTW, how do you up-and-down integrate with mp-sort upstream? Perhaps add a README entry somewhere if it is non-trivial.

sbird:
I haven't got a real process yet; at the moment I'm just backporting patches back and forth. Most things in the Gadget version don't want to go upstream.

sbird:
I'm going to merge this: it fixes some bugs and doesn't seem to introduce any new ones! Will fix up MP-Sort when the fix appears.

PR #364, opened 2019-11-05
Heliumreion merge

sbird:
This is the PR for the helium reionization model. @rainwoodman @uptonsanderbeck ?

The model works by placing 30 Mpc bubbles around quasar sized halos, and ionizing the particles in those bubbles.


libgadget/cooling.c
```diff
@@ -21,33 +21,55 @@
...
+        LambdaNet += get_long_mean_free_path_heating(redshift)  / (rho  * PROTONMASS);
+    }
+    return LambdaNet;
+}
+
 /* returns new internal energy per unit mass.
  * Arguments are passed in code units, density is proper density.
  */
-double DoCooling(double redshift, double u_old, double rho, double dt, struct UVBG * uvbg, double *ne_guess, double Z, double MinEgySpec)
+double DoCooling(double redshift, double u_old, double rho, double dt, struct UVBG * uvbg, double *ne_guess, double Z, double MinEgySpec, int isionized)
```
rainwoodman: IsHeIIIonized seems to be closer to what this actually means.

libgadget/cooling.c
```diff
@@ -21,33 +21,55 @@
...
+ * rho: density in protons/cm^3 (physical)
+ * u: internal energy in units of erg/g
+ * Z: metallicity
+ * redshift: redshift
+ * isionized: flags whether the particle has been HeII reionized.
+ */
+static double
+get_lambdanet(double rho, double u, double redshift, double Z, struct UVBG * uvbg, double * ne_guess, int isionized)
+{
+    double LambdaNet = get_heatingcooling_rate(rho, u, 1 - HYDROGEN_MASSFRAC, redshift, Z, uvbg, ne_guess);
```
rainwoodman: Does the meaning of this function change when you HeII to the picture? e.g. do we now have to be aware it is under an approximation, e.g. only the short mean free path heating?

libgadget/cooling_qso_lightup.c
```diff
@@ -0,0 +1,613 @@
...
+ *
+ * 3. Add a switch to place bubbles at random in massive halos if there are no black holes.
+ * 4. Bubble size correlates with halo mass.
+ */
+
+#include <math.h>
+#include <mpi.h>
+#include <string.h>
+#include <gsl/gsl_interp.h>
+#include "cooling_qso_lightup.h"
```
rainwoodman: Move per-module header to the end of includes, if possible.
If the move causes an error then we may have early signs of circular dependency.

libgadget/cooling_qso_lightup.c
```diff
@@ -0,0 +1,613 @@
...
+static double * He_zz;
+static double * XHeIII;
+static double * LMFP;
+static gsl_interp * HeIII_intp;
+static gsl_interp * LMFP_intp;
+/* Counter for particles ionized this step*/
+static int N_ionized;
+
+/*This is a helper for the tests*/
+void set_qso_lightup_par(struct qso_lightup_params qso)
```
rainwoodman: 'set_qso_lightup_for_testing' -- if it really is tests-only.

libgadget/cooling_qso_lightup.c
```diff
@@ -0,0 +1,613 @@
...
+/* Load in reionization history file and build the interpolators between redshift and XHeII.
+ * Format is:
+ * Need to load a text file with the shape of reionization, which gives the overall HeIII fraction,
+ * and the uniform background heating rate. Includes quasar spectral index, etc and most of the physics.
+ * Be careful. Do not double-count uniform long mean free path photons with the homogeneous UVB.
+ *
+ * quasar spectral index
+ * instantaneous absorption threshold energy (in eV)
+ * table of 3 columns, redshift, HeIII fraction, uniform background heating.
+ * The text file specifies the end redshift of reionization.
```
rainwoodman: Perhaps also list the first few lines of the file as an example, or make add a link to a real file. The comments alone still seem to be a bit cryptic.

rainwoodman: Also, is there a tool to generate the file? If so, perhaps also mention it here.

libgadget/cooling_qso_lightup.c
```diff
@@ -0,0 +1,613 @@
...
+    double z1 = sqrt(-2 * log(u1) ) * cos(2 * M_PI * u2);
+    return mu + sigma * z1;
+}
+
+/* Build a list of halos which are candidates for becoming a quasar.
+ * We use only halos with the right mass range.*/
+static int
+build_qso_candidate_list(int ** qso_cand, FOFGroups * fof)
+{
+    /* Run FOF to get a halo catalogue*/
```
rainwoodman: The comment seems to belong to the outside comments.

libgadget/cooling_qso_lightup.c
```diff
@@ -0,0 +1,613 @@
...
+        /* Check that it has the right mass*/
+        if(fof->Group[i].Mass < QSOLightupParams.qso_candidate_min_mass)
+            continue;
+        if(fof->Group[i].Mass > QSOLightupParams.qso_candidate_max_mass)
+            continue;
+        /*Add to the candidate list*/
+        (*qso_cand)[ncand] = i;
+        ncand++;
+    }
+    /*Poison value at the end for safety.*/
```
rainwoodman: Are you concerned if the quasars are local to the domains hosted on this rank?

libgadget/cooling_qso_lightup.c
```diff
@@ -0,0 +1,613 @@
...
+choose_QSO_halo(int ncand, int64_t * ncand_tot, FOFGroups * fof, MPI_Comm Comm)
+{
+    int64_t ncand_total = 0, ncand_before = 0;
+    int NTask, i, ThisTask;
+    MPI_Comm_size(Comm, &NTask);
+    MPI_Comm_rank(Comm, &ThisTask);
+    int * candcounts = (int*) ta_malloc("qso_cand_counts", int, NTask);
+
+    /* Get how many candidates are on each processor*/
+    MPI_Allgather(&ncand, 1, MPI_INT, candcounts, 1, MPI_INT, Comm);
```
rainwoodman: Patterns like this for computing ncand_before and ncand_total has been around in many places. We probably want to make a function under MPIU for this. Could you add a 'TODO: add a function in MPIU' to remind us?

libgadget/cooling_qso_lightup.c
```diff
@@ -0,0 +1,613 @@
...
+static void
+ionize_ngbiter(TreeWalkQueryBase * I,
+        TreeWalkResultBase * O,
+        TreeWalkNgbIterBase * iter,
+        LocalTreeWalk * lv)
+{
+
+    if(iter->other == -1) {
+        /* Gas only*/
+        iter->mask = 1;
```
rainwoodman: Please indicate 1 comes from  (1 << 0) in the comment.

libgadget/cooling_qso_lightup.c
```diff
@@ -0,0 +1,613 @@
...
+
+    int ionized = ionize_single_particle(other);
+
+    unlock_spinlock(other, spin);
+
+    if(!ionized)
+        return;
+
+    /* Add to the ionization counter*/
+    #pragma omp atomic
```
rainwoodman: Do we have a better place to store these variables, e.g. in the Iterator? I remember @sbird may have tried to eliminate thread contentions due to atomics?

libgadget/cooling_qso_lightup.c
```diff
@@ -0,0 +1,613 @@
...
+    N_ionized ++;
+}
```
rainwoodman: BOOKMARK: This is where I have stopped for today.

libgadget/cooling.c
sbird: No, not at all: this is always the heating under the ionization equilibrium assumption, while the other code takes care of deviations from that.

libgadget/cooling.c
sbird: Done

libgadget/cooling_qso_lightup.c
sbird: We have a convention in the other modules of naming functions like this in this way, so I'd like to keep it.

libgadget/cooling_qso_lightup.c
sbird: I'm not sure if I understand what you mean here? Could you explain why that would be a concern? Are you worried about efficiency or correctness?

libgadget/cooling_qso_lightup.c
sbird: Excellent point!

libgadget/cooling_qso_lightup.c
rainwoodman: I haven't got to the part how quasars are modelled in the code. Are they modelled as particles? (Since I noticed there is a tree walk). 

If they are modelled as particles, a domain decomposition on these particles would mean you need fewer exports; but there shouldn't be any correctness issue.

If they are not modelled as particles, then -- what if the quasar hosted on this ranks are very far away from the gas particles hosted on this rank? FOF halos are sorted by length, so this can totally happen.

libgadget/cooling_qso_lightup.c
```diff
@@ -0,0 +1,622 @@
...
+            ncand_before += candcounts[i];
+        ncand_total += candcounts[i];
+    }
+
+    ta_free(candcounts);
+    double drand = get_random_number(fof->TotNgroups);
+    int qso = (drand * ncand_total);
+    *ncand_tot = ncand_total;
+    /* No quasar on this processor*/
+    if(qso < ncand_before || qso >= ncand_before + ncand)
```
rainwoodman: Here is where you select the quasar.
This is very inefficient. The number of MPI_Allgather calls scales like O(NTask) due to the rejection sampling.

rainwoodman: Perhaps give up on invariance or use the particle ID of the BH as the seed?

libgadget/cooling_qso_lightup.c
sbird: In the first version of the code they were just black hole particles, but now they are FOF halos. I don't think they are sorted by length here: after fof_fof they are sorted according to fof_radix_Group_OriginalTaskMinID (they have to be so we can make local black holes with them). In any case, it should not matter much, because the bubble size is 30 Mpc, sp we definitely cross processor boundaries anyway.

libgadget/cooling_qso_lightup.c
sbird: It's true, but the average number of quasars per timestep is <1 (in the whole box) so it isn't the leading order cost. The FOF will dominate.

sbird: We very much need invariance because of this! And I am deliberately not using the BH particles, because I need to use this in a lyman alpha forest run.

libgadget/cooling_qso_lightup.c
```diff
@@ -0,0 +1,622 @@
...
+        (*qso_cand)[ncand] = i;
+        ncand++;
+    }
+    /*Poison value at the end for safety.*/
+    (*qso_cand)[ncand] = -1;
+    *qso_cand = myrealloc(*qso_cand, (ncand+1) * sizeof(int));
+    return ncand;
+}
+
+/* Choose a black hole particle at random to host a quasar from the list of black hole particles.
```
rainwoodman: So this is not choosing a black hole particle, but chosing a halo?

libgadget/cooling_qso_lightup.c
```diff
@@ -0,0 +1,622 @@
...
+    if(!ionized)
+        return;
+
+    int tid = omp_get_thread_num();
+    /* Add to the ionization counter for this thread*/
+    QSO_GET_PRIV(lv->tw)->N_ionized[tid] ++;
+}
+
+static void
+ionize_copy(int place, TreeWalkQueryBase * I, TreeWalk * tw)
```
rainwoodman: I see. You avoided particle altogether by directly accessnig the fof group.

I never realized tw_copy can be used this way!

libgadget/cooling_qso_lightup.c
rainwoodman: I see. This is not rejection sampling because you are always returning exactly 1 QSO_halo globally.

libgadget/cooling_qso_lightup.c
```diff
@@ -0,0 +1,622 @@
...
+    }
+    /*Poison value at the end for safety.*/
+    (*qso_cand)[ncand] = -1;
+    *qso_cand = myrealloc(*qso_cand, (ncand+1) * sizeof(int));
+    return ncand;
+}
+
+/* Choose a black hole particle at random to host a quasar from the list of black hole particles.
+ * This function chooses a single quasar from the total candidate list of quasars on *all* processors.
+ * We seed the random number generator off the number of existing quasars.
```
rainwoodman: add 'Returns the local index of the halo in fof' if the halo is hosted on this rank, -1 if the halo is not hosted on this rank', or something alike.

libgadget/cooling_qso_lightup.c
```diff
@@ -0,0 +1,622 @@
...
+    int64_t non_overlapping_bubble_number = n_gas_tot * totbubblegasmass / All.CP.OmegaBaryon;
+    double initionfrac = gas_ionization_fraction();
+    double curionfrac = initionfrac;
+
+    if(curionfrac < desired_ion_frac) {
+        ncand = build_qso_candidate_list(&qso_cand, fof);
+        walltime_measure("/HeIII/Find");
+    }
+
+    while (curionfrac < desired_ion_frac){
```
rainwoodman: Perhaps add a counter on the number of iterations, and issue warning or die if there are too many.
I also wonder if this shall be stochastic -- like how mass tracks BH_mass; such that the curionfrac is no longer always biased lower than desired. But doing so probably requires computing the amount of new ion_frac a new qso can produce.

libgadget/cooling_qso_lightup.c
```diff
@@ -0,0 +1,622 @@
...
+        int64_t tot_qso_ionized = 0;
+        /* Check that the ionization fraction changed*/
+        MPI_Allreduce(&n_ionized, &tot_qso_ionized, 1, MPI_INT64, MPI_SUM, MPI_COMM_WORLD);
+        curionfrac += (double) tot_qso_ionized / (double) n_gas_tot;
+        tot_n_ionized += tot_qso_ionized;
+        if(new_qso > 0)
+            message(1, "HeII: Quasar %d changed the HeIII ionization fraction to %g, ionizing %ld\n", qso_cand[new_qso], curionfrac, tot_qso_ionized);
+        /* Break the loop if we do not ionize enough particles this round.
+         * Try again next timestep when we will hopefully have new BHs.*/
+        if(tot_qso_ionized < 0.01 * non_overlapping_bubble_number)
```
rainwoodman: It feels a bit unstable. If I compute tot_qso_ionized for all halos, does it it lean heavily to the low end? If this happens then we would be skipping almost every step.

libgadget/cooling_qso_lightup.c
```diff
@@ -0,0 +1,622 @@
...
+    double curionfrac = initionfrac;
+
+    if(curionfrac < desired_ion_frac) {
+        ncand = build_qso_candidate_list(&qso_cand, fof);
+        walltime_measure("/HeIII/Find");
+    }
+
+    while (curionfrac < desired_ion_frac){
+        /* Get a new quasar*/
+        int new_qso = choose_QSO_halo(ncand, &ncand_tot, fof, MPI_COMM_WORLD);
```
rainwoodman: The MPI gather is only used for detecting if the global index falls into the current rank; so technically you could have generated a list of non-repeating global fof indices, and convert them to local with a precomputed offset table.

tools/HeII_input_file_maker.py
```diff
@@ -0,0 +1,280 @@
...
+
+The parameters used will be printed in a comment at the top of the output table."""
+
+import argparse
+import numpy as np
+import scipy.integrate
+import scipy.interpolate
+
+class Cosmology:
+    """Little module to specify the mean densities and the Hubble rate"""
```
rainwoodman: Probably want to shout loudly this is Flat LCDM, no radiation or neutrino.

tools/HeII_input_file_maker.py
```diff
@@ -0,0 +1,280 @@
...
+        else:
+            args.z_i = 6.0
+    if args.z_f < 0:
+        if args.hist == "linear":
+            args.z_f = 2.8
+        else:
+            args.z_f = 2.
+
+    heat = HeIIheating(hist = args.hist, z_i = args.z_i, z_f= args.z_f, Emax=args.Emax, alpha_q = args.alphaq, clumping_fac = args.cf)
+    heat.setUpInterpTable(args.outfile)
```
rainwoodman: WriteInterpTable? the spelling setUp does not suggest writing, making this line a bit mysterious.

tools/HeII_input_file_maker.py
```diff
@@ -0,0 +1,280 @@
...
+
+    def dXHeIIIdz(self, redshift):
+        """Change in the derivative of XHeIII, where XHeIII evolves linearly with redshift.
+           Set initial and final HeII reion redshifts (default is reion_z_f=2.8)."""
+        if self.z_f <= redshift <= self.z_i:
+            return 1./(self.z_f-self.z_i)
+        return 0.
+
+class QuasarHistory(LinearHistory):
+    """Determines the HeII reionization history from a quasar emissivity function"""
```
rainwoodman: QuasarHistory as subclass of a LinearHistory? 

They seem to be two different implementations of the HeIIReionHistory interface?

rainwoodman:
Now I understand how this works. Thanks for the clarifications!

libgadget/cooling_qso_lightup.c
sbird: Correct: the comment is outdated

tools/HeII_input_file_maker.py
sbird: LinearHistory is the simplest possible implementation of the reionization fraction, so I thought it made sense to have QuasarHistory extend it. The commonality is pretty small, so we could also make them separate classes.

libgadget/cooling_qso_lightup.c
sbird: Exactly! It pays off that you wrote a general interface!

libgadget/cooling_qso_lightup.c
sbird: Right!

libgadget/cooling_qso_lightup.c
sbird: Yes - unfortunately that is impossible without looping over all particles (because particles cannot be ionized twice).

libgadget/cooling_qso_lightup.c
sbird: The reason for the break is that we might have no non-ionized regions close to a halo (especially near the end of reionization). The code solves this by flash-ionizing everything at some final time.

libgadget/cooling_qso_lightup.c
sbird: We could have done that, but it is much more complicated. We can try it if this ever turns into a scalability problem

sbird:
Thanks for the review!

sbird:
Your comments on the quasar choice loop and the break condition have made me think again!

libgadget/cooling_qso_lightup.c
sbird: In fact, I will do this!

libgadget/cooling_qso_lightup.c
sbird: We have no need to have a counter because we always remove a quasar from the candidate list, and once we run out the loop will terminate.

libgadget/cooling_qso_lightup.c
sbird: Note that when a quasar goes off we ionize everything around it, even if that pushes us above the desired ionization fraction. So it is not always biased low, it is just a step function.

tools/HeII_input_file_maker.py
```diff
@@ -0,0 +1,292 @@
...
+        return e
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser()
+    parser.add_argument('--alphaq', type=float, default=1.7, help='QSO spectral index', required=True)
+    parser.add_argument('--Emax', type=float, default=150, help='Threshold long-mean-free-path photon energy in eV', required=False)
+    parser.add_argument('--cf', type=float, default=3., help='Subgrid clumping factor', required=False)
+    parser.add_argument('--z_i', type=float, default=-1, help='Start redshift of helium reionization', required=False)
+    parser.add_argument('--z_f', type=float, default=-1, help='End redshift of helium reionization', required=False)
+    parser.add_argument('--hist', type=str, default="quasar", help='Type of reionization history', required=True)
```
rainwoodman: You can declare all allowed choices with https://docs.python.org/3/library/argparse.html#choices

Also, the default z_i, and z_f can probably be written as a dict. Though it is debatable if dict is more readable than if else, shorter and avoids explicit branching certainly.

libgadget/partmanager.h
```diff
@@ -26,7 +26,8 @@ struct particle_data
         unsigned int Swallowed            :1; /* True if the particle is being swallowed; used in BH to determine swallower and swallowee;*/
         unsigned int spare_0            :1;
 
-        unsigned char Generation; /* How many particles it has spawned; used to generate unique particle ID.
+        unsigned int HeIIIionized              :1; /*True if the particle has undergone helium reionization*/
```
rainwoodman: Shouldn't this remove spare_0?

libgadget/cooling_qso_lightup.c
```diff
@@ -0,0 +1,642 @@
...
+        for (i = 0; i < PartManager->NumPart; i++){
+            if (P[i].Type == 0)
+                nionized += ionize_single_particle(i);
+        }
+        sumup_large_ints(1, &nionized, &nion_tot);
+        message(0, "HeII: Helium ionization finished, flash-ionizing %ld particles (%g of total)\n", nion_tot, (double) nion_tot /(double) n_gas_tot);
+    }
+
+    double rhobar = All.CP.OmegaBaryon * (3 * HUBBLE * All.CP.HubbleParam * HUBBLE * All.CP.HubbleParam)/ (8 * M_PI * GRAVITY) / pow(1 + redshift,3);
+    double totbubblegasmass = 4 * M_PI / 3. * pow(QSOLightupParams.mean_bubble, 3) * rhobar;
```
rainwoodman: Aren't the bubbles in overdense regions? If so, then mean_bubble is probably not the bubble's proper 'radius'. mean_bubble_radius_param?

libgadget/cooling_qso_lightup.c
```diff
@@ -0,0 +1,642 @@
...
+        int64_t tot_qso_ionized = 0;
+        /* Check that the ionization fraction changed*/
+        MPI_Allreduce(&n_ionized, &tot_qso_ionized, 1, MPI_INT64, MPI_SUM, MPI_COMM_WORLD);
+        curionfrac += (double) tot_qso_ionized / (double) n_gas_tot;
+        tot_n_ionized += tot_qso_ionized;
+        if(new_qso > 0)
+            message(1, "HeII: Quasar %d changed the HeIII ionization fraction to %g, ionizing %ld\n", qso_cand[new_qso], curionfrac, tot_qso_ionized);
+        /* Break the loop if we do not ionize enough particles this round.
+         * Try again next timestep when we will hopefully have new BHs.*/
+        if(tot_qso_ionized < 0.01 * non_overlapping_bubble_number && iteration > 10)
```
rainwoodman: might be cleaner if iteration > 10 is moved to 564. Then we see the loop is bound by 10 iteration without parsing the loop body.

rainwoodman: oops. Except if the ionization has been efficient then we'll keep doing it forever -- is this behavior intended?

libgadget/cooling_qso_lightup.c
sbird: Well - at z=3 with a 30 Mpc bubble all bubbles are basically mean density

libgadget/partmanager.h
sbird: Oops, yes! Merge error!

libgadget/cooling_qso_lightup.c
sbird: The idea is that we exit the loop if ionization is getting slow and if we have already tried 10 times. This is to guard against picking an already ionized quasar once or twice, and no ionization happening. If the ionization is efficient then the loop will exit only when we have ionized enough, which is intentional. The loop will also eventually exit when we run out of quasars.

tools/HeII_input_file_maker.py
sbird: Nice! I did not know about choices.

sbird:
Merging!

PR #366, opened 2019-11-16
Hireionscript

sbird:
This PR introduces and cleans up the UV fluctuation script file, and fixes the units bug in MP-Gadget  when loading it. 

@rainwoodman : Are you absolutely sure that pmesh.pm takes Box in Mpc/h and not Mpc?

### DIFF
--- a/libgadget/cooling.c
+++ b/libgadget/cooling.c
@@ -32,14 +32,12 @@
 static struct cooling_units coolunits;
 
 /*Do initialisation for the cooling module*/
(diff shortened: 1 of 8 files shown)


rainwoodman:
I am positive the script uses Mpc/h. ParticleMesh itself does not assume any units. The unit is induced by the PowerSpectrum code, which uses Mpc/h.

sbird:
Thank you! I will merge this when it goes green.

Issue #67, opened 2017-04-06
Preserve swallowed particles in snapshot files.


rainwoodman:
The only way to make sure the full swallowing history is preserved is to keep swallowed particles in the snapshot files. They shall carry an attribute 'SwallowerID', which is the swallower.

We need to rework the garbage detection code to support this. We need to add IsGarbage for truely removed particles. IsGarbage.

(Mass == 0) will still skip all interactions.

Gas swallowing will produce garbages. BH swallowing will not. We shall be able to track merger history with this.

sbird:
We have IsGarbage now: we need to change the force tree build, drift and timestep parts of the code to not look at Swallowed particles, and then stop marking these particles Garbage. We should also change the writing code to mark them garbage after IO is done. Let's first do this for BH particles (type 5) only, as it is easiest.

PR #371, opened 2019-11-26
Issue67

sbird:
Issue67 v 2.

### DIFF
--- a/libgadget/blackhole.c
+++ b/libgadget/blackhole.c
@@ -589,7 +589,8 @@ blackhole_feedback_ngbiter(TreeWalkQueryBHFeedback * I,
     if(P[other].ID == I->ID) return;
 
      /* BH does not accrete wind */
-    if(P[other].Type == 0 && SPHP(other).DelayTime > 0) return;
+    if(P[other].Type == 0 && SPHP(other).DelayTime > 0)
+        return;
 
     struct SpinLocks * spin = BH_GET_PRIV(lv->tw)->spin;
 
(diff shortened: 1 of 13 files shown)


sbird:
This is Karime's issue 67 fix with a couple more changes on top. Notably save some more useful BH fields to disc.

libgadget/blackhole.c
```diff
@@ -585,28 +592,22 @@ blackhole_feedback_ngbiter(TreeWalkQueryBHFeedback * I,
...
-        /* We do not know how to notify the tree of mass changes. so
-         * blindly enforce a mass conservation for now. */
+        /* Leave the swallowed BH mass around
+         * so we can work out mass at merger. */
         O->Mass += (P[other].Mass);
         O->BH_Mass += (BHP(other).Mass);
-        P[other].Mass = 0;
-        BHP(other).Mass = 0;
-
-        slots_mark_garbage(other, PartManager, SlotsManager);
```
rainwoodman: Do we no longer mark the swallowed particle a  garbage?

sbird: That's right - we don't want to GC it as per #67 we want to keep the swallowed particle around for merger tree stuff.

sbird: This is actually a much better way of tracking mergers than the usual merger tree code - I'm surprised other people haven't done it. It kind of obsoletes an entire complicated subfield...

maamari: What is meant by obsoletes an entire subfield? Which subfield is that in reference to

sbird: There are a whole bunch of people who build algorithms to follow the mergers of dark matter halos from incomplete temporal sampling of snapshots, for example https://arxiv.org/abs/1307.3577

rainwoodman:
Using "tracer particles" is not exactly a new idea though.

People do worry about the tracer gets ejected or stolen.

sbird:
It's true, this is not novel really. But it does seem rarer than I would expect.

libgadget/blackhole.c
maamari: Ah okay interesting, thanks for the link

rainwoodman:
I guess it was because it was not in a public version of Gadget.

PR #372, opened 2019-11-27
Smallchanges

sbird:
This does some cleanups and removes the last global variables in allvars - including ThisTask! Hooray!


sbird:
Commit 9500783c947be9 changes output, none of the others do, merging.

PR #373, opened 2019-11-27
Fixes

sbird:
Fix the UVBG model, which had forgotten to account for the random position offset, generating very weird results!


sbird:
The FOF table saving code as well!

PR #376, opened 2019-12-02
Remove FOF splay from unlocked HEAD function

sbird:
The HEAD() function is called from a threaded context, and not protected
by locks (unlike HEADl()), so we can't change the head particle inside
it. The trees should still be short because we call update_root inside
HEADl while protected by a lock. Hopefully this fixes the pesky deadlock
on stampede.


sbird:
This fixes the deadlock, merging.

PR #377, opened 2019-12-12
Add checked version of MPI_Allreduce and MPI_Reduce

sbird:
This will detect when the MPI library is incorrectly returning zeros
from collective calls. Checked are MPI_INT, MPI_DOUBLE and MPI_LONG,
with MPI_MIN and MPI_MAX.

### DIFF
--- a/libgadget/utils/system.c
+++ b/libgadget/utils/system.c
@@ -13,7 +13,7 @@
 #include <signal.h>
 #include <gsl/gsl_rng.h>
 
-
+#define __UTILS_SYSTEM_C
 #include "system.h"
 #include "mymalloc.h"
 #include "endrun.h"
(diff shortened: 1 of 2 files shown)


sbird:
To check whether frontera is working

libgadget/utils/system.h
```diff
@@ -83,4 +83,18 @@ int _MPIU_Barrier(const char * fn, const int ln, MPI_Comm comm);
 /* Fancy barrier which warns if there is a lot of imbalance. */
 #define MPIU_Barrier(comm) _MPIU_Barrier(__FILE__, __LINE__, comm)
 
+#ifdef DEBUG
+/* Checked versions of some MPI routines which make sure that the number we get out is sensible*/
+int MPI_Allreduce_Checked(const void *sendbuf, void *recvbuf, int count, MPI_Datatype datatype, MPI_Op op, MPI_Comm comm, const int line, const char * file);
+int MPI_Reduce_Checked(const void *sendbuf, void *recvbuf, int count, MPI_Datatype datatype, MPI_Op op, int root, MPI_Comm comm, const int line, const char * file);
+/* Defines, guarded so the implementation can still use the original*/
+#ifndef __UTILS_SYSTEM_C
```
rainwoodman: header that behaves differently if included from the impl is not a good pattern.

What about lifting the function rename macros to mpi-checked.h, and only include it in the soruce files that you want mpi to be checked?

libgadget/utils/system.c
```diff
@@ -636,3 +636,93 @@ void gadget_setup_thread_arrays(int * dest, int * srcs[], size_t sizes[], size_t
...
+            else if(datatype == MPI_DOUBLE) {
+                if(((double *) inputcpy)[i] < ((double *) recvbuf)[i])
+                    endrun(12, "MPI_Allreduce with MPI_DOUBLE MPI_MAX | MPI_SUM has int %d bad: (input %g < out %g) at %s:%d\n", i, *((double*) inputcpy), *((double*) recvbuf), file, line);
+            }
+        }
+    }
+}
+
+int
+MPI_Allreduce_Checked(const void *sendbuf, void *recvbuf, int count, MPI_Datatype datatype, MPI_Op op, MPI_Comm comm, const int line, const char * file)
```
rainwoodman: What about CHECKED_MPI_Allreduce?

libgadget/utils/system.h
sbird: I agree! I really don't want a different header though - we have no other way of knowing that there is a bug here and I want maximal coverage, since Allreduce returning zero seems to only happen intermittently. It is too easy to forget the include if it is separate.

libgadget/utils/system.c
sbird: If you feel strongly about it, sure, but I prefer MPI_Allreduce_Checked!

PR #378, opened 2019-12-12
Notransients

sbird:
This moves the things in the union{} to their respective modules.

### DIFF
--- a/libgadget/density.c
+++ b/libgadget/density.c
@@ -139,6 +139,9 @@ typedef struct {
 } TreeWalkResultDensity;
 
 struct DensityPriv {
+    /* Current number of neighbours*/
+    MyFloat *NumNgb;
+    /* Lower and upper bounds on smoothing length*/
     MyFloat *Left, *Right;
     MyFloat (*Rot)[3];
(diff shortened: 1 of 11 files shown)


sbird:
> This can be done, because petapm does not exchange particles or permute particles. thus, RegionInd is stable.
> Am I correct on this assertion?

Yes, that is exactly right. All communication is done at the Pencil level.

sbird:
> I recall this may corrupt snapshots can contain wrong GrNr ; does it not?

I will check. I remember the problem you're thinking of, and I think it has been fixed by refactorings.

sbird:
Confirmed this makes no difference to FOF or particle output

sbird:
and is very slightly faster (but not a lot)

rainwoodman:
And even in force calculation particles do not move in memory locations. The copy / reduce function takes care of packing P and non-P data fields during communication.

Besides GC and Domain, any other component changes the memory location of particles?

sbird:
That's right. I don't believe anything else changes the memory location of the particles - this is anyway an invalid operation when the tree is allocated.

Issue #54, opened 2016-09-29
Time step limiter


rainwoodman:
Joop Schaye mentioned time step limiter on gas is a good idea. After applying a limiter isotropic kinetic feedback and thermal feedback become the same (which it should have been). 

It is not horribly hard to implement and we shall put it in.

PR #379, opened 2019-12-23
Fronterafix

sbird:
Fix a rare issue found in the frontera run

### DIFF
--- a/libgadget/drift.c
+++ b/libgadget/drift.c
@@ -44,6 +44,16 @@ static void real_drift_particle(int i, inttime_t ti1, const double ddrift, const
     int j;
     if(P[i].IsGarbage || P[i].Swallowed) {
         P[i].Ti_drift = ti1;
+        /* Keep the random shift updated so the
+         * physical position of swallowed particles remains unchanged.*/
+        for(j = 0; j < 3; j++) {
+            P[i].Pos[j] += random_shift[j];
+            while(P[i].Pos[j] > All.BoxSize) P[i].Pos[j] -= All.BoxSize;
+            while(P[i].Pos[j] <= 0) P[i].Pos[j] += All.BoxSize;
+        }
+        /* Swallowed particles still need a peano key.*/
+        if(P[i].Swallowed)
+            P[i].Key = PEANO(P[i].Pos, All.BoxSize);
         return;
     }
     inttime_t ti0 = P[i].Ti_drift;
--- a/libgadget/drift.c
+++ b/libgadget/drift.c
@@ -110,6 +110,11 @@ static void real_drift_particle(int i, inttime_t ti1, const double ddrift, const
         if(fac > 1.25)
             fac = 1.25;
         P[i].Hsml *= fac;
+        /* Cap the Hsml: if DivVel is large for a particle with a long timestep
+         * (most likely a wind particle) Hsml can very rarely run away*/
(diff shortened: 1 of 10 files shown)


libgadget/drift.c
```diff
@@ -110,6 +110,10 @@ static void real_drift_particle(int i, inttime_t ti1, const double ddrift, const
         if(fac > 1.25)
             fac = 1.25;
         P[i].Hsml *= fac;
+        /* Cap the Hsml: if DivVel is large for a particle with a long timestep
+         * (most likely a wind particle) Hsml can very rarely run away*/
+        if(P[i].Hsml > All.BoxSize /2.)
```
rainwoodman: Isn't half of box a bit too large? I remember we used something like 30 Kpc/h in MB-II for the hsml cap at some point.

With such large hsml the particle will interact with every particle in the entire simulation -- so it must be exported to all ranks.

libgadget/hydra.c
```diff
@@ -292,12 +292,18 @@ hydro_ngbiter(
                   " We haven't implemented tracer particles and this shall not happen\n");
     }
 
+    /* Wind particles do not interact hydrodynamically: don't produce hydro acceleration
+     * or change the signalvel.*/
+    if(HYDRA_GET_PRIV(lv->tw)->WindOn && winds_is_particle_decoupled(other))
```
rainwoodman: Yes this makes sense -- if decoupled it doesn't make sense they still affect the signal velocity.

libgadget/drift.c
sbird: Yes, I think you are right! I think the right thing is to scale with the MeanInterparticleSpacing. Perhaps DMIN(MeanInterparticleSpacing * 100, All.BoxSize)?

rainwoodman: I am a bit short circuited here. Looks like the factor (50) translate to the density / mean density ratio -- so if you know the target lowest baryon density you want to model, then you can work out the number.

sbird: That's right - I was thinking we could bound it using a lognormal density distribution. But I am unsure that this is the right thing to do. All.BoxSize/2 has the advantage that it is the largest physically sensible value for any particle distribution, even pathological ones where, eg, one particle is in the middle and all the others are at the box edges.

All the other choices could potentially give the wrong answer in some subtle way in a future paper studying deep voids, or easily hide real bugs. So I actually think that BoxSize/2 is the right choice.

sbird:
Would you say that the true fix to this is a timestep limiter?

sbird:
I think this causes a problem because if we have two gas particles, A and B and A has a timestep much longer than B, then A's Hsml can become large through DivVel. But DivVel is only large because B has a large acceleration and thus a short timestep: if we limit A's timestep to be max, eg, 2 bins higher than B's, then A.Hsml will not have time to become large. What do you think? This is Issue #54

sbird:
What do you think of this? It is a better fix, I think.

sbird:
In fact, this is what WindFreeTravelDensFac is for - it recouples the gas once the density around it drops. But it is set to zero in the current code. I wonder why?

sbird:
I think the issues I was having was because WindFreeTravelDensFac was not enabled, so dense gas was escaping too far. But I still think these are the right fixes: this way the decoupling is really just setting the pressure to zero.

PR #384, opened 2020-01-06
Windmodelfix

sbird:
The only really important commit is 2bb1165 which changes the wind model to be less timing dependent: this should help the code give more similar answers in threaded mode vs. non-threaded mode for the hydro run. It does this by allowing particles to be made wind twice. Also switch some of the examples to isotropic winds, which modern practice shows is simpler and somewhat cleaner.


sbird:
Runs on frontera and does solve the small timesteps problem, merging.

PR #387, opened 2020-01-13
Bufferfix

sbird:
Disable the large buffer fix again, since the TACC admins fixed the root cause.


sbird:
but hide it behind a compile flag in case the bug comes back!

PR #388, opened 2020-01-14
Swallowfix

sbird:
Fix a rare case where gas particles were half swallowed, the black hole merges, and then they are swallowed but not garbage.

### DIFF
--- a/libgadget/blackhole.c
+++ b/libgadget/blackhole.c
@@ -84,7 +84,8 @@ typedef struct {
 } TreeWalkNgbIterBHFeedback;
 
 struct BHPriv {
-    /*Temporary array to store the IDs of the swallowing black hole for gas*/
+    /* Temporary array to store the IDs of the swallowing black hole for gas.
+     * We store ID + 1 so that SwallowID == 0 can correspond to the unswallowed case. */
     MyIDType * SPH_SwallowID;
     /* These are temporaries used in the accretion treewalk*/
     MyFloat * MinPot;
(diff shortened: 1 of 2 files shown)


libgadget/blackhole.c
```diff
@@ -84,7 +84,8 @@ typedef struct {
 } TreeWalkNgbIterBHFeedback;
 
 struct BHPriv {
-    /*Temporary array to store the IDs of the swallowing black hole for gas*/
+    /* Temporary array to store the IDs of the swallowing black hole for gas.
+     * We store ID + 1 so that SwallowID == 0 can correspond to the unswallowed case. */
```
rainwoodman: Might be cleaner using 0xffff ffff ffff ffff for un-swallowed.

libgadget/blackhole.c
```diff
@@ -233,20 +234,7 @@ blackhole(const ActiveParticles * act, ForceTree * tree, FILE * FdBlackHoles)
...
-                priv->SPH_SwallowID[P[p_i].PI] = -1;
-        }
-    }
-    else {
-        #pragma omp parallel for
-        for(i = 0; i < SlotsManager->info[0].size; i ++) {
-            priv->SPH_SwallowID[i] = -1;
-        }
-    }
+    memset(priv->SPH_SwallowID, 0, SlotsManager->info[0].size * sizeof(MyIDType));
```
rainwoodman: memset, -1 will set to all 0xff-s.

libgadget/blackhole.c
sbird: Then I can't use < cleanly and I need to worry about integer overflow. This is a temporary only used in this routine so a constant offset is not a big deal.

PR #390, opened 2020-01-23
Fofmasschanges

sbird:
This implements what we talked about: the smallest seed black hole now scales smoothly with the resolution of the simulation. Two less parameters we usually need to change!


sbird:
Decided not to do this

PR #391, opened 2020-01-23
Batchedtree

sbird:
This commit keeps around the suns array in the force tree node and uses it for leaf tree walks. In principle this should make the tree walking faster because all leaf particle accesses are linear, and the tree build faster because memcpy is no longer needed at every step. It also makes debugging easier (because now one can see the tree structure at all times) and removes a common source of bugs. 

It does bloat the size of the tree nodes by ~30%, but we get some of the memory back because we don't need the Nextnode array anymore. 

Mostly this is a reorganisation: the real reason is that it enables future optimizations, like skipping the moments build (because nextnode and sibling can now be set during the initial node creation) and walking the tree for all particles in a node at once.

Please review! Lightly tested so far.

### DIFF
--- a/libgadget/forcetree.c
+++ b/libgadget/forcetree.c
@@ -738,14 +738,10 @@ force_update_particle_node(int no, int sib, const ForceTree * tree, const int Hy
     if(tree->Nodes[no].f.ChildType != PARTICLE_NODE_TYPE)
         endrun(3, "force_update_particle_node called on node %d of wrong type!\n", no);
     /*Last value of tails is the return value of this function*/
-    int j, suns[NMAXCHILD];
+    int j;
 
-    /* this "backup" is necessary because the nextnode
-     * entry will overwrite one element (union!) */
-    int noccupied = tree->Nodes[no].u.s.noccupied;
-    for(j = 0; j < noccupied; j++) {
-        suns[j] = tree->Nodes[no].u.s.suns[j];
-    }
+    const int noccupied = tree->Nodes[no].u.s.noccupied;
+    int * suns = tree->Nodes[no].u.s.suns;
 
     /*After this point the suns array is invalid!*/
     force_zero_union(&tree->Nodes[no]);
--- a/libgadget/forcetree.c
+++ b/libgadget/forcetree.c
@@ -233,6 +233,13 @@ static void init_internal_node(struct NODE *nfreep, struct NODE *parent, int sub
     for(j = 0; j < NMAXCHILD; j++)
         nfreep->u.s.suns[j] = -1;
     nfreep->u.s.noccupied = 0;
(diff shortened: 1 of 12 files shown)


sbird:
This now runs to completion and makes a small difference to speed. If anything it is a bit slower, which is probably memory pressure. It is hard to compare completely because the force accuracy has improved a bit: before some particles in a node were culled, some not, but it must now include all particles within a particle-containing node.

sbird:
Appreciate your review. Meanwhile, I will start doing the interesting stuff with this.

libgadget/forcetree.c
```diff
@@ -78,13 +78,18 @@ force_tree_eh_slots_fork(EIBase * event, void * userdata)
...
+    int no = force_get_father(parent, tree);
+    struct NODE * nop = &tree->Nodes[no];
+    /* FIXME: We lose particles if the node is full.
+     * At the moment this does not matter, because
+     * the only new particles are stars, which do not
+     * participate in the SPH tree walk.*/
+    if(nop->s.noccupied < NMAXCHILD) {
+       nop->s.suns[nop->s.noccupied-1] = child;
+        nop->s.noccupied++;
+    }
```
rainwoodman: From your description, sounds like we shall crash here; at least print something?

rainwoodman:
" but it must now include all particles within a particle-containing node.", did you mean no matter what level the node is?

rainwoodman:
I will get to read this more closely over the next weekend.

It sounds like a good idea.

If a field is disposable, then it is cleaner to separately allocate the field for all nodes and free it when no longer needed, instead of doing unions. We have been moving along this approach with treewalkers with your recent PRs.

libgadget/forcetree.c
sbird: We don't crash, we just don't add the particle to the tree. This has no negative consequences at this time, because we don't do anything with the newly created particle anyway! Perhaps a message at DEBUG? What should it say?

sbird:
This is not making a disposable field separately allocated. What it does is to store the formerly disposable field, making it non-disposable.

sbird:
> " but it must now include all particles within a particle-containing node.", did you mean no matter what level the node is?

Before when a node was fully open each particle would be checked with cull_node: now only the node is checked and all particles in it are either included or not.

sbird:
The current version very slightly increases force accuracy for close particle pairs (commit 70ee2f4dc9) and slightly alters the output. It has a small (negative) impact on performance, likely from the increased size of the NODE structure, or maybe just from the improved force accuracy. I think I will move to tuning the pairwise branch before I merge this, as it doesn't really do anything on its own.

sbird:
Closing because merging this with pairwise.

PR #392, opened 2020-01-24
Moreatomic

sbird:
This is a quite clever use of omp atomic capture to implement an atomic max algorithm in C

### DIFF
--- a/libgadget/blackhole.c
+++ b/libgadget/blackhole.c
@@ -100,8 +100,6 @@ struct BHPriv {
      * in the feedback treewalk*/
     MyFloat * BH_FeedbackWeightSum;
 
-    /* Particle SpinLocks*/
-    struct SpinLocks * spin;
     /* Counters*/
     int64_t * N_sph_swallowed;
     int64_t * N_BH_swallowed;
@@ -246,10 +244,7 @@ blackhole(const ActiveParticles * act, ForceTree * tree, FILE * FdBlackHoles)
     priv->BH_Entropy = mymalloc("BH_Entropy", SlotsManager->info[5].size * sizeof(MyFloat));
     priv->BH_SurroundingGasVel = (MyFloat (*) [3]) mymalloc("BH_SurroundVel", 3* SlotsManager->info[5].size * sizeof(priv->BH_SurroundingGasVel[0]));
 
-    /* This allocates memory*/
-    priv[0].spin = init_spinlocks(PartManager->NumPart);
     treewalk_run(tw_accretion, act->ActiveParticle, act->NumActiveParticle);
-    free_spinlocks(priv[0].spin);
 
     myfree(priv->BH_SurroundingGasVel);
     myfree(priv->BH_Entropy);
--- a/libgadget/blackhole.c
+++ b/libgadget/blackhole.c
@@ -463,7 +463,9 @@ blackhole_accretion_ngbiter(TreeWalkQueryBHAccretion * I,
     /* Accretion / merger doesn't do self interaction */
     if(P[other].ID == I->ID) return;
 
-    struct SpinLocks * spin = BH_GET_PRIV(lv->tw)->spin;
+    /* This variable has to be set so the atomic
+     * satisfies the atomic capture form, but it is not used*/
+    MyIDType maxsw = (MyIDType) -1;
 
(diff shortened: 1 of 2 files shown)


sbird:
Yu, what do you think? It removes two of the remaining uses of the locks.

libgadget/blackhole.c
```diff
@@ -592,10 +581,14 @@ blackhole_feedback_ngbiter(TreeWalkQueryBHFeedback * I,
...
 
-    if(P[other].Swallowed && P[other].Type == 5)	/* we have a black hole merger */
+     /* we have a black hole merger! */
+    if(P[other].Type == 5 && BHP(other).SwallowID != (MyIDType) -1)
     {
         if(BHP(other).SwallowID != I->ID) return;
 
+        /* Swallow the particle*/
+        BHP(other).SwallowTime = All.Time;
+        P[other].Swallowed = 1;
```
rainwoodman: The distinguish between SwallowID != -1 and Swallowed is not clear (even before the PR).

libgadget/winds.c
```diff
@@ -471,18 +440,25 @@ sfr_wind_feedback_ngbiter(TreeWalkQueryWind * I,
...
-        /* we may want to add another lock to avoid this. */
-        if(P[other].ID != I->base.ID)
-            lock_spinlock(other, WIND_GET_PRIV(lv->tw)->spin);
+        double dir[3];
+        get_wind_dir(other, dir);
+        int j;
+        for(j = 0; j < 3; j++)
+        {
+            #pragma omp atomic update
+            P[other].Vel[j] += v * dir[j];
```
rainwoodman: I think this is different from the old behavior. the scope of the critical region has reduced -- you are using 3, not 1. 

Were we only locking the other particle here to make sure velocity change is atomic?

rainwoodman:
Let me think about this a bit more over the day.

libgadget/blackhole.c
sbird: I will try to improve this, at least with a comment

sbird: Swallowid is set when swallowing is in progress, swallowed is set when swallowing completes. Things gravitate up to swallowing being completed.

libgadget/winds.c
sbird: In the isotropic winds case yes - the complex case was the non-isotropic wind, which uses the current velocity to work out the new wind direction. I tried to think of clever ways to get around it but since we don't use it and most people have moved away from it I thought: why bother?

rainwoodman: I thought the non-isotropic wind only uses (should only use) a reference direction computed from nearby DM particles or something like that, to compute the kick momentum, which is then added to the current velocity of P[other] just like this.

The kick momentum shouldn't depend on the current velocity of other, then?

libgadget/blackhole.c
rainwoodman: Thanks. This makes sense -- 'Swallowed' is after it is done. Perhaps SwallowingID? ;)

sbird: hehe :-)

libgadget/winds.c
sbird: In principle I agree that the non-isotropic wind should do something like that, but what it actually does is that the wind direction is set by (v - vmean) . a   

vmean is set from the nearby DM particles and a is the short-range gravitational force. 

I thought about this and...I rather feel that this is essentially a random number: v may depend on whether the particle has already been kicked, and the short-range force is likely very much not the total acceleration. Since the physics seem at best confused, I thought maybe best to just remove it.

sbird:
I'm worried that this doesn't actually do what I think it does. Looking at the godbolt it looks like this exchange isn't actually using an atomic instruction except for the final swap. It seems that it does the ? outside the lock and then only the swap inside the lock https://gcc.godbolt.org/z/9dNLQk . If that is the case I need to use a read/modify/update algorithm with a capture. Or am I misunderstanding x86 assembly?

sbird:
I'm dropping this. Although it seems to work, I think it is actually not allowed by the openmp standard. I can't tell if this means that it works by chance, or that it doesn't work (if there is a race it is very narrow). The standard says:
"Neither of v and expr (as applicable) may access the storage location designated by x."
which I think means this is not allowed.

PR #395, opened 2020-01-31
Nospinlocks

sbird:
This is attempt 2 at removing the spinlocks. It is documented to work (unless I have made a bug). It is a compare exchange algorithm which will work best when the memory is uncontended (often for us). Technically it is a gcc extension (from v 5) but it looks like icc and clang both support it on the versions that we care about. If we find a compiler that doesn't support it, I may back this out.

### DIFF
--- a/libgadget/blackhole.c
+++ b/libgadget/blackhole.c
@@ -100,8 +100,6 @@ struct BHPriv {
      * in the feedback treewalk*/
     MyFloat * BH_FeedbackWeightSum;
 
-    /* Particle SpinLocks*/
-    struct SpinLocks * spin;
     /* Counters*/
     int64_t * N_sph_swallowed;
     int64_t * N_BH_swallowed;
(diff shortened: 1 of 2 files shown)


sbird:
By the way I can't use C11 atomics because they need to operate on an Atomic type, not arbitrary memory.

rainwoodman:
That's alarming. You probably need to declare atomic (somewhat more restrictive than volatile) to avoid the compiler from doing crazy things. 

It may be a good idea to declare the field atomic to shout out it is special anyway.
BTW, are we already requiring c11?

rainwoodman:
IBM says it only needs volatile:
https://www.ibm.com/support/knowledgecenter/en/SSLTBW_2.1.0/com.ibm.zos.v2r1.cbclx01/atomic_func_comp_exch.htm

The macro definition in glibc doesn't seem to require any special prefix:
https://github.com/gcc-mirror/gcc/blob/b92709388b9aee52db3cd7c227818e0ecccda2e5/gcc/ginclude/stdatomic.h#L160

sbird:
The docs say "A: One of the atomic types." As long as the base type is something the processor can hardware lock (ie, <= 64-bit) it compiles to the same thing, but if it is a more complex structure the compiler will add a mutex which will change sizeof. 

Using C11 atomics to do this does actually compile fine in gcc, but not in clang (I didn't try icc). This makes me think that it is technically a compiler bug and may behave oddly. I don't think we require C11 elsewhere, just C99.

sbird:
It shouldn't be Atomic because then *any* access needs to be an atomic one, whereas we only care during this very short parallel section where they may be edited by multiple threads.

PR #396, opened 2020-02-02
Bhdiet

sbird:
Some modifications to the black hole model to avoid overweight black holes


sbird:
Fixes it, will merge when travis goes green

PR #397, opened 2020-02-05
Pairwise

sbird:
This is the PR to track work on a pairwise force in short timesteps.


sbird:
This is a huge speedup! 

On my small test case the improvement to the time for Moments is balanced by slower hmaxUpdate. I will try to mitigate that if it persists on a bigger box.

The main benefit though is that it seems the Pairwise gravity computation is much faster than the tree gravity walk for small particle numbers! I guess all the branching is bad for the processor. We shall see if this persists for larger boxes.

Current hmax changes make hydro slightly faster (but the load-balancing for density slightly worse) but seem positive overall.

sbird:
Alas, something seems to have been wrong with that test. This is a moderate speedup!

sbird:
I think I'm ready to merge this - we get an overall speedup of 10% of walltime in a hydro run by skipping the moment build on small timesteps. Not bad. @rainwoodman Do you want to take a look at the last few commits from ddeff4b ? The heavy lifting is all done by the code you reviewed already.

PR #399, opened 2020-02-21
Localvars

sbird:
This PR removes all the uses of allvars.h outside of libgadget, because I was getting tired of link-time optimization miscompiling everything every time the size of the struct changed.

In particular this moves all the libgenic configuration into genic_config.

### DIFF
--- a/gadget/main.c
+++ b/gadget/main.c
@@ -98,7 +98,7 @@ int main(int argc, char **argv)
     }
 
     if(RestartFlag == 1) {
-        RestartSnapNum = find_last_snapnum();
+        RestartSnapNum = find_last_snapnum(All.OutputDir);
         message(0, "Last Snapshot number is %d.\n", RestartSnapNum);
     }
 
--- a/gadget/main.c
+++ b/gadget/main.c
@@ -7,8 +7,8 @@
 #include <math.h>
 #include <gsl/gsl_math.h>
 #include <gsl/gsl_errno.h>
+#include <omp.h>
 
-#include <libgadget/allvars.h>
 #include <libgadget/slotsmanager.h>
 #include <libgadget/partmanager.h>
 
@@ -74,7 +74,9 @@ int main(int argc, char **argv)
 
     tamalloc_init();
 
-    read_parameter_file(argv[1]);	/* ... read in parameters for this run */
+    int ShowBacktrace;
+    double MaxMemSizePerNode;
+    read_parameter_file(argv[1], &ShowBacktrace, &MaxMemSizePerNode);	/* ... read in parameters for this run */
 
     int RestartFlag, RestartSnapNum;
 
@@ -98,23 +100,23 @@ int main(int argc, char **argv)
     }
 
     if(RestartFlag == 1) {
-        RestartSnapNum = find_last_snapnum(All.OutputDir);
-        message(0, "Last Snapshot number is %d.\n", RestartSnapNum);
+        /* Last snapshot will be detected in begrun*/
+        RestartSnapNum = -2;
     }
 
     /*Set up GSL so it gives a proper MPI termination*/
     gsl_set_error_handler(gsl_handler);
 
     /*Initialize the memory manager*/
-    mymalloc_init(All.MaxMemSizePerNode);
+    mymalloc_init(MaxMemSizePerNode);
 
     /* Make sure memory has finished initialising on all ranks before doing more.
      * This may improve stability */
     MPI_Barrier(MPI_COMM_WORLD);
 
-    init_endrun(All.ShowBacktrace);
+    init_endrun(ShowBacktrace);
 
-    begrun(RestartSnapNum);
+    RestartSnapNum = begrun(RestartSnapNum);
 
     switch(RestartFlag) {
         case 3:
--- a/gadget/main.c
+++ b/gadget/main.c
@@ -99,11 +99,6 @@ int main(int argc, char **argv)
         endrun(0, "Need to give the snapshot number if FOF is selected for output\n");
     }
 
-    if(RestartFlag == 1) {
-        /* Last snapshot will be detected in begrun*/
-        RestartSnapNum = -2;
-    }
-
     /*Set up GSL so it gives a proper MPI termination*/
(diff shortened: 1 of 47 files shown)


gadget/main.c
```diff
@@ -98,23 +100,23 @@ int main(int argc, char **argv)
...
 
     /* Make sure memory has finished initialising on all ranks before doing more.
      * This may improve stability */
     MPI_Barrier(MPI_COMM_WORLD);
 
-    init_endrun(All.ShowBacktrace);
+    init_endrun(ShowBacktrace);
 
-    begrun(RestartSnapNum);
+    RestartSnapNum = begrun(RestartSnapNum);
```
rainwoodman: What about passing RestartSnapFlag annd RestartSnapNum both to begrun to avoid defining a new magic number -2?

libgadget/checkpoint.h
```diff
@@ -3,8 +3,8 @@
 
 #include "forcetree.h"
 
-void write_checkpoint(int WriteSnapshot, int WriteFOF, ForceTree * tree);
-void dump_snapshot(const char * dump);
-int find_last_snapnum(void);
+void write_checkpoint(int snapnum, int WriteSnapshot, int WriteFOF, double Time, const char * OutputDir, const char * SnapshotFileBase, const int OutputDebugFields, ForceTree * tree);
```
rainwoodman: is it necessary to declare in const?

libgadget/drift.c
```diff
@@ -13,9 +12,9 @@
 #include "partmanager.h"
 #include "utils.h"
 
-static void
-real_drift_particle(int i, inttime_t ti1, const double ddrift, const double random_shift[3]);
+static void real_drift_particle(int i, inttime_t ti1, const double ddrift, const double BoxSize, const double random_shift[3]);
 
+#if 0
```
rainwoodman: Disabled because ??
Or perhaps explain when this may need to be reenabled.

libgadget/drift.c
```diff
@@ -138,21 +138,21 @@ static void real_drift_particle(int i, inttime_t ti1, const double ddrift, const
 }
 
 /* Update all particles to the current time, shifting them by a random vector.*/
-void drift_all_particles(inttime_t ti1, const double random_shift[3])
+void drift_all_particles(inttime_t ti1, const double BoxSize, Cosmology * CP, const double random_shift[3])
```
rainwoodman: What about combinding BoxSize, CP, random_shift into an object? e.g. TimeIntegrator?

libgadget/run.c
```diff
@@ -79,8 +80,13 @@ close_outputfiles(void);
  *  parameterfile is set, then routines for setting units, reading
  *  ICs/restart-files are called, auxialiary memory is allocated, etc.
  */
-void begrun(int RestartSnapNum)
+int begrun(int RestartSnapNum)
 {
+    if(RestartSnapNum == -2) {
```
rainwoodman: Or perhaps define RESTART_SNAPNUM_AUTO -2

libgadget/stats.c
```diff
@@ -73,7 +73,7 @@ struct state_of_system compute_global_quantities_of_system(void)
 
         if(P[i].Type == 0)
         {
-            struct UVBG uvbg = get_local_UVBG(redshift, P[i].Pos, All.CurrentParticleOffset);
+            struct UVBG uvbg = get_local_UVBG(redshift, P[i].Pos, CurrentParticleOffset);
```
rainwoodman: Is CurrentParticleOffset related to the smearing (randomShift)?

libgadget/winds.c
```diff
@@ -254,61 +286,55 @@ winds_evolve(int i, double a3inv, double hubble)
 static void
 sfr_wind_weight_postprocess(const int i, TreeWalk * tw)
 {
+    int done = 0;
     if(P[i].Type != 4)
         endrun(23, "Wind called on something not a star particle: (i=%d, t=%d, id = %ld)\n", i, P[i].Type, P[i].ID);
-    int diff = WINDP(i).Ngb - 40;
+    struct winddata * Windd = WIND_GET_PRIV(tw)->Winddata;
```
rainwoodman: CamelCases tends to invoke imaginations of global variables.

As a local var name it can be confusing, but this trend started with the config variables.

libgadget/drift.c
sbird: It is unused and I didn't want to waste time fixing it. I would be happy to delete it too, but I keep it around as an example of how the locking should work.

gadget/main.c
sbird: Yes, good idea.

libgadget/stats.c
sbird: Yes - the local UVBG needs to use the real physical position, not the shifted one.

libgadget/checkpoint.h
sbird: Not sure I understand?

libgadget/drift.c
sbird: This suggests a better idea: move CP inside PartManager

libgadget/drift.c
sbird: Deleted

libgadget/drift.c
sbird: I might move BoxSize in there as well next time and then make the petaio functions take a single argument which is a struct part_manager

rainwoodman: The name TimeIntegrator can smoothly related to the behaviours -- a time integrator drifts and kicks particles; so it's a reusable interface instead of a bare collection of plain old data. 

Combining interfaces into PartManager is producing a bigger interface, which usually feel somewhat uncomfortable. But rules are there to be broken..

sbird: I'm sorry, I meant random_shift, not CP! random_shift is not really a property of the time integration: it is a property of the particle table and as such belongs in the PartManager.

sbird: This does also suggest moving Time et al into CP - maybe in a future pull.

sbird:
Merging minus one commit which changed output and needs further investigation.

PR #401, opened 2020-02-26
Fix load balance

sbird:
I did a quick test and I found that our load balancing is making the final time to solution for the simulation substantially *worse*.

Generally it looks like the tree was helped by the load balancing (the wait times were shorter), but the density and hydro were hurt and this dominated. Setting HydroCostFactor = 0 by default helped, but I could not do better than a simple particle load balance! It might be that our shift towards doing more gravity in the PM code means the load balancing has too few samples to work.

@rainwoodman  what do you think? The total time to solution drops by about 10% with this PR.

### DIFF
--- a/libgadget/domain.c
+++ b/libgadget/domain.c
@@ -414,8 +414,10 @@ domain_balance(DomainDecomp * ddecomp)
 
     walltime_measure("/Domain/Decompose/Sumcost");
 
-    /* first try work balance */
-    domain_assign_balanced(ddecomp, TopLeafWork, 1);
+    /* Use a particle load balance instead of work balance.
+     * I was unable to find a workload where the work balance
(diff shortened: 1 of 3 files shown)


sbird:
I think another reason why the load balancing might not be working is that we now randomly shift the particle positions every timestep. The whole point of this is to ensure that the tree opens a different amount and thus avoid correlated errors, but it means that the number of interactions during one domain decomposition is not predictive of the number of interactions during the next one.

I vote for ripping out the GravCost infrastructure completely and just particle load balancing everything, I think.

rainwoodman:
I am convinced in short.

sbird:
Ok.

I will leave the "cost balancing" part of the domain decomposition mostly intact (just change the policy) so that if we think of a clever way to load balance the processes we don't have to reimplement it.

PR #403, opened 2020-02-28
change long MFP heating to scale with mass

uptonsanderbeck:


libgadget/cooling.c
```diff
@@ -59,7 +59,7 @@ get_lambdanet(double rho, double u, double redshift, double Z, struct UVBG * uvb
     if(!isHeIIIionized) {
         /* get_long_mean_free_path_heating returns the heating in units of erg/s/cm^3,
          * the factor of rho converts to erg/s/proton and then PROTONMASS to erg/s/g */
-        LambdaNet += get_long_mean_free_path_heating(redshift)  / (rho  * PROTONMASS);
+        LambdaNet += get_long_mean_free_path_heating(redshift)  / (3 * OmegaBaryon * pow((HUBBLE * HubbleParam), 2) / (8 * M_PI * GRAVITY));
```
sbird: Could you fix the comment as well to explain why rho_c? It also, er, needs to compile!

sbird: Perhaps pass rho_c as an argument to get_lambdanet

PR #405, opened 2020-03-06
Lmfp heating

uptonsanderbeck:


uptonsanderbeck:
I don't know if putting rho_crit_baryon in init_cooling in cooling.c is the best place for it, but it seemed the least invasive way to implement it.

libgadget/cooling.h
```diff
@@ -30,6 +30,7 @@ struct cooling_units
 
 /*Initialise the cooling module.*/
 void init_cooling(char * TreeCoolFile, char * MetalCoolFile, char * reion_hist_file, struct cooling_units cu, Cosmology * CP);
+double rho_crit_baryon;
```
sbird: Could you please make this a member of cooling_units?

uptonsanderbeck: Done, is this what you had in mind?

sbird:
Yup, looks good! I will merge when travis goes green

PR #413, opened 2020-03-20
Longtimeline

sbird:
This extends the timeline variable to 64 bits, which we have been flirting with for a while. With the 6k^3 simulation I think it is time to really do it. Of course, if we get to more than 10 timesteps deep it will still take a long time to get anywhere...but the overhead is much reduced now we don't need to build the tree.


sbird:
Changes output slightly but does not crash. I think that is expected because the integer timeline will be subdivided slightly differently. Shall I merge this?

sbird:
Probably dropping this because the timestepping changes mean the timesteps become less deep.

PR #414, opened 2020-03-23
Softeningcleanup

sbird:
Cleanup the force softening code and remove a parameter from the Node structure. This removes the ability to have different softening parameters for different particle types, which is probably a bad idea from the point of view of force accuracy.


sbird:
Identical output, very tiny speed up in the tree.

PR #417, opened 2020-04-03
Largedebugmerge

sbird:
This is the largedebug PR with the actual debugging code stripped out. For merge, fixes some bugs.


sbird:
Verified that this makes zero difference to output on 1 thread.

PR #419, opened 2020-04-08
Guidedtreewalk

sbird:
Some work to reduce the treewalk imbalance time by smarter scheduling.

### DIFF
--- a/libgadget/treewalk.c
+++ b/libgadget/treewalk.c
@@ -410,13 +410,10 @@ treewalk_build_queue(TreeWalk * tw, int * active_set, const size_t size, int may
 static void
 ev_primary(TreeWalk * tw)
 {
-    const int NTask = tw->NTask;
     double tstart, tend;
     tw->BufferFullFlag = 0;
     tw->Nexport = 0;
 
-
-    size_t i;
     tstart = second();
 
     struct TreeWalkThreadLocals export = ev_alloc_threadlocals(tw, tw->NTask, tw->NThread);
@@ -669,7 +666,7 @@ static struct SendRecvBuffer ev_get_remote(TreeWalk * tw)
     /* Can I avoid doing this sort?*/
     qsort_openmp(DataIndexTable, tw->Nexport, sizeof(struct data_index), data_index_compare);
     int NTask = tw->NTask;
-    int i;
+    size_t i;
     double tstart, tend;
 
     struct SendRecvBuffer sndrcv = {0};
--- a/libgadget/treewalk.c
+++ b/libgadget/treewalk.c
@@ -185,6 +185,9 @@ ev_begin(TreeWalk * tw, int * active_set, const size_t size)
      * sfr/bh we should change this*/
     treewalk_build_queue(tw, active_set, size, 0);
 
+    /* Start first iteration at the beginning*/
+    tw->WorkSetStart = 0;
+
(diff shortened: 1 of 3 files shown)


sbird:
On a simple 24 thread 1 MPI testcase this is 12% faster. Not bad!

libgadget/treewalk.c
```diff
@@ -567,11 +543,12 @@ int treewalk_export_particle(LocalTreeWalk * lv, int no) {
         }
         /* out of buffer space. Need to discard work for this particle and interrupt */
         if(nexp >= tw->BunchSize) {
-            #pragma omp atomic write
-            tw->BufferFullFlag = 1;
-
+            /* Mark the first not fully exported target*/
+            if(nexp == tw->BunchSize)
+                tw->FirstNotExported = target;
```
rainwoodman, line 548: FirstFailedExport?

libgadget/treewalk.c
```diff
@@ -623,14 +600,26 @@ treewalk_run(TreeWalk * tw, int * active_set, size_t size)
...
+            const int p_i = tw->WorkSet ? tw->WorkSet[i] : i;
             tw->preprocess(p_i, tw);
         }
     }
 
+    /* First run does all the local interactions*/
+    tw->exportonly = 0;
+    tw->FirstNotExported = 0;
+
+    /* The first iteration of this loop does all local interactions
```
rainwoodman, line 615: I am a bit confused about first loop does all local interactions subsequence skips all local interactions. It appears that ev_primary supposed to be doing all local particles, and ev_second supposed to be only working on imported particles?

It seems to be clearer if there is a way to life the export logic from ev_primary to ev_get_remove(perhaps change name to ev_exchange). Then the loop will explicitly follow what you described here in the comments:

ev_primary(....)
while {
 ev_get_remote()
 ev_secondary()
}

And perhaps the export-only flag and FirstNotExported can be simplified to some extent.

sbird:
Treat this as a WIP for now as I found it has a problem: there may be some particle exports missed due to multithreading when the buffer fills up. I think at the moment we have some particle exports which are done twice when the buffer fills up.

libgadget/treewalk.c
sbird: I like that idea!

sbird: I wondered about splitting it like "do_toptree_walk" which finds exports and then doing the rest of the walk in a separate function.

rainwoodman: Do you plan to use the same `do_toptree_walk` for other activities than exporting?
The usual distributed narrative is "export", "work", "reduce" -- in this narrative, whether the export is from toplevel tree walk or something else is a detail.

sbird: I was toying with the idea that the do_toptree_walk would also specify which of the local toptree nodes would be used for walking. I think though that "export", "work" and "reduce" are the right verbs and I will use them in the writeup!

sbird:
A quick test on a small dataset suggests that  this: a) works and b) does speed things up (although the imbalance is small on the small dataset).

sbird:
With the export buffer artificially reduced to 20MB so that it almost always fills up, this works and reduces imbalance substantially. Doesn't really seem to change output.

sbird:
@yueyingn Can you see if this helps with your problem? @rainwoodman review?

PR #421, opened 2020-04-08
Remove the extra tree evaluation on the first step

sbird:
This is done so the error estimate is consistent. This was
always a small effect, I think, and has become much smaller since we
started randomising the particle positions after every domain.
So save us some time on startup.

### DIFF
--- a/libgadget/run.c
+++ b/libgadget/run.c
@@ -456,15 +456,6 @@ void compute_accelerations(const ActiveParticles * act, int is_PM, PetaPM * pm, 
             energy_statistics(FdEnergy, All.Time, PartManager);
     }
 
-    /* For the first timestep, we do tree force twice
-     * to allow usage of relative opening
-     * criterion for consistent accuracy.
-     * This happens after PM because we want to
-     * use the total acceleration for tree opening.
-     */
-    if(FirstStep && All.TreeGravOn)
-        grav_short_tree(act, pm, tree, rho0, NeutrinoTracer, All.FastParticleType);
-
     MPIU_Barrier(MPI_COMM_WORLD);
     message(0, "Forces computed.\n");
 }


sbird:
@rainwoodman ?

rainwoodman:
Sorry for getting this late. Do you have some numbers that shows the force error of one-force+randomization vs two-forces as vs two-force+randomization? I can sort of picture randomizing particle locations reduce the force error pattern, but I am having a hard time picturing it changes accuracy control.

sbird:
I don't have the numbers to hand, but essentially the tree opening parameters are set up so the second treewalk evaluation always makes the force error *worse*: the point of it is to make it consistently worse, but since we randomise the positions it is not consistent anymore.

rainwoodman:
I guess the main reason for consistently worse is such that restarts are
not introducing inconsistent steps.

I recalled a bit more -- the opening angle (or something like that) is a
function of the last step acceleration. On a restart the last acceleration
is zero, requring all nodes to be open, making the short range force
'exact'. This is true whether the tree is randomized or not.

Yu

sbird:
That's right - the idea is that the extra tree evaluation on the first timestep is that it makes doing the evolution of the code with and without a restart more similar, at the cost of making the force error on the first timestep slightly worse than it otherwise would be.

But I think that our randomized particle positions will introduce a larger difference on restarts than doing the first timestep exactly. I need to confirm this with a test.

sbird:
The effect of this is very very small. It is much smaller than the effect of enabling threads and also much smaller than the effect of position randomization on restart.

sbird:
The maximum position shift from this is O(1e-5) kpc, while a normal restart shifts max final positions by O(0.1) kpc. Threads shift final max restart by O(1) kpc. This is for a small DM only run.

PR #422, opened 2020-04-10
Cooling strang split

sbird:
This PR follows upstream Gadget-3 and Arepo in computing the cooling and BH operators after the kicks. It is the recommended way of permanently solving the short timestep problem, because the cooling and BH operators now happen at a consistent time and do not affect DtEntropy (so the predicted entropy in hydra is not off), rather changing Entropy directly.


sbird:
On a small run this has a moderate effect: somewhat reduces the number of timesteps by ~5% although slightly increasing the runtime (by 2%). Honestly the impact is really consistent with the normal variance between threaded simulations, but I will merge it anyway because it will bring us in line with upstream and because the step reduction may be more apparent in larger runs.

PR #423, opened 2020-04-10
Add an example script to rebin the power spectrum

sbird:
Counts the modes at each bin and so smooths out the jagged parts.

### DIFF
--- a/tools/rebin_powerspectrum.py
+++ b/tools/rebin_powerspectrum.py
@@ -1,0 +1,42 @@
+"""Small module to rebin the power spectrum text files output by MP-Gadget so that each bin
+has at least the desired number of modes. """
+import numpy as np
+
+def modecount_rebin(kk, pk, modes, minmodes=20, ndesired=200):
+    """Rebins a power spectrum so that there are sufficient modes in each bin"""
+    assert np.all(kk) > 0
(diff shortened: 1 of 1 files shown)


tools/rebin_powerspectrum.py
```diff
@@ -0,0 +1,42 @@
...
+def modecount_rebin(kk, pk, modes, minmodes=20, ndesired=200):
+    """Rebins a power spectrum so that there are sufficient modes in each bin"""
+    assert np.all(kk) > 0
+    logkk=np.log10(kk)
+    mdlogk = (np.max(logkk) - np.min(logkk))/ndesired
+    istart=iend=1
+    count=0
+    k_list=[kk[0]]
+    pk_list=[pk[0]]
+    targetlogk=mdlogk+logkk[istart]
```
rainwoodman, line 14: This looks rather convoluted. Why not numpy.histogram or numpy.digitize + numpy.bincount?

sbird: I would love to, but I didn't realise I can do that with np.histogram? Isn't that for finding the number of points in constant width bins? What I want to do here is to find the variable bin width such that the number of objects in each one is roughly constant.

PR #434, opened 2020-05-06
Revert ce3e0e2a84 due to major performance loss for threaded code.

sbird:
ce3e0e2a84:
    Replace final alloca calls in treewalk

    Use instead allocations on the ta malloc heap. This removes a
    substantial source of thread-local stack allocations.

Removed thread-local stack allocations. This had no effect on a pure
MPI run but costs us a factor of 2 (!) in threaded performance, making
the threaded code half as fast as the MPI code.

I am not sure why this is. I suspect it may be that the stack variable
can be optimized out entirely by the compiler, but the heap variable
cannot. It may also be some effect of the processor caches.

### DIFF
--- a/libgadget/treewalk.c
+++ b/libgadget/treewalk.c
@@ -275,8 +275,6 @@ static int real_ev(struct TreeWalkThreadLocals export, TreeWalk * tw, int * curr
     ev_init_thread(export, tw, lv);
     lv->mode = 0;
 
-    int tid = omp_get_thread_num();
-
     /* use old index to recover from a buffer overflow*/;
     TreeWalkQueryBase * input = alloca(tw->query_type_elsize);
     TreeWalkResultBase * output = alloca(tw->result_type_elsize);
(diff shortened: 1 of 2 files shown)


sbird:
@rainwoodman : Any idea why this works?

rainwoodman:
Without reading the assembly it is hard to tell.

sbird:
The assembly is actually about the same. I think the problem may have been a false cacheline share: because the temporary for each thread is allocated contiguously and the size of the temporary is < 1 page it may be that each thread write caused a cache miss for the neighbouring threads. Probably this is not an issue for the remaining writes because they are of size int * NTask so they do not share a page with other threads. Also they are much less hot because they are only touched during writes.

rainwoodman:
That's to say when we do the per thread allocation (we had a helper function), the per thread result be rounded to pages?

sbird:
Yes probably we should to be safe.

PR #435, opened 2020-05-06
Iotweaks

sbird:
I am wondering if the zeros in the Position array are not because it is large but because it is the first array written. So the recursive mkdir in big_file_mpi_create_block() has to create multiple directories. Maybe there is a race somewhere in the fs code related to writing to a directory before creation of the parent has completed?

This works on that theory. First I create the 0/ 1/ etc subdirectories separately before writing the blocks. Then in case this doesn't work I make sure the first array written is not Position but Potential or Mass and that we can recover from a corrupted Mass array.

If it is a subdirectory creation race then I think this fix in bigfile should also do it:

in big_file_mpi_create_block():

 BCAST_AND_RAISEIF(rt, comm);
+    MPI_Barrier(comm);
   
on line 109

@rainwoodman what do you think?

### DIFF
--- a/libgadget/init.c
+++ b/libgadget/init.c
@@ -131,7 +131,7 @@ set_init_params(ParameterSet * ps)
     MPI_Bcast(&All, sizeof(All), MPI_BYTE, 0, MPI_COMM_WORLD);
 }
 
-static void check_omega(void);
+static void check_omega(int generations);
 static void check_positions(void);
 
 static void
@@ -175,7 +175,7 @@ void init(int RestartSnapNum, DomainDecomp * ddecomp)
 
     domain_test_id_uniqueness(PartManager);
 
-    check_omega();
+    check_omega(get_generations());
 
     check_positions();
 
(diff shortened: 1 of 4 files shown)


rainwoodman:
It is good observation that Position is the first field. But big_block_mpi_broadcast already has an implicit barrier -- thus I don't think lifting the block creation would have any effects.

If it is the directory race, then adding a sleep of a fraction of a second in bigfile is a cleaner solution than lifting block creation.

It could be a hardware issue too, 16MB of zeros is just a huge page or a DMA buffer, something like that -- I somewhat recall we saw the same problem with MB-II on Kraken (but could be data transfer errors) and that was writing 1 file per rank. If we pin down the rank tand node hat these zeros are from ....

One application that we can run for this sort of tests is bigfile-iosim.

sbird:
That's true, there is already a barrier in BCast. A small sleep in bigfile is the better workaround indeed. What do you think about merging making Mass the first block? Maybe it will change the times enough that we don't hit whatever this is so often.

I completely agree by the way that this is a problem with the machine: writes should not fail nor should they return zero. We should get them to fix it. But I also want to explore workarounds in case we encounter other machine misconfigured in the same way.

rainwoodman:
+1 to move a different field to the top. Perhaps ID? Or add the particle types -- it is much smaller and disposable.

sbird:
Particle types is not good because the correct value is often zero, so we wouldn't notice any corruption...

sbird:
How about that? Lead with Mass because we can recover from it.

rainwoodman:
Yes Mass is good. Forgot particle types are mostly zeros.

PR #436, opened 2020-05-12
MPI test for the FOF code.

sbird:
Add a test for the FOF routines

### DIFF
--- a/libgadget/tests/test_fof.c
+++ b/libgadget/tests/test_fof.c
@@ -1,0 +1,123 @@
+/*Simple test for the exchange function*/
+
+#include <stdarg.h>
+#include <stddef.h>
+#include <setjmp.h>
+#include <cmocka.h>
+#include <math.h>
+#include <mpi.h>
+#include <stdio.h>
+#include <string.h>
+#include <time.h>
+#include <gsl/gsl_rng.h>
+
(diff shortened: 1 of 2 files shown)


libgadget/tests/test_fof.c
```diff
@@ -0,0 +1,123 @@
...
+    /* Example assertion: this checks that the address of a struct is not NULL,
+     * which will always pass! */
+    assert_all_true(&fof);
+
+    MPI_Comm_size(MPI_COMM_WORLD, &NTask);
+    MPI_Comm_rank(MPI_COMM_WORLD, &ThisTask);
+
+    /* Assert some more things about the particles,
+     * maybe checking the halo properties*/
+    for(i = 0; i < PartManager->NumPart; i ++) {
```
rainwoodman, line 106: 1. Is it possible to control the number of threads (for testing the other PR).

2. I recommend testing two cases: 1 particle per halo with very short linking length. and all particles in the same halo with very long linking length.

The halo properties are straight-forward in both cases, and I caught plenty bugs with these two in kdcount/kd_fof.

sbird: Those are excellent suggestions! @mahdiqezlou is implementing this as an exercise to get him into the code, so he can follow them.

PR #437, opened 2020-05-13
Use compare exchange in the FoF code

sbird:
This PR cleans up the locking in the FoF code. I noticed there was a catastrophic performance dropoff when we switched to dynamic treewalking, because particles in the same halos were on different threads and there was massive contention. This fixes it by moving to a lockless algorithm and by increasing the chunk size. Also fix a locking bug in the secondary treewalk.

### DIFF
--- a/libgadget/fof.c
+++ b/libgadget/fof.c
@@ -505,16 +505,17 @@ fof_primary_ngbiter(TreeWalkQueryFOF * I,
         }
     } else /* mode is 1, target is a ghost */
     {
-            struct SpinLocks * spin = FOF_PRIMARY_GET_PRIV(tw)->spin;
+        int head = HEAD(other, tw);
+        struct SpinLocks * spin = FOF_PRIMARY_GET_PRIV(tw)->spin;
 //        printf("locking %d by %d in ngbiter\n", other, omp_get_thread_num());
-        lock_spinlock(other, spin);
-        if(HaloLabel[HEAD(other, tw)].MinID > I->MinID)
+        lock_spinlock(head, spin);
+        if(HaloLabel[head].MinID > I->MinID)
         {
-            HaloLabel[HEAD(other, tw)].MinID = I->MinID;
-            HaloLabel[HEAD(other, tw)].MinIDTask = I->MinIDTask;
+            HaloLabel[head].MinID = I->MinID;
+            HaloLabel[head].MinIDTask = I->MinIDTask;
         }
 //        printf("unlocking %d by %d in ngbiter\n", other, omp_get_thread_num());
-        unlock_spinlock(other, spin);
+        unlock_spinlock(head, spin);
     }
 }
 
--- a/libgadget/fof.c
+++ b/libgadget/fof.c
@@ -315,7 +315,10 @@ update_root(int i, const int r, int * Head)
             t = Head[i];
             Head[i]= r;
         }
-    } while(t != i);
+        /* Stop if we reached the top (new head is the same as the old)
+         * or if the new head is less than or equal to the desired head, indicating
+         * another thread changed us*/
+    } while(t != i && (t > r));
 }
 
 /* Find the current head particle by walking the tree. No updates are done
--- a/libgadget/fof.c
+++ b/libgadget/fof.c
@@ -321,17 +321,17 @@ update_root(int i, const int r, int * Head)
 /* Find the current head particle by walking the tree. No updates are done
  * so this can be performed from a threaded context. */
 static int
-HEAD(int i, TreeWalk * tw)
+HEAD(int i, const int * const Head)
 {
     int r = i;
-    while(FOF_PRIMARY_GET_PRIV(tw)->Head[r] != r) {
-        r = FOF_PRIMARY_GET_PRIV(tw)->Head[r];
+    while(Head[r] != r) {
+        r = Head[r];
     }
     return r;
 }
 
 static void fof_primary_copy(int place, TreeWalkQueryFOF * I, TreeWalk * tw) {
-    int head = HEAD(place, tw);
+    int head = HEAD(place, FOF_PRIMARY_GET_PRIV(tw)->Head);
     I->MinID = HaloLabel[head].MinID;
     I->MinIDTask = HaloLabel[head].MinIDTask;
 }
@@ -407,7 +407,8 @@ void fof_label_primary(ForceTree * tree, MPI_Comm Comm)
         link_across = 0;
 #pragma omp parallel for reduction(+: link_across)
         for(i = 0; i < PartManager->NumPart; i++) {
-            MyIDType newMinID = HaloLabel[HEAD(i, tw)].MinID;
+            int head = HEAD(i, FOF_PRIMARY_GET_PRIV(tw)->Head);
+            MyIDType newMinID = HaloLabel[head].MinID;
             if(newMinID != FOF_PRIMARY_GET_PRIV(tw)->OldMinID[i]) {
                 FOF_PRIMARY_GET_PRIV(tw)->PrimaryActive[i] = 1;
                 link_across ++;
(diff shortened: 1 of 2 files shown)


libgadget/fof.c
```diff
@@ -401,13 +378,25 @@ void fof_label_primary(ForceTree * tree, MPI_Comm Comm)
         treewalk_run(tw, NULL, PartManager->NumPart);
 
         t1 = second();
-
+        /* Just in case we missed some during HEAD building.*/
```
rainwoodman, line 381: If it is abnormal, what about adding an error or a warning?
It it is normal, then briefly describe how it happens?

libgadget/fof.c
```diff
@@ -401,13 +378,25 @@ void fof_label_primary(ForceTree * tree, MPI_Comm Comm)
...
+            }
+        }
         /* let's check out which particles have changed their MinID,
          * mark them for next round. */
         link_across = 0;
 #pragma omp parallel for reduction(+: link_across)
         for(i = 0; i < PartManager->NumPart; i++) {
-            MyIDType newMinID = HaloLabel[HEAD(i, tw)].MinID;
+            int head = HEAD(i, FOF_PRIMARY_GET_PRIV(tw)->Head);
+            if(i != head) {
```
rainwoodman, line 395: This loop is also changing the MinID -- but the comment above still says it is only checking changed particles.

What is the difference between this minid assignment and the 'just in case' above? Looks like they are just overwriting each other. Can they be merged?

libgadget/fof.c
```diff
@@ -442,33 +424,38 @@ fofp_merge(int target, int other, TreeWalk * tw)
...
+        if(h1 > h2) {
+            int tmp = h2;
+            h2 = h1;
+            h1 = tmp;
         }
-        //printf("unlocking %d by %d in merge\n", h2, omp_get_thread_num());
-        unlock_spinlock(h2, spin);
+     /* Atomic compare exchange to make h2 a subtree of h1.
+      * Set Head[h2] = h1 iff Head[h2] is still h2. Otherwise loop.*/
+    } while(!__atomic_compare_exchange(&Head[h2], &h2, &h1, 0, __ATOMIC_RELAXED, __ATOMIC_RELAXED));
```
rainwoodman, line 443: Do we need to mark Head, h2 and/or h1 volatile?

libgadget/fof.c
```diff
@@ -442,33 +424,38 @@ fofp_merge(int target, int other, TreeWalk * tw)
...
-            HaloLabel[h1].MinID = HaloLabel[h2].MinID;
-            HaloLabel[h1].MinIDTask = HaloLabel[h2].MinIDTask;
+        h1 = HEADl(-1, target, Head);
+        /* Done if we find h1 along the path
+         * (because other is already in the same halo) */
+        h2 = HEADl(h1, other, Head);
+        if(h2 < 0)
+            return;
+        /* Ensure that we always merge to the lower entry*/
+        if(h1 > h2) {
```
rainwoodman, line 436: I somehow thought we tried this before an run into a bug. Do you recall an event like that?

libgadget/fof.c
```diff
@@ -442,33 +424,38 @@ fofp_merge(int target, int other, TreeWalk * tw)
...
+     /* Atomic compare exchange to make h2 a subtree of h1.
+      * Set Head[h2] = h1 iff Head[h2] is still h2. Otherwise loop.*/
+    } while(!__atomic_compare_exchange(&Head[h2], &h2, &h1, 0, __ATOMIC_RELAXED, __ATOMIC_RELAXED));
+
+    struct SpinLocks * spin = FOF_PRIMARY_GET_PRIV(tw)->spin;
+
+    /* update MinID of h1: h2 is now just another child of h1
+     * so h2 does not need to be locked.
+     * h1 may not be its own head anymore,
+     * so we need to double check later.
```
rainwoodman, line 450: If we need to double check later, then perhaps the update can be entirely deferred? I feel this is possible as the loop above is no longer touching the MinID and MinTask.

rainwoodman: If it cannot be entirely deferred (e.g. interaction with the remote particles on mode==1), then we must be relying on some less level of correctness here? Would the catch later be still correct for the remote particles?

rainwoodman:
Hi! first round,  a few quick questions before I can comprehend the changes better.

libgadget/fof.c
sbird: I don't think so - the C standard doesn't seem to require it with atomic_compare_exchange. It would be Head that would be volatile, as the others are local variables. I understood that volatile basically means that the memory may be altered by another process and is used to avoid the compiler optimizing out pointless loops. I think that with openmp threads it is not necessary.

libgadget/fof.c
sbird: It was "just in case" when I introduced it but in the final commit it comes from a fairly normal race. I will add a comment.

libgadget/fof.c
sbird: The difference is that the above case sets the MinID of the HEAD to min(ID of all particles). This sets the ID of the particles to the minID of the HEAD. So they must be separate loops.

libgadget/fof.c
sbird: Yes, we tried it before and hit a deadlocking bug: we were taking mutexes on both h1 and h2 and had to be careful of the order. Because this is lockless we don't have that problem anymore. There is still a circular reference problem, which is why the swap. I will add a comment.

libgadget/fof.c
sbird: The update can be entirely deferred, indeed. So we would lose all the locking in fofp_merge. 
Both the current code and the catch later are correct, I think, but I am not sure which performs better.

The issue is exactly interaction with the remote particles: I am concerned that we would end up with a lot more updates from the remote and so more contention. I might run a test on a large sized problem.

sbird: I wanted to use atomic or compare-exchange on MinID/MinIDTask but it needs to be a scalar variable.

sbird:
My small test was doing fof on a z=4 snapshot from the hydro example. There were no changes to the FOF output and this PR restores the threaded time to about the same as the MPI time.

libgadget/fof.c
sbird: It does seem a little slower if I entirely defer the update loop, although this is a small problem and the difference is only a few percent. I think I might leave that to a future PR, maybe to use thread-locals in fof and remove the spinlocks completely.

sbird:
On a larger 1024^3 test it still works. There were a few single-particle changes in the smallest halos, but that could be ordering and floating point roundoff.

sbird:
hmm, no, there is a race condition somewhere. The new code with MPI is identical output to the old, the new code with threads is a bit different.

libgadget/fof.c
rainwoodman: Thanks. I see.

libgadget/fof.c
rainwoodman: There is a DWCAS on x86 that gcc seems to support (128 bit). Declaring the type as int128_t could probably work out of box. 

I compiled the code at:
https://gcc.gnu.org/legacy-ml/gcc-help/2017-06/msg00156.html

and it seems to be reasonable, calling "__atomic_compare_exchange_16@LIBATOMIC_1.0".

sbird: Brilliant idea! I see also with godbolt that it works for icc > 17 and clang as well. I will implement this. Is the aligned attribute necessary?

sbird:
See PR https://github.com/MP-Gadget/MP-Gadget/pull/438

libgadget/fof.c
sbird: Alas, this makes it slower and more unreliable.

sbird:
I think there is no race, just that the earlier variant was actually completely serialized and now there is the usual inter-thread variation at the margins. Removing the lock for minID in the primary treewalk does make things a bit slower at scale. I think merging.

PR #438, opened 2020-05-16
Fofcmpxchg16

sbird:
This implements compare exchange on the final spinlock in fof, using the __int128 types suggested by @rainwoodman . I'm a little worried about portability issues.

The example code from the gcc mailing list is here: https://godbolt.org/z/ntEkDG
icc generates a straight cmpxchg16b. clang generates the instruction if -mcx16 is passed to it. gcc though always generates a call to a library, libatomic (which needs an extra link flag not mentioned in the docs). There is some indication that the call actually does a mutex, ie reduces this to a critical section, which would be worse than the spinlock. See: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=80878

In practice all our production runs are done with icc so this may not be too much of a problem, but I am concerned about relying on something that differs on different compilers. @rainwoodman what do you think? I could also have some magic calls to __atomic_is_lock_free, but I don't think I want to maintain two implementations for this.

I will check performance.

### DIFF
--- a/libgadget/fof.c
+++ b/libgadget/fof.c
@@ -113,15 +113,14 @@ extern void fof_save_particles(FOFGroups * fof, int num, int SaveParticles, MPI_
 typedef struct {
     TreeWalkQueryBase base;
     MyFloat Hsml;
-    MyIDType MinID;
+    __int128 MinIDandTask;
     MyIDType MinIDTask;
 } TreeWalkQueryFOF;
 
 typedef struct {
     TreeWalkResultBase base;
     MyFloat Distance;
-    MyIDType MinID;
-    MyIDType MinIDTask;
+    __int128 MinIDandTask;
 } TreeWalkResultFOF;
 
 typedef struct {
@@ -130,11 +129,35 @@ typedef struct {
 
 static struct fof_particle_list
 {
-    MyIDType MinID;
-    MyIDType MinIDTask;
+    /* This variable is MinID * 1<<64 + MinIDTask
+     * and is defined as such to allow us to use
+     * the __atomic_exchange functions.
+     * Semantically this definition is:
+     * MyIDType MinID
+     * MyIDType MinIDTask*/
+    __int128 MinIDandTask;
     int Pindex;
 }
 *HaloLabel;
+
+/* Accessor functions*/
+static inline MyIDType MinID(__int128 MinIDandTask)
+{
+    return MinIDandTask >> 64L;
+}
+
+static inline MyIDType MinIDTask(__int128 MinIDandTask)
+{
+    return MinIDandTask % 64L;
+}
+
+static inline __int128 MinIDandTask(MyIDType MinID, MyIDType MinIDTask)
+{
+    __int128 MinIDandTask = MinID;
+    MinIDandTask <<= 64L;
+    MinIDandTask += MinIDTask;
+    return MinIDandTask;
+}
 
 static MPI_Datatype MPI_TYPE_GROUP;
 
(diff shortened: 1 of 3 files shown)


libgadget/fof.c
```diff
@@ -110,18 +110,20 @@ static void fof_assign_grnr(struct BaseGroup * base, const int NgroupsExt, MPI_C
...
 extern void fof_save_particles(FOFGroups * fof, int num, int SaveParticles, MPI_Comm Comm);
 
+/* The alignment is important for the hardware atomic instruction to work*/
+#define __aint128 __int128 __attribute__ (( aligned(16)))
+
 typedef struct {
     TreeWalkQueryBase base;
     MyFloat Hsml;
-    MyIDType MinID;
+    __aint128 MinIDandTask;
```
rainwoodman, line 119: Use an anonymous union?

rainwoodman: Or perhaps define typedef union {
... uint128_t ....
struct {
  MinID,
  Task,
} FOFLabel;

sbird: That was my first idea, but I'm not sure how to guarantee that the compiler won't pad the struct. Any change in alignment or padding would mean the hardware instruction couldn't be used and the locking could silently not be working.

libgadget/fof.c
```diff
@@ -501,16 +519,16 @@ fof_primary_ngbiter(TreeWalkQueryFOF * I,
...
-        {
-            HaloLabel[head].MinID = I->MinID;
-            HaloLabel[head].MinIDTask = I->MinIDTask;
-        }
-//        printf("unlocking %d by %d in ngbiter\n", other, omp_get_thread_num());
-        unlock_spinlock(head, spin);
+        /* update MinID.*/
+        __aint128 hminread;
+        do {
+            #pragma omp atomic read
```
rainwoodman, line 525: Is the atomic read needed? The atomic cas would probably block a partial read access?

If the atomic read is necessary for 128, then using a spinlock is definitely cleaner.

libgadget/fof.c
sbird: I think that gcc behaves correctly, but I don't think it is guaranteed by the standard (is it?).

Any future icc release that pads struct members would just silently introduce race conditions which I think we wouldn't notice

rainwoodman: Is it allowed to add the alignment (16) attribute the the union?

struct {} is all 8 bytes, so it will be tightedly packed.

libgadget/fof.c
sbird: I am not sure - I put it in only because I have been using it in the other CAS routines and it was there just for safety. If you think it is not needed I will remove it.

libgadget/fof.c
sbird: Really I have no idea! It seems like one of those standard corners where things are often wrongly implemented.

libgadget/fof.c
sbird: I see. While for the 64 bit types atomic read compiles to a no-op, for the 128 bit types it is not, but has a call to GOMP_atomic_start (which probably locks a mutex). cmpxchg16b has a lock prefix which applies to the whole cacheline so tearing reads are not possible and the atomic read isn't needed. https://gcc.godbolt.org/z/3XU2Nk (compare the two reads).

I'm still worried though that if we don't get to use cmpxchg16b for some reason, losing the atomic read would cause half reads. __atomic_load()  compiles to another cmpxchg16 (!) with icc

sbird: See the GOMP_atomic_start code: https://www.cs.rice.edu/~la5/doc/ligbomp-doc/db/d9c/critical_8c-source.html#l00125

I think I will remove the atomic read and if it is worth it performance wise, I will assert __atomic_always_lockfree(__int128) and maybe have a backup spinlock.

sbird:
Alas! This is what happens when I compile this code with icc:
": internal error: 20000_0

sbird:
It looks like the atomic read is actually *forbidden* if we want icc to compile the code!

sbird:
Benchmarking the current HEAD of this branch on frontera, it is slightly *slower* than the version with the mutex. Given that and the extremely dubious compiler support I think we drop this PR.

PR #444, opened 2020-06-17
Do not heat star forming particles with BH

sbird:
This should let high z BHs grow faster

### DIFF
--- a/libgadget/blackhole.c
+++ b/libgadget/blackhole.c
@@ -381,6 +381,9 @@ blackhole(const ActiveParticles * act, ForceTree * tree, FILE * FdBlackHoles, FI
     {
         if(P[i].Type == 0 && priv->Injected_BH_Energy[P[i].PI] > 0)
         {
+            /* Do not heat star-forming particles*/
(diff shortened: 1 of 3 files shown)


sbird:
@yueyingn What does this do? (from the skype meeting)

libgadget/blackhole.c
```diff
@@ -381,6 +381,9 @@ blackhole(const ActiveParticles * act, ForceTree * tree, FILE * FdBlackHoles, FI
     {
         if(P[i].Type == 0 && priv->Injected_BH_Energy[P[i].PI] > 0)
         {
+            /* Do not heat star-forming particles*/
```
rainwoodman, line 384: This is more like assuming feedback energy is immediately cooled away by the star forming particle.
(Since that we do not inject the BH feedback energy to other particles).

sbird: This is true! But it would be a bit messy to have the treewalk avoid star forming particles when finding neighbours (and might not converge).

sbird:
This causes too much accretion:

test-BHfdbk

sbird:
We need the "accelerated cooling" to get high z BH growth.

PR #445, opened 2020-06-18
dynamical-friction

nianyic7:
Implemented dynamical friction for BH with different options controlled by BH_DynamicalFriction in the parameter file. Still in the testing/debugging phase and minor changes need to be made.

### DIFF
--- a/libgadget/blackhole.c
+++ b/libgadget/blackhole.c
@@ -897,7 +897,7 @@ blackhole_feedback_ngbiter(TreeWalkQueryBHFeedback * I,
         TreeWalkNgbIterBHFeedback * iter,
         LocalTreeWalk * lv)
 {
-    endrun(2, "ENTERED INTO FDBKNGBITER");
+//   endrun(2, "ENTERED INTO FDBKNGBITER\n");
 //    message(0, "ENTERED INTO FDBKNGBITER");
 
     if(iter->base.other == -1) {
@@ -913,7 +913,7 @@ blackhole_feedback_ngbiter(TreeWalkQueryBHFeedback * I,
         density_kernel_init(&iter->feedback_kernel, hsearch, DENSITY_KERNEL_CUBIC_SPLINE);
         return;
     }
-    endrun(2, "SURVIVED THE FIRST RETURN");
+    endrun(2, "SURVIVED THE FIRST RETURN\n");
 //    message(0, "SURVIVED THE FIRST RETURN");
     int other = iter->base.other;
     double r2 = iter->base.r2;
@@ -930,7 +930,7 @@ blackhole_feedback_ngbiter(TreeWalkQueryBHFeedback * I,
     /* Compute fractional density for DF */
      if(P[other].Type == 1 || P[other].Type == 4) 
      {
-        endrun(2, "FOUND AT LEAST SOME DM/STAR");
+        endrun(2, "FOUND AT LEAST SOME DM/STAR\n");
 //        message(0, "FOUND AT LEAST SOME DM/STAR");
 	if(r2 < iter->feedback_kernel.HH)
         {
--- a/libgadget/blackhole.c
+++ b/libgadget/blackhole.c
@@ -925,11 +925,11 @@ blackhole_feedback_ngbiter(TreeWalkQueryBHFeedback * I,
     if(winds_is_particle_decoupled(other))
         return;
 
-    message(0, "SURVIVED SECOND RETURN\n");
+    message(0, "SURVIVED SECOND RETURN, other type: %i\n", P[other].Type);
 
     /****************************************************************************************/
     /* Compute fractional density for DF */
-     if(P[other].Type == 1 || P[other].Type == 4) 
+     if(P[other].Type == 1 || P[other].Type == 4 || ) 
      {
         endrun(2, "FOUND AT LEAST SOME DM/STAR\n");
 //        message(0, "FOUND AT LEAST SOME DM/STAR");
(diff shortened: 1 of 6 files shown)


libgadget/blackhole.c
```diff
@@ -515,12 +653,60 @@ blackhole_accretion_preprocess(int n, TreeWalk * tw)
...
+        {
+            BHP(n).DFAccel[j] = - 4. * M_PI * All.G * All.G * P[n].Mass * rho_frac * 
+            log_lam * (P[n].Vel[j] - BH_GET_PRIV(tw)->BH_SurroundingVel[PI][j]) / pow(bhvel, 3);
+            P[n].GravAccel[j]  += All.BH_DFBoost * BHP(n).DFAccel[j]; // Add a boost factor
+        }
+    }
+    else
+    {
+        for(j = 0; j < 3; j++) 
+        {
```
sbird, line 702: Better would be to just not allocate the array if DynamicalFriction is off and avoid this else clause. On use one would check for a NULL pointer in this array instead. An alternative is to change this to an unconditional memset(DFAccel, 0, 3) before the if clause.

libgadget/blackhole.c
```diff
@@ -611,6 +797,26 @@ blackhole_accretion_ngbiter(TreeWalkQueryBHAccretion * I,
         } while(!__atomic_compare_exchange_n(&(BHP(other).SwallowID), &readid, newswallowid, 0, __ATOMIC_RELAXED, __ATOMIC_RELAXED));
     }
 
+    /*************************************************************************/
+
+    /* Collect surrounding DM/Star (optional gas) density/velocity for DF computation */
+    if(P[other].Type == 1 || P[other].Type == 4 || (P[other].Type == 0 && All.BH_DynamicalFriction == 0) ){
```
sbird, line 803: At this point I realised that BH_DynamicalFriction == 0 means Dynamical Friction is ON and BH_DynamicalFriction == 3 means it is OFF! That's odd.

libgadget/blackhole.c
```diff
@@ -722,6 +935,40 @@ blackhole_feedback_ngbiter(TreeWalkQueryBHFeedback * I,
...
+            for(k = 0; k < 3; k++) 
+            {
+                bhvel += pow(I->Vel[k] - I->SurroundingVel[k], 2);
+                othervel += pow(P[other].Vel[k] - I->SurroundingVel[k], 2);
+            }
+            bhvel = sqrt(bhvel);
+            othervel = sqrt(othervel);
+
+        
+            if (othervel < bhvel)
```
sbird, line 962: Don't really need the sqrt here

libgadget/drift.c
```diff
@@ -47,7 +48,7 @@ static void real_drift_particle(int i, inttime_t ti1, const double ddrift, const
     /* Jumping of BH */
     if(P[i].Type == 5) {
         int k;
-        if (BHP(i).JumpToMinPot) {
+        if (BHP(i).JumpToMinPot && All.BH_DynamicalFriction == 3) {
```
sbird, line 51: Cleaner to just not set JumpToMinPot in blackhole.c if repositioning is off.

libgadget/init.c
```diff
@@ -96,6 +96,8 @@ set_init_params(ParameterSet * ps)
 
         All.BlackHoleOn = param_get_int(ps, "BlackHoleOn");
         All.WriteBlackHoleDetails = param_get_int(ps,"WriteBlackHoleDetails");
+        All.BH_DynamicalFriction = param_get_int(ps,"BH_DynamicalFriction");
+        All.BH_DFBoost = param_get_int(ps,"BH_DFBoost");
```
sbird, line 100: Could you add a separate extra parameter that enables and disables repositioning? Something like BH_Reposition? That way we can easily disable repositioning without turning on dynamical friction.

libgadget/blackhole.c
```diff
@@ -819,10 +1068,15 @@ blackhole_accretion_reduce(int place, TreeWalkResultBHAccretion * remote, enum T
...
             BHP(place).MinPotPos[k] = remote->BH_MinPotPos[k];
-            /* We set the velocity as well because
-             * we don't want drift to move us too far.*/
-            P[place].Vel[k] = remote->BH_MinPotVel[k];
+            /*********************************************************************/
+            if (All.BH_DynamicalFriction == 3){ /* reset vel only when reposition is on*/
+                /* We set the velocity as well because
+                 * we don't want drift to move us too far.*/
+                P[place].Vel[k] = remote->BH_MinPotVel[k];
+            }
```
sbird, line 1076: This is a good catch: I think I will fix this in the master branch

nianyic7:
I might have opened my PR too early and there are a few more changes I want to make to DF. I didn't know that after opening a PR everyone watching MP-Gadget will keep receiving notifications when I push... In order not to bother other people with my commits I will close this pull request for now and make another one after more testings/changes. Sorry about that!

sbird:
Hi Nianyi - please don't apologise, I prefer it if you open the PR early. That way everything is in the open and we know about conflicts early, plus we can pull any bugfixes directly into master. eg, there are now conflicts with https://github.com/MP-Gadget/MP-Gadget/pull/446 . If we didn't want to receive push emails we can turn them off. Anyway, I get messages every time someone pushes to the LSST repos. If you are not ready for review you can just put "WIP:" in the PR name and we will ignore it.

libgadget/blackhole.c
rainwoodman: In any case, BH_DynamicalFrictionMethod or Mode / Type is more specific than DynamiccalFriction alone.

PR #447, opened 2020-06-20
Compute the BH Entropy accounting for the effective equation of state

sbird:
Instead of cooling the gas explicitly.

### DIFF
--- a/libgadget/sfr_eff.c
+++ b/libgadget/sfr_eff.c
@@ -502,6 +502,22 @@ static int make_particle_star(int child, int parent, int placement)
     return retflag;
 }
 
+/* If gas is not star forming, return the standard entropy.
+ * If gas is star forming, get the cold-cloud mass-weighted entropy,
+ * an entropy which accounts for the cold cloud fraction.
+ * This is used for the BH accretion.*/
+MyFloat
+get_effective_entropy(int i, double dtime, const double a3inv)
+{
+    /* Normal entropy if not star-forming*/
+    if(!sfreff_on_eeqos(&SPHP(i), a3inv))
+        return SPHP(i).Entropy;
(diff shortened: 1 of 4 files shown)


libgadget/sfr_eff.c
```diff
@@ -502,6 +502,22 @@ static int make_particle_star(int child, int parent, int placement)
...
+MyFloat
+get_effective_entropy(int i, double dtime, const double a3inv)
+{
+    /* Normal entropy if not star-forming*/
+    if(!sfreff_on_eeqos(&SPHP(i), a3inv))
+        return SPHP(i).Entropy;
+    struct sfr_eeqos_data sfr_data = get_sfr_eeqos(&P[i], &SPHP(i), dtime, a3inv);
+    const double egyeff = sfr_params.EgySpecCold * sfr_data.cloudfrac + (1 - sfr_data.cloudfrac) * sfr_data.egyhot;
+    const double enttou = pow(SPH_EOMDensity(i) * a3inv, GAMMA_MINUS1) / GAMMA_MINUS1;
+    return egyeff/enttou;
```
rainwoodman, line 518: ent_to_u?

rainwoodman: Worth to check if the SFR model uses EOMDensity (the discontinous egywtdensity under pressure entropy) or always use the smooth Density.

There was no particular reason which to prefer, but the choice should be consistent here and there.

sbird: It always uses EOMDensity when doing the entropy to internal energy conversion (although it uses Density when working out the cooling rate). Does that seem right to you?

rainwoodman: Yes. That's correct.

PR #448, opened 2020-06-24
Bhtcool

sbird:
Bring back the use of the cooling time for the black hole heated star-forming gas, instead of the relaxation time


sbird:
Replaced by bhtcool2

PR #449, opened 2020-06-29
bhtcool-df

nianyic7:
Added DF and hydro drag into bhtcool branch. Addressed some comments in the previous PR (moved allvars params into bhparams etc., but there are still some comments to be addressed). Now working on making DF density kernel correct & other popular versions of DF.


sbird:
Thanks Nianyi!

sbird:
The drag force makes sense to me now

PR #450, opened 2020-07-05
Fixsfr

sbird:
Fixes to the printed star formation rate, which was inconsistent across all three of PART_, PIG_ and sfr.txt!


rainwoodman:
lol.  legacy == debt.

PR #451, opened 2020-07-05
Windfix

sbird:
The original wind feedback model was dependent on the ordering in which the particles were evaluated: once a particle had DelayTime set it was immune to feedback from any other nearby stars. Because this makes the final result uncomfortably sensitive to the threading, I changed this so that extra star formation just kicks particles a little more. Unfortunately this a) changes the model and b) introduces a potential bug where partially evaluated star particles (when the export buffer fills up) may kick gas twice. This is extremely rare for the wind treewalk, but with the big run it is a concern.

This PR fixes the problem more comprehensively by first treewalking to find the closest star to the gas particle, then using that to kick.

### DIFF
--- a/libgadget/winds.c
+++ b/libgadget/winds.c
@@ -117,6 +117,9 @@ winds_decoupled_hydro(int i, double atime)
     SPHP(i).MaxSignalVel = hsml_c * DMAX((2 * windspeed), SPHP(i).MaxSignalVel);
 }
 
+static int
+get_wind_dir(int i, double dir[3]);
+
 static void
 sfr_wind_reduce_weight(int place, TreeWalkResultWind * remote, enum TreeWalkReduceMode mode, TreeWalk * tw);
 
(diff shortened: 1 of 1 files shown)


libgadget/winds.c
```diff
@@ -255,13 +261,48 @@ winds_and_feedback(int * NewStars, int NumNewStars, const double Time, const dou
     ta_free(priv->NPRedo);
     ta_free(priv->NPLeft);
 
+    /* Some particles may be kicked by multiple stars on the same timestep.
+     * To ensure this happens only once and does not depend on the order in
+     * which the loops are executed, particles are kicked by the nearest new star.*/
+    priv->StarKickVelocity = (double * ) mymalloc("NearestStar", SlotsManager->info[0].size * sizeof(double));
```
rainwoodman, line 267: What if two stars are of the same distance away from the particle?

Perhaps there is a more fundamental way to break the tie than tagging by the distance?

sbird: I could break the tie with the particle ID?

My thinking with the distance was that this was modelling a shockwave from a supernova moving on sub-timestep scales, and the shockwave from the nearest star should arrive first.

sbird:
Hopefully this approximately restores the behaviour as before c240a23

Issue #453, opened 2020-07-23
How to interpret run.sh error


yamatteo:
Hello everybody, thank you for developing this version of Gadget.

I managed to compile and run dm-only example but, during execution, run.sh prompt the following errors. How can I interpret then to understand what went wrong?

```
[ 000127.67 ] Begin Step 14, Time: 0.02 (100000), Redshift: 49, Nf = 00000002097152, Systemstep: 0.000847934, Dloga: 0.0433217, status: PM-Step
[ 000127.67 ] TotNumPart: 0000002097152 SPH 0000000000000 BH 0000000000 STAR 0000000000000 
[ 000127.67 ] Occupied:            0            1            2            3            4            5 dt
[ 000127.67 ]  X bin=17            0      2081920            0            0            0            0 0.20118
[ 000127.67 ]  X bin=16            0        15232            0            0            0            0 0.10059
[ 000127.67 ]                -----------------------------------
[ 000127.67 ] Total:               0      2097152            0            0            0            0  Sum:       2097152
[ 000127.67 ] Tree construction.  (presently allocated=114.219 MB)
[ 000127.67 ] Allocated 50.4619 MByte for 365936 tree nodes. firstnode 786432. (presently allocated 164.691 MB)
[ 000127.71 ] Tree constructed (moments: 1). First node 786432, number of nodes 107401, first pseudo 1152368. NTopLeaves 64
[ 000127.71 ] Begin force computation.
[ 000127.71 ] Begin tree force.  (presently allocated=131.16 MB)
[ 000135.04 ] Short-range gravitational tree force is done.
[ 000135.05 ] max number of regions is 16
[ 000135.27 ] PetaPM:  0000415754/0000627264 Pencils and 0028838986 Cells
[ 000143.50 ] Task 0: Writing Power Spectrum to output/powerspectrum-0.0200.txt
[ 000143.50 ] Forces computed.
[ 000143.58 ] saving snapshot into output/PART_000
[ 000143.63 ] Will write 2097152 particles to 4 Files for 1/Mass
[ 000143.63 ] Done writing 2097152 particles to 4 Files
[ 000143.66 ] Will write 2097152 particles to 4 Files for 1/Position
[ 000143.70 ] Done writing 2097152 particles to 4 Files
[ 000143.71 ] Will write 2097152 particles to 4 Files for 1/Velocity
[ 000143.73 ] Done writing 2097152 particles to 4 Files
[...]

PR #455, opened 2020-07-23
Print ERROR: if message is from endrun.

sbird:
Fixes: #453


sbird:
@rainwoodman What do you think?

PR #457, opened 2020-08-15
Bhswallowlist

sbird:
This moves the actual evaluation of the mergers outside of the blackhole feedback treewalk into specialised code. Te advantage of doing it this way is that we can lose the special threading.


sbird:
Too many conflicts, don't want to fix

PR #459, opened 2020-08-18
df and BH merge

yueyingn:
add DF to BH and apply GravBound criteria to BH merger


libgadget/blackhole.c
```diff
@@ -599,40 +945,73 @@ blackhole_accretion_ngbiter(TreeWalkQueryBHAccretion * I,
...
-        } while(!__atomic_compare_exchange_n(&(BHP(other).SwallowID), &readid, newswallowid, 0, __ATOMIC_RELAXED, __ATOMIC_RELAXED));
+        O->encounter = 1; // mark the event when two BHs encounter each other
+        
+        int d;
+        double KE = 0;
+        double PE = 0;
+        
+        for(d = 0; d < 3; d++){
+            KE += 0.5 * pow(I->Vel[d] - P[other].Vel[d], 2);
+            PE += (I->Accel[d] - P[other].GravAccel[d]) * (I->Pos[d] - P[other].Pos[d]);
```
sbird, line 979: Should this include GravPM?

sbird: Also periodic_wrap()

libgadget/blackhole.c
```diff
@@ -766,8 +1149,16 @@ blackhole_feedback_ngbiter(TreeWalkQueryBHFeedback * I,
         O->BH_CountProgs += BHP(other).CountProgs;
         /* Leave the swallowed BH mass around
          * so we can work out mass at merger. */
-        O->Mass += (P[other].Mass);
         O->BH_Mass += (BHP(other).Mass);
+
```
sbird, line 1153: Please pull this chunk to a new branch

libgadget/blackhole.c
```diff
@@ -945,8 +1354,13 @@ void blackhole_make_one(int index) {
     int j;
     for(j = 0; j < 3; j++) {
         BHP(child).MinPotPos[j] = P[child].Pos[j];
+        BHP(child).HaloMinPotPos[j] = P[child].Pos[j];
     }
     BHP(child).JumpToMinPot = 0;
+    
+    if (blackhole_params.SeedBHDynMass>0){
```
sbird, line 1361: This to the same new branch as 1156

libgadget/blackhole.c
```diff
@@ -599,40 +945,73 @@ blackhole_accretion_ngbiter(TreeWalkQueryBHAccretion * I,
             }
         }
     }
+    
+    
+    /********************************************/
+    /* Find min pot pos within halo */
+    if(r < 10)
+    {
```
sbird, line 953: Please remove this

libgadget/blackhole.c
```diff
@@ -514,17 +830,42 @@ blackhole_accretion_postprocess(int i, TreeWalk * tw)
...
+    /*************************************************************************/
+    
+    if(blackhole_params.BH_DRAG > 0){
+        double fac = 0;
+        if (blackhole_params.BH_DRAG == 1) fac = BHP(i).Mdot/P[i].Mass; 
+        if (blackhole_params.BH_DRAG == 2) fac = blackhole_params.BlackHoleEddingtonFactor * meddington/BHP(i).Mass;
+        for(k = 0; k < 3; k++) {
+            fac *= All.cf.a; /* dv = acc * kick_fac = acc * a^{-1}dt, therefore acc = a*dv/dt  */
+            BHP(i).DragAccel[k] = -(P[i].Vel[k] - BH_GET_PRIV(tw)->BH_SurroundingGasVel[PI][k])*fac;
+            P[i].GravAccel[k]  += BHP(i).DragAccel[k];
```
sbird, line 843: Instead of adding this to GravAccel here, why not add DragAccel and DFAccel inside the do_the_short_range_kick() in timestep.c like we do for HydroAccel. So inside do_the_short_range_kick you would have:
if(Type == 5) {
Vel += BHP(i).DragAccel[k];
}
(with the appropriate factors)

libgadget/blackhole.c
```diff
@@ -723,12 +1102,15 @@ blackhole_feedback_ngbiter(TreeWalkQueryBHFeedback * I,
         double hsearch;
         hsearch = decide_hsearch(I->Hsml);
 
-        iter->base.mask = 1 + 32;
+        /**************************************************************/
+        /* Need to visit all types for DF computation so changed mask */
+
+        iter->base.mask = 1 + 2 + 4 + 8 + 16 + 32;
```
sbird, line 1108: Remove this

libgadget/blackhole.c
```diff
@@ -471,6 +643,149 @@ blackhole(const ActiveParticles * act, ForceTree * tree, FILE * FdBlackHoles, FI
...
+            f_of_x = 0;
+
+        lambda = 1. + blackhole_params.BH_DFbmax * pow((bhvel/All.cf.a),2) / All.G / P[n].Mass;
+
+        for(int j = 0; j < 3; j++) 
+        {
+            BHP(n).DFAccel[j] = - 4. * M_PI * All.G * All.G * P[n].Mass * BH_GET_PRIV(tw)->BH_SurroundingDensity[PI] * 
+            log(lambda) * f_of_x * (P[n].Vel[j] - BH_GET_PRIV(tw)->BH_SurroundingVel[PI][j]) / pow(bhvel, 3);
+            BHP(n).DFAccel[j] *= All.cf.a;  // convert to code unit of acceleration
+            P[n].GravAccel[j]  += blackhole_params.BH_DFBoostFactor * BHP(n).DFAccel[j]; // Add a boost factor
```
sbird, line 708: Do not mutate

libgadget/blackhole.c
```diff
@@ -32,6 +32,17 @@ struct BlackholeParams
     double SeedBlackHoleMass;	/*!< Seed black hole mass */
     double BlackHoleEddingtonFactor;	/*! Factor above Eddington */
     int BlackHoleRepositionEnabled; /* If true, enable repositioning the BH to the potential minimum*/
+    
+    /**********************************************************************/
+    int MergeGravBound; /*if 1, apply gravitational bound criteria for BH mergers */
+    
+    int BH_DynFrictionMethod;/*0 for off; 1 for Star Only; 2 for DM+Star; 3 for DM+Star+Gas */
+    int BH_DFBoostFactor; /*Optional boost factor for DF*/
+    double BH_DFbmax; /* keep it as a parameter for tuning purpose */
```
sbird, line 41: What are the units for this one? Is it kpc/h?

gadget/params.c
```diff
@@ -208,6 +208,13 @@ create_gadget_parameter_set()
...
     param_declare_int(ps,"WriteBlackHoleDetails",OPTIONAL, 0, "If set, output BH details at every time step.");
 
+    param_declare_int(ps,"BH_DynFrictionMethod",OPTIONAL, 0, "If set to non-zero, dynamical friction is applied through this method.");
+    param_declare_int(ps,"BH_DFBoostFactor",OPTIONAL, 1, "If set, dynamical friction is boosted by this factor.");
+    param_declare_int(ps,"BH_DFbmax",OPTIONAL, 20, "Maximum impact parameter for dynamical friction.");
+    param_declare_int(ps,"BH_DRAG",OPTIONAL, 0, "Add drag force to the BH dynamic");
+    param_declare_int(ps,"MergeGravBound",OPTIONAL, 1, "If set to 1, apply gravitational bound criteria for merging event. This criteria would be automatically turned off if reposition is enabled."); 
+    param_declare_double(ps, "SeedBHDynMass", OPTIONAL, -1, "The initial dynamic mass of BH, default -1 will use the mass of gas particle. Larger Mdyn would help BH to be more stable in the early phase if turn off reposition.");
```
sbird, line 216: Split this guy out to a separate PR

libgadget/blackhole.c
```diff
@@ -599,40 +945,73 @@ blackhole_accretion_ngbiter(TreeWalkQueryBHAccretion * I,
...
-            } else if(readid == (MyIDType) -1 && P[other].ID < I->ID) {
-                /* Unmarked, the BH with bigger ID swallows */
-                newswallowid = I->ID;
-            }
-            else
-                break;
-        /* Swap in the new id only if the old one hasn't changed:
-         * in principle an extension, but supported on at least clang >= 9, gcc >= 5 and icc >= 18.*/
-        } while(!__atomic_compare_exchange_n(&(BHP(other).SwallowID), &readid, newswallowid, 0, __ATOMIC_RELAXED, __ATOMIC_RELAXED));
+        O->encounter = 1; // mark the event when two BHs encounter each other
```
sbird, line 971: I wasn't so clear what this was for? It marks events where they are close, and is useful because sometimes they are close but do not merge? If so would it be useful to store it as a counter in the BHP struct?

libgadget/blackhole.c
```diff
@@ -599,40 +945,73 @@ blackhole_accretion_ngbiter(TreeWalkQueryBHAccretion * I,
...
+        for(d = 0; d < 3; d++){
+            KE += 0.5 * pow(I->Vel[d] - P[other].Vel[d], 2);
+            PE += (I->Accel[d] - P[other].GravAccel[d]) * (I->Pos[d] - P[other].Pos[d]);
+        }
+        
+        KE /= (All.cf.a*All.cf.a); /* convert to proper velocity */
+        PE /= All.cf.a; /* convert to proper unit */
+        
+        /* add option to merge the BHs if they are gravitationally bounded */
+        if(blackhole_params.BlackHoleRepositionEnabled == 1 || blackhole_params.MergeGravBound == 0 || (PE + KE < 0 && blackhole_params.MergeGravBound == 1))
```
sbird, line 986: The "blackhole_params.MergeGravBound == 0" is redundant

libgadget/blackhole.c
```diff
@@ -853,13 +1247,24 @@ blackhole_accretion_reduce(int place, TreeWalkResultBHAccretion * remote, enum T
             BHP(place).MinPotVel[k] = remote->BH_MinPotVel[k];
         }
     }
+    
+    /****************************************************************************/
+    if(HaloMinPot[PI] > remote->BH_HaloMinPot)
```
sbird, line 1252: I think this is MinPotPos but over a wider area? Please make this some kind of debug code.

libgadget/sfr_eff.c
```diff
@@ -409,6 +411,17 @@ sfreff_on_eeqos(const struct sph_particle_data * sph, const double a3inv)
     if(sph->Density * a3inv >= sfr_params.PhysDensThresh)
         flag = 1;
 
+    if(sfr_params.BHFeedbackUseTcool == 2) {
```
sbird, line 414: Drop this hunk.

libgadget/blackhole.c
yueyingn: Yes, this is in kpc/h.

sbird: Please add a comment that it is in internal units

libgadget/blackhole.c
yueyingn: Yes, it gives me a quick estimate of how long does the Gravitational Bound criteria delay the merging event: (time difference between the first encounter>0 and acBHMass>0.)  It is already in BHP struct. By 'counter' you mean increment every time step?

libgadget/blackhole.c
yueyingn: I would like to keep the option to turn off the Gravitational Bound criteria.

libgadget/blackhole.c
sbird: Yes, I mean increment every time it happens, so we can track objects that have multiple close encounters on the way to a merger.

libgadget/blackhole.c
sbird: I see, I misread the conditional. Could you reword it to make it a bit shorter?
Maybe pull this code out to a short function that computes the PE and KE and checks if they are gravitationally bound?

libgadget/blackhole.c
yueyingn: I am not sure how to do this. I think we need to distinguish between different P[other].ID to account the multiple encounters? Otherwise we are just increment 1 every time step.

sbird: Hmm, yes, that is a good point. That would be quite difficult to do. Just keep it as it is then

libgadget/blackhole.c
```diff
@@ -49,12 +61,14 @@ typedef struct {
...
     MyFloat BH_MinPot;
-
+    
     int BH_minTimeBin;
+    int encounter;
     MyFloat FeedbackWeightSum;
 
     MyFloat SmoothedEntropy;
     MyFloat GasVel[3];
+
```
sbird, line 71: Could you possibly fix up the whitespace changes?

libgadget/blackhole.c
```diff
@@ -93,6 +136,17 @@ struct BHPriv {
     MyFloat * MinPot;
     MyFloat * BH_Entropy;
     MyFloat (*BH_SurroundingGasVel)[3];
+    
+    /*************************************************************************/
+    /* Temporaries computed in the accretion treewalk and used
+     * in the feedback treewalk*/
```
sbird, line 142: Caption needs updating: you mean in the dynamic friction treewalk.

libgadget/blackhole.c
```diff
@@ -154,6 +225,13 @@ void set_blackhole_params(ParameterSet * ps)
 
         blackhole_params.BlackHoleFeedbackMethod = param_get_enum(ps, "BlackHoleFeedbackMethod");
         blackhole_params.BlackHoleRepositionEnabled = param_get_int(ps, "BlackHoleRepositionEnabled");
+        /***********************************************************************************/
+        blackhole_params.BH_DynFrictionMethod = param_get_int(ps, "BH_DynFrictionMethod");
+        blackhole_params.BH_DFBoostFactor = param_get_int(ps, "BH_DFBoostFactor");
+        blackhole_params.BH_DFbmax = param_get_int(ps, "BH_DFbmax");
```
sbird, line 231: This should be param_get_double()

sbird: I think it doesn't matter right now because the default variable is actually an integer, which will be cast to a double, but it could be confusing if someone tries to set it to 20.5 and nothing happens!

libgadget/blackhole.c
```diff
@@ -204,6 +308,9 @@ blackhole_feedback_ngbiter(TreeWalkQueryBHFeedback * I,
 static double
 decide_hsearch(double h);
 
+static double 
+periodic_wrap(double x, double BoxSize);
+
```
sbird, line 313: Sorry, the right macro to use here is actually NEAREST

libgadget/init.c
```diff
@@ -96,6 +96,7 @@ set_init_params(ParameterSet * ps)
 
         All.BlackHoleOn = param_get_int(ps, "BlackHoleOn");
         All.WriteBlackHoleDetails = param_get_int(ps,"WriteBlackHoleDetails");
+
```
sbird, line 99: Please remove these whitespace changes too, and in sfr_eff.c

libgadget/blackhole.c
```diff
@@ -49,8 +61,9 @@ typedef struct {
     MyFloat BH_MinPotPos[3];
     MyFloat BH_MinPotVel[3];
     MyFloat BH_MinPot;
-
+
```
sbird, line 64: whitespace

libgadget/blackhole.c
```diff
@@ -63,13 +76,40 @@ typedef struct {
...
+
 typedef struct {
     TreeWalkQueryBase base;
     MyFloat Hsml;
     MyFloat BH_Mass;
     MyIDType ID;
     MyFloat FeedbackEnergy;
     MyFloat FeedbackWeightSum;
+
```
sbird, line 111: whitespace

libgadget/blackhole.c
```diff
@@ -78,11 +118,13 @@ typedef struct {
     MyFloat AccretedMomentum[3];
     MyFloat BH_Mass;
     int BH_CountProgs;
+
```
sbird, line 121: whitespace

libgadget/blackhole.c
```diff
@@ -78,11 +118,13 @@ typedef struct {
...
 } TreeWalkResultBHFeedback;
 
 typedef struct {
     TreeWalkNgbIterBase base;
     DensityKernel feedback_kernel;
+
```
sbird, line 127: whitespace

libgadget/blackhole.c
```diff
@@ -471,6 +634,148 @@ blackhole(const ActiveParticles * act, ForceTree * tree, FILE * FdBlackHoles, FI
...
+        double bhvel;
+        double lambda, x, f_of_x;
+        const double a_erf = 8 * (M_PI - 3) / (3 * M_PI * (4. - M_PI));
+
+
+        /* normalize velocity/dispersion */
+        BH_GET_PRIV(tw)->BH_SurroundingRmsVel[PI] /= BH_GET_PRIV(tw)->BH_SurroundingDensity[PI];
+        BH_GET_PRIV(tw)->BH_SurroundingRmsVel[PI] = sqrt(BH_GET_PRIV(tw)->BH_SurroundingRmsVel[PI]);
+        for(int k = 0; k < 3; k++)
```
sbird, line 669: We avoid declaring variables in the loop body because (I think) one of the compilers on BlueTides doesn't support it properly. Could you move the variable declaration of of the loop? Just use j for all the loops in this clause.

libgadget/blackhole.c
```diff
@@ -471,6 +634,148 @@ blackhole(const ActiveParticles * act, ForceTree * tree, FILE * FdBlackHoles, FI
...
+            BHP(n).DFAccel[j] *= All.cf.a;  // convert to code unit of acceleration
+            BHP(n).DFAccel[j] *= blackhole_params.BH_DFBoostFactor; // Add a boost factor
+        }
+        message(0,"x=%e, log(lambda)=%e, fof_x=%e, Mbh=%e, ratio=%e \n",
+           x,log(lambda),f_of_x,P[n].Mass,BHP(n).DFAccel[0]/P[n].GravAccel[0]);
+    }
+    else
+    {   
+        message(0, "Density is zero in DF kernel, kernel may be too small.\n");
+        for(int j = 0; j < 3; j++)
```
sbird, line 706: Maybe easier to memset(0) when it is malloced and remove this.

libgadget/blackhole.c
```diff
@@ -471,6 +634,148 @@ blackhole(const ActiveParticles * act, ForceTree * tree, FILE * FdBlackHoles, FI
...
+            BHP(n).DFAccel[j] = 0;
+        }
+    }
+}
+    /*******************************************************************/
+
+static int
+blackhole_dynfric_haswork(int n, TreeWalk * tw){
+    /*Black hole not being swallowed*/
+    return (P[n].Type == 5) && (!P[n].Swallowed) && (blackhole_params.BH_DynFrictionMethod > 0);
```
sbird, line 717: Just don't run the treewalk if BH_DynFrictionMethod == 0 : guard treewalk_run()

libgadget/blackhole.c
```diff
@@ -543,6 +868,7 @@ blackhole_feedback_postprocess(int n, TreeWalk * tw)
         P[n].Mass += accmass;
         BHP(n).Mass += BH_GET_PRIV(tw)->BH_accreted_BHMass[PI];
     }
+
```
sbird, line 871: whitespace

libgadget/blackhole.c
```diff
@@ -602,37 +930,68 @@ blackhole_accretion_ngbiter(TreeWalkQueryBHAccretion * I,
...
+            /* check if the two BHs are gravitationally bounded */
+            int d;
+            double KE = 0;
+            double PE = 0;
+
+            for(d = 0; d < 3; d++){
+                KE += 0.5 * pow(I->Vel[d] - P[other].Vel[d], 2);
+                double dx = NEAREST(I->Pos[d] - P[other].Pos[d], All.BoxSize);
+                double da = (I->Accel[d] - P[other].GravAccel[d] - P[other].GravPM[d] - BHP(other).DFAccel[d]);
+                PE += da * dx;
```
sbird, line 955: Please make this a separate function

libgadget/blackhole.c
```diff
@@ -766,8 +1125,9 @@ blackhole_feedback_ngbiter(TreeWalkQueryBHFeedback * I,
         O->BH_CountProgs += BHP(other).CountProgs;
         /* Leave the swallowed BH mass around
          * so we can work out mass at merger. */
-        O->Mass += (P[other].Mass);
         O->BH_Mass += (BHP(other).Mass);
+        O->Mass += (P[other].Mass);
```
sbird, line 1129: Can you drop this hunk? Also, whitespace.

libgadget/slotsmanager.h
```diff
@@ -37,16 +37,22 @@ struct bh_particle_data {
     /* Merger time of the black hole.
      * After this, all values are fixed. */
     MyFloat SwallowTime;
-
+
```
sbird, line 40: whitespace

libgadget/timestep.c
```diff
@@ -371,6 +371,11 @@ do_the_short_range_kick(int i, inttime_t tistart, inttime_t tiend)
     {
         P[i].Vel[j] += P[i].GravAccel[j] * Fgravkick;
     }
+
```
sbird, line 374: Please add a comment

libgadget/blackhole.c
```diff
@@ -602,37 +930,68 @@ blackhole_accretion_ngbiter(TreeWalkQueryBHAccretion * I,
...
+        if(blackhole_params.MergeGravBound == 1){ 
+            /* check if the two BHs are gravitationally bounded */
+            int d;
+            double KE = 0;
+            double PE = 0;
+
+            for(d = 0; d < 3; d++){
+                KE += 0.5 * pow(I->Vel[d] - P[other].Vel[d], 2);
+                double dx = NEAREST(I->Pos[d] - P[other].Pos[d], All.BoxSize);
+                double da = (I->Accel[d] - P[other].GravAccel[d] - P[other].GravPM[d] - BHP(other).DFAccel[d]);
```
sbird, line 954: Please add a comment here

libgadget/blackhole.c
```diff
@@ -602,37 +930,68 @@ blackhole_accretion_ngbiter(TreeWalkQueryBHAccretion * I,
 
     /* Accretion / merger doesn't do self interaction */
     if(P[other].ID == I->ID) return;
-
-    if(P[other].Type == 5 && r2 < iter->accretion_kernel.HH)	/* we have a black hole merger */
+    
+    /* we have a black hole merger, use 2 times GravitationalSoftening as merging criteria*/
```
sbird, line 934: Add a comment here explaining that the previous implementation uses the softening length.

libgadget/blackhole.c
```diff
@@ -514,6 +820,24 @@ blackhole_accretion_postprocess(int i, TreeWalk * tw)
     double dtime = get_dloga_for_bin(P[i].TimeBin) / All.cf.hubble;
 
     BHP(i).Mass += BHP(i).Mdot * dtime;
+
+    /*************************************************************************/
+    
+    if(blackhole_params.BH_DRAG > 0){
+        double fac = 0;
```
sbird, line 827: Please add comments explaining where the physics for this comes from, like a link to an equation in a paper somewhere.

libgadget/blackhole.c
```diff
@@ -471,6 +634,148 @@ blackhole(const ActiveParticles * act, ForceTree * tree, FILE * FdBlackHoles, FI
...
+    /* acc_friction = -4*pi*G^2 * Mbh * log(lambda) * rho * f_of_x * bhvel / |bhvel^3| */
+    /*       f_of_x = [erf(x) - 2*x*exp(-x^2)/sqrt(pi)]                                */
+    /*       lambda = b_max * v^2 / G / (M+m)                                          */
+    /*        b_max = Size of system (e.g. Rvir)                                       */
+    /*            v = Relative velocity of BH with respect to the environment          */
+    /*            M = Mass of BH                                                       */
+    /*            m = individual mass elements composing the large system (e.g. m<<M)  */
+    /*            x = v/sqrt(2)/sigma                                                  */
+    /*        sigma = width of the max. distr. of the host system                      */
+    /*                (e.g. sigma = v_disp / 3                                         */
```
sbird, line 657: Could you add a link to Tremmel's paper?

gadget/params.c
```diff
@@ -208,6 +208,12 @@ create_gadget_parameter_set()
     param_declare_double(ps, "BlackHoleFeedbackRadiusMaxPhys", OPTIONAL, 0, "If set, the physical radius at which the black hole feedback energy is deposited. When both this flag and BlackHoleFeedbackRadius are both set, the smaller radius is used.");
     param_declare_int(ps,"WriteBlackHoleDetails",OPTIONAL, 0, "If set, output BH details at every time step.");
 
+    param_declare_int(ps,"BH_DynFrictionMethod",OPTIONAL, 0, "If set to non-zero, dynamical friction is applied through this method.");
```
sbird, line 211: Could you add some more documentation here and in blackhole.c about what each BH_DynFrictionMethod value means? It distinguishes between DynFrictionMethod = 1 and  DynFrictionMethod = 3 in a few places.

libgadget/blackhole.c
```diff
@@ -336,6 +472,22 @@ blackhole(const ActiveParticles * act, ForceTree * tree, FILE * FdBlackHoles, FI
...
+
+    MPIU_Barrier(MPI_COMM_WORLD);
+    message(0, "Beginning dynamical friction calculation. \n");
+    /* Environment variables for DF */
+    priv->BH_SurroundingRmsVel = mymalloc("BH_SurroundingRmsVel", SlotsManager->info[5].size * sizeof(priv->BH_SurroundingRmsVel));
+    priv->BH_SurroundingVel = (MyFloat (*) [3]) mymalloc("BH_SurroundingVel", 3* SlotsManager->info[5].size * sizeof(priv->BH_SurroundingVel[0]));
+    priv->BH_SurroundingParticles = mymalloc("BH_SurroundingParticles", SlotsManager->info[5].size * sizeof(priv->BH_SurroundingParticles));
+    priv->BH_SurroundingDensity = mymalloc("BH_SurroundingDensity", SlotsManager->info[5].size * sizeof(priv->BH_SurroundingDensity));
+
+    treewalk_run(tw_dynfric, act->ActiveParticle, act->NumActiveParticle);
```
sbird, line 488: Guard this by if(DynFrictionMethod > 0)

yueyingn: I tried to guard this by
if(DynFrictionMethod > 0)
         treewalk_run(tw_dynfric, act->ActiveParticle, act->NumActiveParticle);

but that would make the run with DynFrictionMethod = 0 stuck after the first time step, at 
[ 000109.26 ] Attempting a domain exchange

libgadget/blackhole.c
yueyingn: Sorry, actually in Gizmo's implementation, this DFbmax is already in physical unit of kpc. (Already corrected it.)  However the value of this factor doesn't have large effect.

libgadget/blackhole.c
sbird: hmm, could you figure out why that is and fix it? My guess is that it is because one of the *Accel variables is not initialised in this case, so the particle receives an arbitrary kick and moves out of the box.

libgadget/blackhole.c
```diff
@@ -872,6 +1247,8 @@ blackhole_accretion_copy(int place, TreeWalkQueryBHAccretion * I, TreeWalk * tw)
     for(k = 0; k < 3; k++)
     {
         I->Vel[k] = P[place].Vel[k];
+        I->Accel[k] = P[place].GravAccel[k] + P[place].GravPM[k] + BHP(place).DFAccel[k];
```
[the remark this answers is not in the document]
sbird: I do apologise, I was getting confused by the diff. I think this is fine then. I was thinking about whether DragAccel should be included, but it is small and besides I don't think it is really important during mergers.

libgadget/blackhole.c
```diff
@@ -231,6 +323,28 @@ add_injected_BH_energy(double unew, double injected_BH_energy, double mass)
...
+    double PE = 0;
+
+    for(j = 0; j < 3; j++){
+        KE += 0.5 * pow(dv[j], 2);
+        PE += da[j] * dx[j];
+    }
+    
+    KE /= (All.cf.a*All.cf.a); /* convert to proper velocity */
+    PE /= All.cf.a; /* convert to proper unit */
+
```
yueyingn, line 342: Shall we briefly review the h and a_scale factor in dx, dv, and da here?

From my understanding:
dx = v * drift_fac = v * a^{-2} dt             // drift_integ = da/(H(a)*a^3) = dt/a^2
dv = acc * kick_fac = acc * a^{-1} dt     // kick_integ = da/(H(a)*a^2) = dt/a

for a_scale factor,
=> dx ~ a (by '~', I mean dx * a = dx_proper), dv ~ a^{-1},  acc ~ a^{-2}
Therefore  KE /= a^2, PE /= a  converts to proper unit.

for h factor,
=> dx ~ h^{-1},  dt ~ h^{-1},  dv ~ h^{0}, d_acc ~ h^{1}
Therefore, da * dx has h factor canceled out, there's no h factor in both KE and PE.

PR #460, opened 2020-08-20
add SeedBHDynMass parameter to control the initial dynamic mass of BH.

yueyingn:


libgadget/blackhole.c
```diff
@@ -766,8 +770,17 @@ blackhole_feedback_ngbiter(TreeWalkQueryBHFeedback * I,
         O->BH_CountProgs += BHP(other).CountProgs;
         /* Leave the swallowed BH mass around
          * so we can work out mass at merger. */
-        O->Mass += (P[other].Mass);
+        
         O->BH_Mass += (BHP(other).Mass);
+        
+        /* Prevent large DynMass accumulates during merger */
```
sbird, line 776: This seems odd to me?

rainwoodman: If (I->BH_Mass > SeedBHDynMass && BHP(other).Mass > SeedBHDynMass ) {
   /* both have grown beyond Seed, directly add the masses*/
   I->BH_Mass  += BHP(other)->Mass;
   I->Mass += P[other]->mass
}
If (I->BH_Mass > SeedBHDynMass && BHP(other).Mass < SeedBHDynMass ) {
   I->BH_Mass  += BHP(other)->Mass
}
If (I->BH_Mass < SeedBHDynMass && BHP(other).Mass > SeedBHDynMass ) {
   I->BH_Mass  += BHP(other)->Mass;
   I->Mass = P[other]->mass
}
If (I->BH_Mass < SeedBHDynMass && BHP(other).Mass < SeedBHDynMass ) {
   I->BH_Mass  += BHP(other)->Mass;
   I->Mass = SeedBHDynMass
}

sbird: I think that the SeedBHDynMass should be almost totally decoupled from the BH_Mass. I also think that merging a seed mass black hole should do *something* just not as much as a full BHDynMass. So I propose:

O->BH_Mass += BHP(other)->Mass;
/* This code ensures that the extra dynamic mass created during seeding does not persist after mergers*/
If (I->Mass > SeedBHDynMass && P[other].Mass > SeedBHDynMass ) {
/* both have grown beyond Seed, directly add the masses*/
O->Mass += P[other]->mass
}
If (I->Mass <= SeedBHDynMass && P[other].Mass >= SeedBHDynMass ) {
O->Mass = P[other]->mass += (one gas particle mass);
}
If (I->Mass <= SeedBHDynMass && P[other].Mass <= SeedBHDynMass ) {
O->Mass = SeedBHDynMass 
/* Except this is not right: we have 'lost' the mass of a gas particle if a BH was seeded just before accretion*/
}

yueyingn: Note that O->Mass is the 'BH_accreted_Mass' or accmass instead of the final dynamic mass (see feedback_reduce, feedback_postprocess)

You actually mean 
If (I->Mass <= SeedBHDynMass && P[other].Mass <= SeedBHDynMass ) {O->Mass += 0}
Is that right?

But that will cause, e.g.,  {Mdyn: 1e7, Mint:9e6} + {Mdyn:1e7, Mint:9e6} = {Mdyn:1e7, Mint:1.8e7} and it will quickly swallow 4 gas particles after merger.  Not sure whether it will cause issue.

yueyingn: My previous implementation enforces {Mdyn: 1e7, Mint:9e6} + {Mdyn:1e7, Mint:9e6} = {Mdyn:1.8e7, Mint:1.8e7}, but that will fail to swallow some gas particle for the good of mass conservation. 
We might need to modify the swallow of gas carefully anyway if we want to improve mass conservation by swallowing gas.

libgadget/blackhole.c
```diff
@@ -766,8 +770,17 @@ blackhole_feedback_ngbiter(TreeWalkQueryBHFeedback * I,
...
+        if (I->BH_Mass + BHP(other).Mass < blackhole_params.SeedBHDynMass) {
+            O->Mass += 0; 
+        }
+        else {
+            O->Mass += I->BH_Mass + BHP(other).Mass - I->Mass;
```
sbird, line 781: I think I am just unhappy with breaking mass conservation under any circumstances. It seems like a bad idea that will have Consequences. I think that what we really want to do is to increase the dynamical friction force for newly-seeded BHs, right? So why don't we just do that?

yueyingn: Just BH born with higher dynamic mass is much more stable in the minimum potential than just boosting the dynamic friction.

sbird: So can we modify the seeding so it uses two particles?

yueyingn: Can we convert, say, 6 gas particles into BH when seeding a new blackhole? that will make mass conserves then.

yueyingn: Not sure how to do this.  Implementing it in swallowing of gas particle sounds easier to me.

sbird: That's not a bad idea: implement faster accretion just after seeding.

sbird: We could get fancy and just implement a different seeding model. Maybe an initial mode that seeds from and accretes stars to model the runaway GC mergers that are supposed to start these things.

sbird:
You know, thinking about this - this is a lot of hassle to get a seed which is basically the dark matter particle mass (because anything lighter is pushed around by the DM). Why don't we just seed the BH from a DM particle?

yueyingn:
Well, I feel it sounds weird that BH born from dark matter and swallow baryon afterward. Beyond we still need to worry about how to justify the addition of dynamic mass when two BHs merge especially when dark matter take part in.
I think PR#461 properly does the mass conservation.  I am testing it right now.

sbird:
For merging two BHs in the DM seed model we would just add the dynamic masses, no? It does sound a little weird but is not really so: we should really be seeding the BH from star particles anyway.

yueyingn:
Star particle is made of gas/baryon. I thought mass conservation is done within the baryon.. I never heard that cold dark matter can form a black hole.

yueyingn:
If we just add dynamic mass, then we are again quickly building up the dynamic mass if seed BHs merge.

sbird:
I thought the problem with that was just that the dynamic mass is not conserved in this PR, so that having them merge made it very much not conserved? But if dynamic mass is conserved explicitly, what is the problem with increasing it during a merger?

sbird:
> Star particle is made of gas/baryon. I thought mass conservation is done within the baryon.. I never heard that cold dark matter can form a black hole.

The traditional idea for forming these things is runaway mergers in globular clusters due to mass segregation. So my point is that the gas particles are not the thing that really seeds the BH, although they do form the accretion disc. 

The other idea for these seeds is a primordial black hole.

yueyingn:
Just stealing mass from dark matter when a black born will make omega_baryon change with time and our seeding mass/halo, that sounds not good.

yueyingn:
Building up large difference between intrinsic mass and dynamic mass also might cause trouble. I remember that I did spot Mdyn > 1e9 while M_intrinsic ~ 1e7 at some point last year when we did not treat BH_make_one() carefully. But Mdyn = 1e9 is almost the galaxy mass, and that black hole would fail to accrete a lot of gas particle which should have some implication on star formation.

yueyingn:
I was trying to recover as much as possible the behavior of gas accretion / baryon conservation when a black hole grows. Cap seed BH with dm particle mass and let them add together during merger will fail to swallow quite a lot of gas.

libgadget/blackhole.c
rainwoodman: ITtsounds like we need two Mass attributes to handle this. 1 The mass that is accepted by gravity, which is temporarily not conserved (set to DynSeedMass) in order to make sure the blackhole location is stable. 2. Another conserved mass that tracks the swallow and merger. Only when the swallow / merger mass >= DynSeedMass. 

On top of these too we still keep the accretion mass BH_Mass.

The accretion is already effectively faster immediately after seeding, because BH_mass < Mass. It would mean whenever a nearby gas is within the bound limit the probability of swallowing is 1, I suppose?

rainwoodman:
I agree with sbird that the type of particle we pick to gain a BH tag doesn't really matter. However we want to be a bit careful to avoid runoff BH formation that converts too much DM into baryon mass. Or (/and) perhaps we shall change the convention such that the BH particle has two masses, a DM mass and a Baryon mass. The DM mass is gained from the BH and conserved except for mergers. The baryon mass is from swallowing gas and mergers of the the baryon component.

PR #461, opened 2020-08-27
set initial BHDynMass

yueyingn:
Add a mass tracer Mtrack in the BHP struct. When SeedBHDynMass is much larger than SeedBHMass, in order to keep mass conservation, this Mtrack would stochastically swallow gas when BH is growing to SeedBHDynMass. 
Mtrack would stop functioning once Mtrack > SeedBHDynMass.
We also use the Mtrack to do BH-BH merger.

### DIFF
--- a/libgadget/blackhole.c
+++ b/libgadget/blackhole.c
@@ -41,6 +41,7 @@ struct BlackholeParams
     double BH_DFbmax; /* the maximum impact range, in physical unit of kpc. */
     int BH_DRAG; /*Hydro drag force*/
     
+    double SeedBHDynMass; /* The initial dynamic mass of BH particle */
     /************************************************************************/
 } blackhole_params;
 
@@ -53,6 +54,7 @@ typedef struct {
     MyFloat Vel[3];
     MyFloat Accel[3];
     MyIDType ID;
+    MyFloat Mtrack;
 } TreeWalkQueryBHAccretion;
 
 typedef struct {
@@ -102,6 +104,7 @@ typedef struct {
 typedef struct {
     TreeWalkQueryBase base;
     MyFloat Hsml;
+    MyFloat Mtrack;
     MyFloat BH_Mass;
     MyIDType ID;
     MyFloat FeedbackEnergy;
@@ -110,10 +113,11 @@ typedef struct {
 
 typedef struct {
     TreeWalkResultBase base;
-    MyFloat Mass;
+    MyFloat Mass; /* the accreted Mdyn */
     MyFloat AccretedMomentum[3];
     MyFloat BH_Mass;
     int BH_CountProgs;
+    MyFloat acMtrack; /* the accreted Mtrack */
 } TreeWalkResultBHFeedback;
 
 typedef struct {
@@ -144,6 +148,7 @@ struct BHPriv {
     /* These are temporaries used in the feedback treewalk.*/
     MyFloat * BH_accreted_Mass;
     MyFloat * BH_accreted_BHMass;
+    MyFloat * BH_accreted_Mtrack;
     MyFloat * Injected_BH_Energy;
 
     /* This is a temporary computed in the accretion treewalk and used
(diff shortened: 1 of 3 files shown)


sbird:
Ok, I sort of see what you are doing here, but the details are kind of lost. Why is this not the same as increasing the seed mass and letting the black hole accrete material as usual? One reason is because of the feedback that would happen...I'm not sure about this yet. I would like to know what Tiziana thinks.

yueyingn:
Hi Simeon,

Here is a plot showing what this Mtrack is doing here. It does the work of
swallowing gas particles (and maintains mass conservation) when the
blackhole is growing to the SeedDynmass. After Mtrack catches up
the SeedDynmass, everything goes back to original implementation.

I do think gas accretion/swallow plays a crucial role in BH-subgrid. We
kind of miss that in the initial phase when setting SeedDynMass large, and
things will be worse if a lot of seed black holes merge and add up their
large SeedDynMass.
But now this Mtrack can properly do the job.

sbird:
So the basic idea here is that you let mass conservation be temporarily broken to keep the BH stable in the early phases. 
My questions are:
- What is the exact problem that we want to solve with this? I think we kind of know, but it is useful to have it written out.
- Why can we not achieve a similar effect that solves this problem with a parameter change in the current model?
- What is the role of the BH mass in all of this? Does it get increased by the early quick accretion?
- What does the BH feedback do during this early accretion? Does it become very strong?
- Most importantly: how will you write this model up in a way that can be easily understood?

I just need more documentation: I understand more or less what this does, but I don't understand why exactly it is the right thing to do and I don't understand the side-effects it might have. When we talked about it on the skype call there were a lot of ideas and discussion and that is often a sign that we need to think carefully before merging.

yueyingn:
I agree that a detailed documentation is necessary for this non-trivial change.  I am going to do one more test (with normal initial Mdyn and boosted dynamic friction) and write a detailed documentation afterward.

sbird:
Great, thanks!

rainwoodman:
Let's make sure there is a clear description of dynamic mass, accretion
mass, BH Mass, MTrack(perhaps conserved mass) in the documentation. The set
of terminology will be our standard from this point on.

yueyingn:
[Initialize large Mdyn of BH in MP-Gadget3.pdf](https://github.com/MP-Gadget/MP-Gadget/files/5151455/Initialize.large.Mdyn.of.BH.in.MP-Gadget3.pdf)

yueyingn:
Hi, Here is a detailed description of this PR.

sbird:
Ok, I understand now what you are trying to do here. I agree that this has fewer side-effects than the idea of seeding the BH from a DM particle. I still find the introduction of a third mass a little inelegant, but I can't think of a better way to do it! I support this merge in principle.

libgadget/blackhole.c
```diff
@@ -1134,10 +1186,29 @@ blackhole_feedback_ngbiter(TreeWalkQueryBHFeedback * I,
         BHP(other).SwallowTime = All.Time;
         P[other].Swallowed = 1;
         O->BH_CountProgs += BHP(other).CountProgs;
-        /* Leave the swallowed BH mass around
-         * so we can work out mass at merger. */
-        O->Mass += (P[other].Mass);
-        O->BH_Mass += (BHP(other).Mass);        
+        O->BH_Mass += (BHP(other).Mass); 
+        
+        if (blackhole_params.SeedBHDynMass>0 && I->Mtrack>0){
```
sbird, line 1191: Needs a big comment

libgadget/blackhole.c
```diff
@@ -1134,10 +1186,29 @@ blackhole_feedback_ngbiter(TreeWalkQueryBHFeedback * I,
...
-         * so we can work out mass at merger. */
-        O->Mass += (P[other].Mass);
-        O->BH_Mass += (BHP(other).Mass);        
+        O->BH_Mass += (BHP(other).Mass); 
+        
+        if (blackhole_params.SeedBHDynMass>0 && I->Mtrack>0){           
+            if(I->Mtrack < blackhole_params.SeedBHDynMass && BHP(other).Mtrack < blackhole_params.SeedBHDynMass){
+                O->acMtrack += BHP(other).Mtrack;
+                double delta_m = I->Mtrack + BHP(other).Mtrack - blackhole_params.SeedBHDynMass;
+                O->Mass += ((0<delta_m)?delta_m:0);
```
sbird, line 1195: Can you avoid the ternary with DMAX?

libgadget/blackhole.c
```diff
@@ -403,6 +411,8 @@ collect_BH_info(int * ActiveParticle,int NumActiveParticle, struct BHPriv *priv,
         info.SwallowID =  BHP(p_i).SwallowID;
         info.CountProgs = BHP(p_i).CountProgs;
         info.Swallowed =  P[p_i].Swallowed;
+        info.Mtrack = BHP(p_i).Mtrack;
+        info.Mdyn = P[p_i].Mass;
```
sbird, line 415: Need some longer comments about the three masses

sbird:
Looks good to me! Let's talk about this in the telecon and then merge and do some tests.

PR #464, opened 2020-09-09
Fixes

sbird:
This is a collection of fixes to work around problems with the Frontera MPI 19.1.1. which crashes in 

Program received signal SIGSEGV, Segmentation fault.
0x00002b94f32d3935 in gsl_linalg_solve_symm_tridiag () from /opt/apps/intel19/gsl/2.5/lib/libgsl.so.23

### DIFF
--- a/libgadget/cooling_rates.c
+++ b/libgadget/cooling_rates.c
@@ -970,7 +970,7 @@ init_cooling_rates(const char * TreeCoolFile, const char * MetalCoolFile, Cosmol
     GrayOpac = gsl_interp_alloc(gsl_interp_linear,NGRAY);
     gsl_interp_init(GrayOpac,GrayOpac_zz,GrayOpac_ydata, NGRAY);
 
-    if(strnlen(TreeCoolFile,100) == 0) {
+    if(!TreeCoolFile || strnlen(TreeCoolFile,100) == 0) {
         CoolingParams.PhotoIonizationOn = 0;
         message(0, "No TreeCool file is provided. Cooling is broken. OK for DM only runs. \n");
     }
--- a/libgadget/cooling_rates.c
+++ b/libgadget/cooling_rates.c
@@ -965,8 +965,9 @@ init_cooling_rates(const char * TreeCoolFile, const char * MetalCoolFile, Cosmol
     CoolingParams.fBar = CP->OmegaBaryon / CP->OmegaCDM;
     CoolingParams.rho_crit_baryon = CP->OmegaBaryon * 3.0 * pow(CP->HubbleParam*HUBBLE,2.0) /(8.0*M_PI*GRAVITY);
 
-    /*Initialize the interpolation for the self-shielding module as a function of redshift.*/
-    GrayOpac = gsl_interp_alloc(gsl_interp_cspline,NGRAY);
+    /* Initialize the interpolation for the self-shielding module as a function of redshift.
+     * A crash has been observed in GSL with a cspline interpolator. */
+    GrayOpac = gsl_interp_alloc(gsl_interp_linear,NGRAY);
     gsl_interp_init(GrayOpac,GrayOpac_zz,GrayOpac_ydata, NGRAY);
 
     if(strlen(TreeCoolFile) == 0) {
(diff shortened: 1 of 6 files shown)


sbird:
This PR fixes the problem. The minimal commit that fixes the segfault is https://github.com/MP-Gadget/MP-Gadget/pull/464/commits/28a57e32e5d51030e20a8d9ea77a427359c68881 the others are for cleanups. 
I don't know why initing a GSL cspline crashes on Frontera: I suspect that GSL is buggy with AVX512 or installed incorrectly. But we use linear interpolation everywhere else except the neutrino code, and for this self-shielding correction it is probably the right thing anyway.

libgadget/cooling_rates.c
```diff
@@ -965,11 +965,12 @@ init_cooling_rates(const char * TreeCoolFile, const char * MetalCoolFile, Cosmol
...
 
-    /*Initialize the interpolation for the self-shielding module as a function of redshift.*/
-    GrayOpac = gsl_interp_alloc(gsl_interp_cspline,NGRAY);
+    /* Initialize the interpolation for the self-shielding module as a function of redshift.
+     * A crash has been observed in GSL with a cspline interpolator. */
+    GrayOpac = gsl_interp_alloc(gsl_interp_linear,NGRAY);
     gsl_interp_init(GrayOpac,GrayOpac_zz,GrayOpac_ydata, NGRAY);
 
-    if(strlen(TreeCoolFile) == 0) {
+    if(strnlen(TreeCoolFile,100) == 0) {
```
rainwoodman, line 973: What if TreeCoolFile is NULL? Probably if(TreeCollFile && TreeCoolFile[0] == 0) ?

rainwoodman:
I agree that numbers in TreeCool are probably with very large uncertainties anyway, such that linear shouldn't be so bad. 

Does this also occur on Frontera with a minimal program that only calls the problematic gsl routine? Is the Frontera support aware of this?

sbird:
> I agree that numbers in TreeCool are probably with very large uncertainties anyway, such that linear shouldn't be so bad.

Right - these particular numbers are generated from Ali Rahmati's radiative transfer simulations and are uncertain at least to the 10% level.

> Does this also occur on Frontera with a minimal program that only calls the problematic gsl routine? Is the Frontera support aware of this?

I haven't checked but I probably should! They are not aware.

libgadget/cooling_rates.c
sbird: if(!TreeCoolFile || TreeCoolFile[0] == 0)

sbird:
The minimal example does indeed segfault, I will open a ticket.

sbird:
Done:
https://portal.tacc.utexas.edu/tacc-consulting/-/consult/tickets/63839?_consultrt_WAR_consultrtportlet_ticketSubmitted=true
and the reproducer is:
 #include <gsl/gsl_interp.h>
#include <stdio.h>
#define NGRAY 6

int main(int argc, char **argv)
{
    double redshift = 2.5;
    double GrayOpac_ydata[NGRAY] = { 2.59e-18, 2.37e-18, 2.27e-18, 2.15e-18, 2.02e-18, 1.94e-18};
    double GrayOpac_zz[NGRAY] = {0, 1, 2, 3, 4, 5};

    if(argc >= 2)
        redshift = atof(argv[1]);
    gsl_interp * GrayOpac = gsl_interp_alloc(gsl_interp_cspline,NGRAY);
    /* A crash has been observed here with a cspline interpolator. */
    gsl_interp_init(GrayOpac,GrayOpac_zz,GrayOpac_ydata, NGRAY);
    double greyopac = gsl_interp_eval(GrayOpac, GrayOpac_zz, GrayOpac_ydata,redshift, NULL);
    printf("Interp at %g is %g\n", redshift, greyopac);
    return 0;
}

compile with:
mpicc -Wall -O0 test_gsl.c `pkg-config --cflags gsl` -o test_gsl `pkg-config --libs gsl`

rainwoodman:
Thanks for the update! It's probably memory alignment biting again.

PR #467, opened 2020-09-19
Cleanups to remove Ti_Current

sbird:
I removed Ti_Current which was embedded deep in the code. I initially started to do something else, but got distracted, hence the strange branch name :-)

### DIFF
--- a/libgadget/gravpm.c
+++ b/libgadget/gravpm.c
@@ -203,10 +203,6 @@ static int pm_mark_region_for_node(int startno, int rid, int * RegionInd, const 
                 int p = nop->s.suns[i];
                 RegionInd[p] = rid;
 #ifdef DEBUG
-                /* when we are in PM, all particles must have been synced. */
-                if (P[p].Ti_drift != All.Ti_Current) {
-                    abort();
-                }
                 /* Check for particles outside of the node. This should never happen,
                 * unless there is a bug in tree build, or the particles are being moved.*/
                 int k;
--- a/libgadget/run.c
+++ b/libgadget/run.c
@@ -158,7 +158,7 @@ run(int RestartSnapNum)
     gravpm_init_periodic(&pm, All.BoxSize, All.Asmth, All.Nmesh, All.G);
 
     DomainDecomp ddecomp[1] = {0};
-    init(RestartSnapNum, ddecomp);          /* ... read in initial model */
+    inttime_t Ti_Current = init(RestartSnapNum, ddecomp);          /* ... read in initial model */
 
     /* Stored scale factor of the next black hole seeding check*/
     double TimeNextSeedingCheck = All.Time;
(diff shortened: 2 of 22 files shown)


sbird:
@rainwoodman Any comments? This came out nicer than I expected.

sbird:
I especially like that if we wanted to have particle specific drift times again the code would now behave roughly correctly, I think.

libgadget/gravpm.c
```diff
@@ -203,10 +203,6 @@ static int pm_mark_region_for_node(int startno, int rid, int * RegionInd, const
                 int p = nop->s.suns[i];
                 RegionInd[p] = rid;
 #ifdef DEBUG
-                /* when we are in PM, all particles must have been synced. */
```
rainwoodman, line 206: I suggest preserving a weaker version of this, e.g. all Ti_drift is identical to P[0]'s.

libgadget/run.c
```diff
@@ -158,7 +158,7 @@ run(int RestartSnapNum)
     gravpm_init_periodic(&pm, All.BoxSize, All.Asmth, All.Nmesh, All.G);
 
     DomainDecomp ddecomp[1] = {0};
-    init(RestartSnapNum, ddecomp);          /* ... read in initial model */
+    inttime_t Ti_Current = init(RestartSnapNum, ddecomp);          /* ... read in initial model */
```
rainwoodman, line 161: Suggest using an inttime_t * on the args as return value, just in case we may return more down the road.

libgadget/tests/test_timebinmgr.c
```diff
@@ -64,27 +62,27 @@ static void test_conversions(void ** state) {
 
 static void test_skip_first(void ** state) {
 
-    setup_sync_points(All.TimeIC, All.TimeIC);
+    setup_sync_points(All.TimeIC, All.TimeMax, All.TimeIC, 0);
```
rainwoodman, line 65: Situations like this makes me wish we can use named arguments in C.

sbird: +100.

Even more annoying when, eg, arguments of type double may be silently converted to int...

libgadget/gravpm.c
sbird: I moved the check to rebuild_activelist, so that it is run on all timesteps and is a strictly stronger check. I also dropped the DEBUG.

libgadget/run.c
sbird: I think we may end up doing that, but let me defer it a little. The time variables are a bit messed up and complicated: I would like to move them out of allvars.h and this may finish with them outside of init altogether.

sbird:
The final commit 2c02806 makes small changes to output as an apparent result of floating point optimizations, apart from that no change to output, so I will merge.

PR #474, opened 2020-10-28
add Nnu_transfer parameter to specify the number of neutrino species used in the transfer function

yueyingn:
This is a clean PR now.


sbird:
Can you auto-detect the number of columns? For example, add a strtok loop in read_power_table that computes the number of columns and passes it to read_transfer and read_power, then read_transfer can use that to compute nnu (and read_power can just check it is 2).

PR #475, opened 2020-10-28
Detect column numbers of the input powerspectrum and transfer fu…

yueyingn:
…nction, and guess neutrino species accordingly


sbird:
Looks good to me!

PR #476, opened 2020-11-09
Adding 21CMFAST to MP-Gadget

daviesje:
I've added an implementation of the excursion set reionisation algorithm, 21CMFAST, to MP-Gadget.

Github won't let me create a pull request to a new branch, so I'm just putting it here for now.

main changes are as follows:
- uvbg.c includes the 21CMFAST implementation, with functions for allocating, populating and saving the grids
- in sfr_eff.c, whenever a star particle is created, we add to a global stellar mass grid, which is stored on all ranks and summed before the UVBG is calculated
- cooling_uvfluc.c has a function which gets the UV background from the excursion set calculations, and is used instead of the TREECOOL files when a flag is passed in the parameter files
- timebinmgr.c includes more syncpoints, starting at z=20, every X Myr (free parameter), where 21CMFAST is run.


gadget/params.c
```diff
@@ -312,6 +312,21 @@ create_gadget_parameter_set()
     param_declare_double(ps, "NuPartTime", OPTIONAL, 0.3333333, "Scale factor at which to turn on hybrid neutrino particles.");
     /*End parameters for the massive neutrino model*/
 
+    /*Parameters for the Excursion Set Algorithm*/
+    param_declare_int(ps, "ExcursionSetFlag", OPTIONAL, 0, "Use the excursion set instead of the global UV field");
+    param_declare_int(ps, "UVBGdim", OPTIONAL, 64, "Side length of the excursion set grid.");
+    param_declare_int(ps, "UVBGSaveGrids", OPTIONAL, 1, "Flag to save reionisation grids.");
+    param_declare_int(ps, "ReionFilterType", OPTIONAL, 0., "Filter type for Excursion set");
```
sbird, line 319: What are the possibilities for this option?

excursionset_rundown.md
```diff
@@ -0,0 +1,16 @@
...
+*Syncpoints are set up every 10 Myr or so (free parameter) in order to trigger the excursion set.
+*Whenever a star particle is created, mass is added to a spatial grid where it formed.
+*When we reach one of the UV syncpoints, stellar mass and total mass grids are created by nearest gridpoint, and decomposed into a slab for each task.
+*We then step through filters of different scales using FFTW and count photons in each cell (from stellar mass * photons per baryon) to determine which cells are ionised, and calculate J21 if so.
+*This J21 is then converted into ionisation and photoheating rates at a certain spectral slope (free parameter) and passed into the existing cooling functions.
+
+
+##Notes:
+
+*Conversion factors from J21 to rates are passed in a table file via the same parameter as the TREECOOL files, if the excursion set flag is on or off it will treat this file as a conversion table or TREECOOL file repsectively.
```
sbird, line 13: Could you create a new option for the excursion set file? It could be ignored if the excursion set flag is off.

gadget/params.c
```diff
@@ -312,6 +312,21 @@ create_gadget_parameter_set()
     param_declare_double(ps, "NuPartTime", OPTIONAL, 0.3333333, "Scale factor at which to turn on hybrid neutrino particles.");
     /*End parameters for the massive neutrino model*/
 
+    /*Parameters for the Excursion Set Algorithm*/
+    param_declare_int(ps, "ExcursionSetFlag", OPTIONAL, 0, "Use the excursion set instead of the global UV field");
+    param_declare_int(ps, "UVBGdim", OPTIONAL, 64, "Side length of the excursion set grid.");
+    param_declare_int(ps, "UVBGSaveGrids", OPTIONAL, 1, "Flag to save reionisation grids.");
```
sbird, line 318: Does this need to be an option or can we just leave it always on?

daviesje: If we always want to save the grids on every output step I can remove it. I put it there along the same lines as the outputFOF flag in case we only wanted the particle info for some runs.

gadget/params.c
daviesje: Filter type for Excursion set: 0 = real-space top-hat, 1 = k-space top-hat, 2 = gaussian. I have gone back and updated the description for this parameter now and will push it with the other changes

excursionset_rundown.md
daviesje: Do you mean just have another string option which is passed into init_cooling instead when the excursion set is on? Or do you mean completely separate the loading and storage of each table, so they could possibly both be loaded in the same run?

libgadget/uvbg.c
```diff
@@ -0,0 +1,971 @@
...
+    // definitely need to be changed! There is no way we should have to search
+    // all of the particles to find out what slab they sit on, and then loop
+    // through all particles again n_slab times!
+    double box_size = All.BoxSize;
+
+    #pragma omp parallel for
+    for(int ii = 0; ii < PartManager->NumPart; ii++) {
+        if((!P[ii].IsGarbage) && (!P[ii].Swallowed) && (P[ii].Type < 5)) {
+            ptrdiff_t ix = pos_to_ngp(P[ii].Pos[0], box_size, uvbg_dim);
+            UVRegionInd[ii] = searchsorted(&ix, slab_ix_start, nranks, sizeof(ptrdiff_t), compare_ptrdiff, -1, -1);
```
sbird, line 303: What is this code doing? Is it finding the nearest slab?

libgadget/uvbg.c
```diff
@@ -0,0 +1,971 @@
...
+        MPI_Reduce(UVBGgrids.stars + grid_index(ix_start, 0, 0, uvbg_dim, INDEX_REAL), buffer_stars_slab, nix*uvbg_dim*uvbg_dim, MPI_FLOAT, MPI_SUM, i_r, MPI_COMM_WORLD);
+        MPI_Reduce(UVBGgrids.prev_stars + grid_index(ix_start, 0, 0, uvbg_dim, INDEX_REAL), buffer_sfr, nix*uvbg_dim*uvbg_dim, MPI_FLOAT, MPI_SUM, i_r, MPI_COMM_WORLD);
+        
+        // TODO(smutch): These could perhaps be precalculated?
+        const float inv_dt = (float)(1.0 / (time_to_present(UVBGgrids.last_a) - time_to_present(All.Time)));
+        message(0, "UVBG calculation dt = %.2e Myr\n", (1.0 / inv_dt));
+
+        #pragma omp parallel for
+        for(int ii=0; ii < buffer_size; ii++) {
+            buffer_sfr[ii] = (buffer_stars_slab[ii] - buffer_sfr[ii]) * inv_dt / All.UnitTime_in_Megayears;
```
sbird, line 360: I see what you are doing but it took a while, please add a comment.

sbird (commented):
I looked through it it looks in pretty good shape, actually! Might be some scalability issues for the big run we should work on and I had some clarification comments (and may have more as I think about it). I like the use of the syncpoints.

gadget/params.c
```diff
@@ -312,6 +312,21 @@ create_gadget_parameter_set()
     param_declare_double(ps, "NuPartTime", OPTIONAL, 0.3333333, "Scale factor at which to turn on hybrid neutrino particles.");
     /*End parameters for the massive neutrino model*/
 
+    /*Parameters for the Excursion Set Algorithm*/
+    param_declare_int(ps, "ExcursionSetFlag", OPTIONAL, 0, "Use the excursion set instead of the global UV field");
+    param_declare_int(ps, "UVBGdim", OPTIONAL, 64, "Side length of the excursion set grid.");
```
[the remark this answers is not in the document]
daviesje: This is just a poor description by me. Its the grid dimension in cells, such that the resolution is BoxSize/UVBGdim. I'll update this.

Options.mk.example
```diff
@@ -7,6 +7,10 @@ OPTIMIZE =  -fopenmp -O3 -g -Wall -ffast-math
 #This is a good non-optimized default for debugging
 #OPTIMIZE =  -fopenmp -O0 -g -Wall
 
+#FFTW for the current excursion set implementation
+FFTW_INCL = -I/apps/skylake/software/mpi/gcc/7.3.0/openmpi/3.0.0/fftw/3.3.7/include
+FFTW_LIBS = -L/apps/skylake/software/mpi/gcc/7.3.0/openmpi/3.0.0/fftw/3.3.7/lib -lfftw3f -lfftw3f_mpi
```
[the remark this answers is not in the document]
daviesje: This was going to be one of the next steps, I spent some time trying to see if I could use existing functions to decompose and FFT these grids in a similar way to how long-range force is done, but I haven't quite got there yet, and I might need some help fully understanding that part of the code.

platform-options/Options.mk.macos
```diff
@@ -4,6 +4,11 @@ OPTIMIZE = -O3 -g -Wall -ffast-math -march=native
 # For some reason apple clang doesn't build openmp into their binary.
 # You must install openmpi from homebrew using gcc first with, eg, "brew install open-mpi --cc=gcc-9 --build-from-source"
 OPTIMIZE += -fno-common -fopenmp
+# Find the sdk path on Mac
+OPT += -isysroot $(shell xcrun -sdk macosx --show-sdk-path)
+
+GSL_INCL = $(shell pkg-config --cflags gsl)
+GSL_LIBS = $(shell pkg-config --libs gsl)
```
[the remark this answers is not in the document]
daviesje: Not sure, this was one of Simon's additions before I started working on this, I can send him an email about it though.

libgadget/timebinmgr.h
```diff
@@ -60,7 +61,7 @@ inttime_t out_from_ti(inttime_t ti);
 
 int OutputListAction(ParameterSet * ps, char * name, void * data);
 void set_sync_params(int OutputListLength, double * OutputListTimes);
-void setup_sync_points(double TimeIC, double TimeMax, double no_snapshot_until_time, int SnapshotWithFOF);
+void setup_sync_points(double TimeIC, double TimeMax, double UVBGTimestep, double no_snapshot_until_time, int SnapshotWithFOF);
```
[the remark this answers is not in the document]
daviesje: I thought about not using the syncpoints here, and calculating the UVBG if it has been longer than UVBGTimestep since the last calculation, by looking at the time difference between steps. Not quite sure how the cadence would work out in that case though.

libgadget/uvbg.c
daviesje: Yes, it fills out an array with which slab each particle is on by comparing the particle x coordinate with the array of slab start coordinates. As the comment says above it there's probably a better way to do the decomposition but I haven't looked at this too closely yet, especially since I was planning to completely change the decomposition later on when switching to pfft.

gadget/params.c
sbird: Ah I see now, thanks!

Options.mk.example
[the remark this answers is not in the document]
sbird: It is a bit opaque. Let me know if you need some help: it should map fairly directly to fftw functions.

examples/J21_to_rates_test.txt
```diff
@@ -0,0 +1,26 @@
...
+3.200000e+00 2.028544e-12 4.301959e-13 5.844777e-15 5.546412e-12 2.566384e-12 5.548895e-14
+3.400000e+00 1.962736e-12 3.676447e-13 4.298541e-15 5.162112e-12 2.099982e-12 3.977389e-14
+3.600000e+00 1.901040e-12 3.146147e-13 3.163537e-15 4.816208e-12 1.723149e-12 2.853534e-14
+3.800000e+00 1.843086e-12 2.695761e-13 2.329777e-15 4.503782e-12 1.417690e-12 2.049074e-14
+4.000000e+00 1.788543e-12 2.312606e-13 1.716865e-15 4.220676e-12 1.169316e-12 1.472711e-14
+4.200000e+00 1.737121e-12 1.986135e-13 1.265987e-15 3.963350e-12 9.667587e-13 1.059395e-14
+4.400000e+00 1.688561e-12 1.707552e-13 9.340829e-16 3.728780e-12 8.010980e-13 7.627360e-15
+4.600000e+00 1.642631e-12 1.469504e-13 6.895978e-16 3.514371e-12 6.652483e-13 5.496187e-15
+4.800000e+00 1.599124e-12 1.265830e-13 5.093924e-16 3.317884e-12 5.535596e-13 3.963824e-15
+5.000000e+00 1.557853e-12 1.091351e-13 3.764847e-16 3.137382e-12 4.615109e-13 2.861064e-15
```
sbird, line 26: How is this file generated? What if we want to change the UVB?

excursionset_rundown.md
sbird: I think my answer depends on another question: how does the excursion set model connect to the normal cooling code? I think that at the moment once a cell is reionized, J21 is set to zero and it just uses the rates as the old cooling code doe. But the excursion set code is still called every few Myr? Is this because the tail of reionization may finish very (arbitrarily) late?

gadget/params.c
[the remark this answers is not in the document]
sbird: Thanks sounds sensible

gadget/params.c
sbird: Do I understand right that the saved grids are needed to resume from the snapshot? In which case I think we would always want them when the ExcursionSetFlag is on?

libgadget/timebinmgr.h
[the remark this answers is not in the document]
sbird: I actually really like your use of extra syncpoints, it's a neat application. Looking at the time difference is clever too, but I like this more.

platform-options/Options.mk.macos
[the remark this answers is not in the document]
sbird: It's not very important, I was just wondering.

libgadget/sfr_eff.c
```diff
@@ -505,6 +508,20 @@ static int make_particle_star(int child, int parent, int placement)
     STARP(child).BirthDensity = oldslot.Density;
     /*Copy metallicity*/
     STARP(child).Metallicity = oldslot.Metallicity;
+
+    // TODO(smutch): Use CIC
+    int coord[3] = {0};
+    for(int ii=0; ii<3; ii++) {
+        coord[ii] = pos_to_ngp(P[child].Pos[ii], All.BoxSize, All.UVBGdim);
```
sbird, line 515: Shouldn't this correct for the CurrentParticleOffset?

libgadget/uvbg.c
sbird: I see! I'm happy with doing this (and maybe we change when moving to pfft), but a comment would be good.

libgadget/uvbg.c
```diff
@@ -0,0 +1,971 @@
...
+        }
+
+
+        // fill the local buffer for this slab
+        // TODO(smutch): This should become CIC
+        unsigned int count_mass = 0;
+        #pragma omp parallel for reduction(+:count_mass)
+        for(int ii = 0; ii < PartManager->NumPart; ii++) {
+            if(UVRegionInd[ii] == i_r) {
+                int ix = pos_to_ngp(P[ii].Pos[0], box_size, uvbg_dim) - ix_start;
```
sbird, line 329: I think you need CurrentParticleOffset here

excursionset_rundown.md
daviesje: J21 is zero until the cell is ionized. After this, the J21 calculated by the excursion set (via the SFR) is multiplied by the photorates given in this table at the chosen spectral slope, which is passed into the cooling code instead of the TREECOOL redshift-dependent photo-rates. As the code exists right now, only one of these files will be used.

sbird: I see, I misread the conditions, sorry. 

I think my concern is really that post-reionization we should retain the good agreement with the lyman alpha forest - I'm worried that once quasars take over the SFR will no longer be a good evolution proportionality constant for the UVB and we need a redshift dependent TREECOOL as well. But I might easily be misunderstanding the model, so please correct me if I'm wrong.

sbird: PS I'm happy to merge this as a reionization era model and not worry about that: people who want lyman-alpha can continue with the older model. But to enable it for the big run we would need to make sure it played nicely with post-reionization physics.

libgadget/timebinmgr.c
```diff
@@ -95,27 +96,60 @@ void set_sync_params(int OutputListLength, double * OutputListTimes)
...
+            SyncPoints[NSyncPoints].write_fof = 0;
+            SyncPoints[NSyncPoints].calc_uvbg = 1;
+            NSyncPoints++;
+            //message(0,"added UVBG syncpoint at a = %.3f, Nsync = %d\n",uv_a,NSyncPoints);
+
+            // TODO(smutch): OK - this is ridiculous (sorry!), but I just wanted to quickly hack something...
+            // TODO(jdavies): fix low-z where delta_a > 10Myr
+            double delta_a = 0.0001;
+            double lbt = time_to_present(uv_a);
+            double delta_lbt = 0.0;
```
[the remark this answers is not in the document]
daviesje: The minimum time between UVBG calculations is set by the UVBGTimestep, this loop just steps through scale factors until we hit a point where that much time has past, and sets the next syncpoint there to avoid more integration.

examples/J21_to_rates_test.txt
daviesje: This is the conversion between spectral slopes (left column) and ionisation/heating rates at J21=1. It was calculated by performing the integral over the spectrum (and cross section) above each ionisation threshold. I could include this code as a util file somewhere.

libgadget/timebinmgr.c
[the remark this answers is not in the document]
sbird: Yup - I mean: what is the physical reason that you set UVBGTimestep to the value that it has? Why is 10Myr the important value?

examples/J21_to_rates_test.txt
sbird: Thanks I'd really appreciate that.

excursionset_rundown.md
daviesje: That's a good point, I'll separate the loading for now, and later on put in a way to switch back to the global UVBG after reionisation

libgadget/timebinmgr.c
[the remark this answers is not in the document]
daviesje: 10Myr was originally chosen because it was a small enough cadence to cover the lifespan of the OB stars that produced most of the UV photons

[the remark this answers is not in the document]
sbird: Aha, I see, thank you. So if we find that this is an expensive part of the simulation, is there still the option to increase UVBGTimestep? Or have you found that doing that causes the results to not be physically valid anymore?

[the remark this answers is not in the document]
daviesje: Actually I might have gotten a bit confused, that timestep was chosen for the lifetime of the large stars, but I think it was mostly relevant for the supernovae feedback from those stars in Meraxes. I haven't done much testing of the timestep, and because we plug in the rates from the excursion set into the existing cooling model, the way the timestep affects things might change. I suspect it will be okay as long as the density or position of gas, or the size of the bubbles doesn't change drastically within the timestep.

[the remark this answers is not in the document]
sbird: Right - one reason I'm asking is that the PMstep has a displacement constraint: a new PM step is triggered whenever particles move through 0.1 of a grid cell. The PM grid cells are much smaller than the UVBG cells, so it might be enough to use the PM timestep (which I think is longer than 10Myr). But I'm not sure about the bubble size, which I guess increases at the speed of light?

gadget/params.c
daviesje: That's true, I don't think I considered the checkpointing when making this flag. It's been removed.

sbird: Thanks

libgadget/timebinmgr.c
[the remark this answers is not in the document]
daviesje: We could definitely try increasing the timestep to see if that works, ionisation fronts move at speeds anywhere between 1e2 km/s in dense regions and 1e5 km/s in the voids during overlap. If we have a reionisation scenario that mostly happens between z=8 to z=6 we have about 300Myr, which would be ~30 steps at a 10Myr spacing, meaning optimistically every step covers at most  ~3-4% of reionisation. I don't think we'd want to go much coarser than this, otherwise particles might stay neutral for a significant portion of the EoR. However, I'd need to test longer steps before understanding how it would effect the evolution of the particles.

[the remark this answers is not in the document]
sbird: That makes sense, thanks for the information!

libgadget/cosmology.c
```diff
@@ -24,6 +25,9 @@ void init_cosmology(Cosmology * CP, const double TimeBegin)
     CP->OmegaCDM = CP->Omega0 - CP->OmegaBaryon;
     CP->OmegaK = 1.0 - CP->Omega0 - CP->OmegaLambda;
 
+    const double G = GRAVITY / pow(All.UnitLength_in_cm, 3) * All.UnitMass_in_g * pow(All.UnitTime_in_s, 2);
+    CP->RhoCrit = 3.0 * CP->Hubble * CP->Hubble / (8.0 * M_PI * G);  // in internal units
```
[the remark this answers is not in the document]
daviesje: done

libgadget/uvbg.c
```diff
@@ -0,0 +1,971 @@
...
+#include "physconst.h"
+#include "walltime.h"
+
+// TODO(smutch): See if something equivalent is defined anywhere else
+#define FLOAT_REL_TOL (float)1e-5
+
+struct UVBGgrids_type UVBGgrids;
+
+double integrand_time_to_present(double a, void *dummy)
+{
```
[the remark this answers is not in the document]
daviesje: This should be done now

libgadget/allvars.h
```diff
@@ -136,6 +136,21 @@ extern struct global_data_all_processes
...
+    double ReionRBubbleMin;
+    double ReionDeltaRFactor;
+    double ReionGammaHaloBias;
+    double ReionNionPhotPerBary;
+    double AlphaUV;
+    double EscapeFraction;
+    double UVBGTimestep;
+    int ReionFilterType;
+    int UVBGSaveGrids;
+    /*End Parameters for the Excursion Set Algorithm*/
```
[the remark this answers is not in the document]
daviesje: So far I've moved everything that's only used in one file to their respective structures. There are a couple parameters there that are used in multiple files that I've kept there until I figure out how best to approach it.

libgadget/uvbg.c
daviesje: I've added the offset to pos_to_ngp

libgadget/uvbg.c
```diff
@@ -0,0 +1,971 @@
...
+    
+    UVBGgrids.deltax = fftwf_alloc_real((size_t)(slab_n_complex * 2));  // padded for in-place FFT
+    UVBGgrids.deltax_filtered = fftwf_alloc_complex((size_t)(slab_n_complex));
+    UVBGgrids.stars_slab = fftwf_alloc_real((size_t)(slab_n_complex * 2));  // padded for in-place FFT
+    UVBGgrids.stars_slab_filtered = fftwf_alloc_complex((size_t)(slab_n_complex));
+    UVBGgrids.sfr = fftwf_alloc_real((size_t)(slab_n_complex * 2));  // padded for in-place FFT
+    UVBGgrids.sfr_filtered = fftwf_alloc_complex((size_t)(slab_n_complex));
+    UVBGgrids.xHI = fftwf_alloc_real((size_t)slab_n_real);
+    UVBGgrids.z_at_ionization = fftwf_alloc_real((size_t)(slab_n_real));
+    UVBGgrids.J21_at_ionization = fftwf_alloc_real((size_t)(slab_n_real));
```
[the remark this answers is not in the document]
daviesje: I know the fftw/pfft allocators ensure proper alignment for the FFTs, but looking at petapm.c they seem to just be typecast after `mymalloc`. Is there something I need to do here to ensure they are allocated correctly or can I just typecast real/complex fields to double/pfft_complex?

libgadget/uvbg.c
```diff
@@ -0,0 +1,1204 @@
...
+{
+    message(0, "Assigning slabs to MPI cores...\n");
+
+    int uvbg_dim = All.UVBGdim;
+    // Allocations made in this function are free'd in `free_reionization_grids`.
+
+    //TODO(jdavies): probably move the first half of this to an initialize function
+    //TODO: do I need to reinitialise this after petapm? I don't see pfft_cleanup anywhere
+    pfftf_init();
+    pfftf_plan_with_nthreads(omp_get_max_threads());
```
sbird, line 120: No, you don't need to initialise this a second time.

libgadget/uvbg.c
```diff
@@ -0,0 +1,1204 @@
...
+    if( pfftf_create_procmesh_2d(MPI_COMM_WORLD, np[0], np[1], &UVBGgrids.comm_cart_2d) ){
+        endrun(0, "Error: This test file only works with %td processes.\n", np[0]*np[1]);
+    }
+
+    MPI_Comm comm_cart_2d = UVBGgrids.comm_cart_2d;
+
+    // find out what slab each rank should get
+    ptrdiff_t local_ni[3], local_i_start[3];
+    ptrdiff_t local_no[3], local_o_start[3];
+    ptrdiff_t local_n_complex = pfftf_local_size_dft_r2c_3d(n, comm_cart_2d, pfft_flags, local_ni, local_i_start, local_no, local_o_start);
```
sbird, line 152: Use pfft_ functions instead of pfftf_ so that it automagically works out the floating point precision

libgadget/uvbg.c
```diff
@@ -0,0 +1,1204 @@
...
+    UVBGgrids.slab_no = slab_no;
+    MPI_Allgather(&local_no, sizeof(ptrdiff_t)*3, MPI_BYTE, slab_no, sizeof(ptrdiff_t)*3, MPI_BYTE, MPI_COMM_WORLD);
+
+    ptrdiff_t *slab_o_start = mymalloc("slab_o_start",sizeof(ptrdiff_t) * n_ranks * 3); ///< array first x cell of every rank
+    UVBGgrids.slab_o_start = slab_o_start;
+    MPI_Allgather(&local_o_start, sizeof(ptrdiff_t)*3, MPI_BYTE, slab_o_start, sizeof(ptrdiff_t)*3, MPI_BYTE, MPI_COMM_WORLD);
+
+    ptrdiff_t *slab_n_complex = mymalloc("slab_n_complex",sizeof(ptrdiff_t) * n_ranks); ///< array of allocation counts for every rank
+    UVBGgrids.slab_n_complex = slab_n_complex;
+    MPI_Allgather(&local_n_complex, sizeof(ptrdiff_t), MPI_BYTE, slab_n_complex, sizeof(ptrdiff_t), MPI_BYTE, MPI_COMM_WORLD);
```
sbird, line 173: This doesn't look right to me - the allocations will be a little different on different ranks and with different padding. Try to reuse the code in petapm.c, especially the petapm_region_init_strides function. There is also a pm struct which will be useful.

libgadget/uvbg.c
daviesje: I was under the impression (from the pfft user guide) that the pfft_ functions only accepted and returned double precision. The grids in meraxes/21CMFAST were single precision using fftwf so I just switched to pfftf. If I change the functions over to pfft_ what kind of casting will I need to do to the inputs/outputs of the functions?

libgadget/uvbg.c
daviesje: These are meant to be arrays of the size and offsets of each region that are stored on all ranks. If I'm reading the init_strides function correctly, this sets the strides to `(size[2]*size[1], size[2], 1)`. The `grid_index` function does the same thing, except it takes `size[1]` as an argument and calculates `size[2]` based on whether the grid is in realspace, is padded for inplace fft, or in k-space (I've checked and these dimensions match the size arrays). I definitely think its a good idea to re-use the petapm functions, but I think the indexing issue is elsewhere. I'm probably missing something so I'll replace the functions and see if I get the same issue

libgadget/uvbg.c
sbird: Ok, then I am confused about pfft, I think. What I thought happened was that the pfft.h file is compiled with a prototype which contains either float or double types depending on the "--enable-single" flag, but I guess not.

Note though that the version of pfft we compile doesn't support single precision floating point because the "--enable-single" flag is not passed in (which is why the travis build fails), so to keep using single precision grids you will need to alter the build system to turn that on.

libgadget/uvbg.c
sbird: I'm not completely sure where, but I definitely remember tripping over a subtlety with the indexing before, which petapm.c takes care of. It might not be this, because what you say is roughly what they do. But I would still ask you to reuse the functions because it is good practice to have things implemented in only one place.

libgadget/uvbg.c
daviesje: If there will be room in memory for these grids to be double precision (at the moment there are 6 grids using the fft) then I can just convert these to double precision.

sbird: That's probably easiest for now - we can try it as a later optimization if this turns out to be a limitation.

libgadget/uvbg.c
daviesje: I fixed the indexing issue in the filter function (I really should have caught it earlier), and I'm getting reasonable bubbles again.

sbird: Yes that looks good! Nice.

daviesje:
The excursion set has now been fully migrated to pfft, all grids are allocated using `mymalloc`, and the 1D slab decomposition has been replaced by a 2D (x,y) decomposition which should scale better. However there are still a few issues and things I'm working on, which I'll list here:

- There are still 3 global grids (J21, stars and prev_stars) which are fully stored on each rank. If I want to distribute these I'll have to add some communication to where stars are created and added to the grid, and where the local UVBG is fetched.

- In order to use the `pfft_` functions, the grids involved in the FFTs are now double precision, but the grids storing the inputs and outputs are still single precision. Some of these may have to be changed to avoid precision loss. In particular `populate_grids` is a bit messy since it came from two different parts of meraxes, I plan on simplifying this function a bit.

- There are some unused grids which were useful results in meraxes, but are not connected to the rest of MP-Gadget or saved at the moment, these are:
    - J21 at ionisation (this was used for reionisation feedback in meraxes)
    - redshift at ionisation
    - neutral fraction (this seems important, but is output anyway for the particles via the cooling code, it is useful for debugging differences between the excursion set output and cooling code output)
    - SFR, this would be used to calculate J21, however the bursty star formation can cause some issues where the equilibrium ionisation calculation in the cooling code would cause particles to rapidly ionise and recombine on un-physical timescales. As a result we smooth the cumulative star formation over the hubble time (there were similar problems in meraxes and we used the same workaround, however we could also experiment with different smoothing scales).

- There is a lot of code in `petapm.c` that I'm sure could be re-used to improve the excursion set further, in particular how the mass and stellar grids are constructed and distributed.

tools/print_J21coeffs.c
```diff
@@ -0,0 +1,140 @@
...
+    }
+    printf("slope = %.1f | R(HI) = %.3e | G(HI) = %.3e | R(HeI) = %.3e | G(HeI) =  %.3e | R(HeII) = %.3e | G(HeII) = %.3e\n",
+           slope,Rc[0],Gc[0],Rc[1],Gc[1],Rc[2],Gc[2]);
+
+
+    fprintf(f_out,"%.6e %.6e %.6e %.6e %.6e %.6e %.6e\n",slope,Rc[0],Rc[1],Rc[2],Gc[0],Gc[1],Gc[2]);
+  }
+  fclose(f_out);
+  return 0;
+}
```
sbird, line 140: This file needs some comments! Also, could you not re-write it as a 20 line python script?

daviesje: That's true, I took this from an older project when I needed some coefficients to test with. I'll rewrite it in python.

sbird:
Apologies for the slightly slow reply, it was a holiday here.

> The excursion set has now been fully migrated to pfft, all grids are allocated using `mymalloc`, and the 1D slab decomposition has been replaced by a 2D (x,y) decomposition which should scale better. However there are still a few issues and things I'm working on, which I'll list here:

Nice. This is looking much more like something we can merge.

>     * There are still 3 global grids (J21, stars and prev_stars) which are fully stored on each rank. If I want to distribute these I'll have to add some communication to where stars are created and added to the grid, and where the local UVBG is fetched.

How large are these grids? If they are small enough to be stored on all ranks then I would be ok doing that. If not then it should be fairly easy to have them follow the domain decomposition. The tricky bit would be updating them when the PosOffset changes.

>     * In order to use the `pfft_` functions, the grids involved in the FFTs are now double precision, but the grids storing the inputs and outputs are still single precision. Some of these may have to be changed to avoid precision loss. In particular `populate_grids` is a bit messy since it came from two different parts of meraxes, I plan on simplifying this function a bit.

Ok.

>     * There are some unused grids which were useful results in meraxes, but are not connected to the rest of MP-Gadget or saved at the moment, these are:
>       
>       * J21 at ionisation (this was used for reionisation feedback in meraxes)
>       * redshift at ionisation
>       * neutral fraction (this seems important, but is output anyway for the particles via the cooling code, it is useful for debugging differences between the excursion set output and cooling code output)
>       * SFR, this would be used to calculate J21, however the bursty star formation can cause some issues where the equilibrium ionisation calculation in the cooling code would cause particles to rapidly ionise and recombine on un-physical timescales. As a result we smooth the cumulative star formation over the hubble time (there were similar problems in meraxes and we used the same workaround, however we could also experiment with different smoothing scales).

reionization redshift you could add to the sph_particle_data struct and save with the particle tables. The others I am happy to hide behind a DEBUG flag, I think. For the SFR, can we use the regular SFR from the star formation code? Or does the excursion set SFR do something different?

>     * There is a lot of code in `petapm.c` that I'm sure could be re-used to improve the excursion set further, in particular how the mass and stellar grids are constructed and distributed.

Once this is taken care of the next thing is to run a bigger validation. Are you set up to do that?

daviesje:
> Apologies for the slightly slow reply, it was a holiday here.

No worries, I'm also starting to apply for postdocs at the moment so work on this PR has been a bit slow for me

> How large are these grids? If they are small enough to be stored on all ranks then I would be ok doing that. If not then it should be fairly easy to have them follow the domain decomposition. The tricky bit would be updating them when the PosOffset changes.

Ideally we would want something a bit better than 1 comoving Mpc resolution, so it could be a few GB per rank for larger boxes.

> Once this is taken care of the next thing is to run a bigger validation. Are you set up to do that?

I've got an account ready on Frontera. Are there any more specific parameter sets you use for this, or tests that I should run?

sbird:
> Ideally we would want something a bit better than 1 comoving Mpc resolution, so it could be a few GB per rank for larger boxes.

hmm, yes, that is a bit large to replicate!

> I've got an account ready on Frontera. Are there any more specific parameter sets you use for this, or tests that I should run?

There are, but the machine is down right now for maintanence and the Texascale run next week. I'll send you a link to our config files when it comes back.

examples/dm-only/paramfile.gadget
```diff
@@ -3,6 +3,9 @@ InitCondFile = output/IC
 OutputDir = output
 OutputList = 0.02,0.1,0.15,0.2,0.33333,0.5,1.0
 
+# Mesh size (in general should be set to 2xNgrid in IC)
+Nmesh = 128
```
sbird, line 7: Please remove this the default is fine.

sbird (commented):
Thanks @jedavies - this is looking close now. I had a few comments but nothing major. One request I do have is that you share the script you used for checking the ionization fraction and ideally plotting the bubbles. It's very useful to have a quick way of checking if a change broke the feature. 

Have you run a reasonable sized (512^3) simulation using this yet? If not, this is probably a good time to start.

libgadget/petapm.c
```diff
@@ -367,6 +375,163 @@ void petapm_force(PetaPM * pm, petapm_prepare_func prepare,
...
+}
+
+/* We need a slightly different flow for reionisation, so I 
+ * will define these here instead of messing with the force functions*/
+void petapm_reion(PetaPM * pm_mass, PetaPM * pm_star, PetaPM * pm_sfr,
+        petapm_prepare_func prepare,
+        PetaPMGlobalFunctions * global_functions, //petapm_transfer_func global_transfer,
+        PetaPMFunctions * functions,
+        PetaPMParticleStruct * pstruct,
+        petapm_reion_func reion_loop,
```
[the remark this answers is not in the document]
daviesje: This function compares the filtered and inverse transformed mass, stellar mass and sfr grids to determine what is ionised and J21 in each cell. I couldn't fit this into the transfer or readout functions without some serious modification, since it is applied after the c2r, can't operate within the iterate functions and requires all three grids.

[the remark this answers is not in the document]
sbird: I see, that makes perfect sense but could you add a comment here saying that?

libgadget/cooling_uvfluc.c
```diff
@@ -238,16 +238,19 @@ static struct UVBG get_local_UVBG_from_J21(double redshift, int heiiionized, dou
...
-//placeholder function so i don't have to delete old get_local_UVBG yet
+//switch function that decides whether to use excursion set or global UV background
+/*TODO: there are a few continuity issues to consider fixing.
+ * switching to the global rates for heii ionised particles is discontinuous.
+ * if the z_reion tables provided finish after ExcursionSetZStop, particles could rapidly recombine.
+ * I'm not sure how initial heating from reionisation (see D'Aloisio et al. 2019) is handled here */
 struct UVBG get_local_UVBG(double redshift, double * Pos, const double * PosOffset, int heiiionized, double J21)
 {
-    if(All.ExcursionSetFlag && (redshift > All.ExcursionSetZStop))
+    if(All.ExcursionSetReionOn && (redshift > All.ExcursionSetZStop) && !(heiiionized))
```
sbird, line 248: I think just assume (or enforce) that All.ExcursionSetZstop < the helium reionisation redshift and so you don't need to worry about it.

libgadget/cooling_uvfluc.c
```diff
@@ -177,6 +179,85 @@ struct UVBG get_local_UVBG(double redshift, double * Pos, const double * PosOffs
...
+    uvbg.epsH0  = J21toUV.epsH0 * J21 * 1.60218e-12;  // erg s-1
+    uvbg.gJHe0  = J21toUV.gJHe0 * J21; // s-1
+    uvbg.epsHe0 = J21toUV.epsHe0 * J21 * 1.60218e-12;  // erg s-1
+
+    //TODO:look more into the qso_lightup model to see if this is correct
+    //TODO: set heating to be consistent with qso model, which activates long_mfp_heating
+    //a self-consistent approach probably needs separate stellar/quasar rates, H+ He+, He++ spectral slopes
+    if(heiiionized){
+        uvbg.gJHep  = J21toUV.gJHep * J21; // s-1
+        uvbg.epsHep = J21toUV.epsHep * J21 * 1.60218e-12;  // erg s-1
```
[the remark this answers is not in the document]
daviesje: Since this excursion set doesn't include quasars yet, I've set those rates to zero. I've also added to the conditional in get_local_UVBG to use the global rates for HeIII ionised particles. There may be some continuity issues if there is significant overlap between the end of hydrogen reionisation and the start of helium reionisation, but getting around this means I'd have to link the two models.

libgadget/cooling_uvfluc.c
daviesje: Fair enough, I'll remove the heiiionized flag and use the highest of the two redshifts to stop the excursion set.

sbird: Thanks - given your comment above it sounds like we don't really do that correctly anyway, so we may as well be clear about that

daviesje:
I believe I have adressed all the issues raised, as well as fixed a couple of bugs I hadn't noticed. I've included the scripts I used to compare ionisation topology with the mass and stellar distributions (get_xgrids.py to grid the particles and plot_grids.py to compare them). As well as the global stellar mass -> gas mass comparison to make sure the global history is roughly as expected (count_photons.py).

I am now running a 512^3 40 Mpc run on Ozstar. I'll let you know how that goes.

tools/count_photons.py
```diff
@@ -0,0 +1,86 @@
+import bigfile as bf
```
sbird, line 1: Could you please comment this a bit?

tools/plot_grids.py
```diff
@@ -0,0 +1,82 @@
+import numpy as np
+import matplotlib
```
sbird, line 2: Please comment

sbird:
It's looking pretty good now - could you resolve the conflicts and add a few comments to the python scripts? Ideally run them through pylint and resolve whatever warnings it gives that seem reasonable (it can be opinionated about names: I don't care about that, but adding docstrings is a good idea.

sbird:
Ok, I resolved the conflicts...and they were trivial :) Please do add some comments and I will do some tests myself.

sbird:
And please could you fix the tests!

libgadget/partmanager.h
```diff
@@ -14,9 +14,14 @@ struct particle_data
     double Pos[3];   /*!< particle position at its current time */
     float Mass;     /*!< particle mass */
 
+    /* particle type.  0=gas, 1=halo, 2=disk, 3=bulge, 4=stars, 5=bndry */
+    /* TODO(jdavies): I moved this out of the bitfield because i need to access it by pointer in petapm.c
+     * This could also be done by passing a struct pointer instead of void* as the petapm pstruct */
```
sbird, line 19: Could you please make this a signed int so some warnings are quiet?

sbird:
Could you please fix this warning:
zeldovich.c: In function ‘displacement_fields’:
zeldovich.c:163:5: warning: missing initializer for field ‘offset_type’ of ‘PetaPMParticleStruct’ [-Wmissing-field-initializers]
  163 |     };
      |     ^
In file included from proto.h:7,
                 from zeldovich.c:9:
../libgadget/petapm.h:88:12: note: ‘offset_type’ declared here
   88 |     size_t offset_type;
      |            ^~~~~~~~~~~
glass.c: In function ‘glass_force’:
glass.c:191:5: warning: missing initializer for field ‘offset_type’ of ‘PetaPMParticleStruct’ [-Wmissing-field-initializers]
  191 |     };
      |     ^
In file included from proto.h:7,
                 from glass.c:10:
../libgadget/petapm.h:88:12: note: ‘offset_type’ declared here
   88 |     size_t offset_type;
      |            ^~~~~~~~~~~
ar r libgenic.a .objs/power.o .objs/zeldovich.o .objs/glass.o .objs/save.o .objs/thermal.o

examples/es_test/paramfile.genic
```diff
@@ -0,0 +1,36 @@
...
+DifferentTransferFunctions = 0
+ScaleDepVelocity = 0
+
+WhichSpectrum = 2         # "1" selects Eisenstein & Hu spectrum,
+                           # "2" selects a tabulated power spectrum in
+                           # the file 'FileWithInputSpectrum'
+                           # otherwise, Efstathiou parametrization is used
+
+# filename of tabulated input in Mpc/h
+FileWithInputSpectrum = ../../powerspectrum-wmap9.txt
```
sbird, line 28: Should be ../ not ../../ I think

sbird: But could you please make this class_tk_99 and class_pk_99? and set DifferentTransferFunctions = 1?

examples/es_test/paramfile.genic
```diff
@@ -0,0 +1,36 @@
...
+
+Omega0 = 0.2814      # Total matter density  (at z=0)
+OmegaLambda = 0.7186      # Cosmological constant (at z=0)
+OmegaBaryon = 0.0464     # Baryon density        (at z=0)
+ProduceGas = 1         # 1 = Produce gas  0 = no gas, just DM.
+HubbleParam = 0.697      # Hubble paramater (may be used for power spec parameterization)
+
+Redshift = 25        # Starting redshift
+
+Sigma8 = 0.810      # power spectrum normalization
```
sbird, line 17: This can be -1 if you use the class_pk_99 files.

examples/es_test/paramfile.genic
```diff
@@ -0,0 +1,36 @@
+OutputDir = output_reion # Directory for output
+FileBase = IC              # Base-filename of output files
+
+Ngrid = 128 # Size of cubic grid on which to create particles.
+
+BoxSize = 40000   # Periodic box size of simulation
+MaxMemSizePerNode = 16000
```
sbird, line 7: Delete this line

examples/es_test/run.sh
```diff
@@ -0,0 +1,6 @@
+export OMP_NUM_THREADS=$4
```
sbird, line 1: $4?

examples/es_test/paramfile.gadget
```diff
@@ -0,0 +1,120 @@
...
+UVBGdim = 64
+ReionRBubbleMax = 20340 #kpc
+ReionRBubbleMin = 406.8
+ReionDeltaRFactor = 1.1
+ReionGammaHaloBias = 2.0
+ReionNionPhotPerBary = 8000.0
+AlphaUV = 3.0
+EscapeFraction = 1.0
+UVBGTimestep = 10.0 #time between UVBG calculations
+UVBGSaveGrids = 1 #save UVBG related grids directly from 21cmfast
```
sbird, line 117: Need to update these parameters.

sbird:
Cheers

sbird:
I think for the example you want to apply:

sbird:
diff --git a/examples/es_test/paramfile.genic b/examples/es_test/paramfile.genic
index 0b523f4d..7d222d1c 100644
--- a/examples/es_test/paramfile.genic
+++ b/examples/es_test/paramfile.genic
@@ -18,20 +18,9 @@ DifferentTransferFunctions = 1
 FileWithInputSpectrum = ../class_pk_99.dat  # filename of tabulated input spectrum (if used)
 FileWithTransferFunction = ../class_tk_99.dat  # filename of transfer functions for different species
 
-ScaleDepVelocity = 0
-
 WhichSpectrum = 2         # "1" selects Eisenstein & Hu spectrum,
                            # "2" selects a tabulated power spectrum in
                            # the file 'FileWithInputSpectrum'
                            # otherwise, Efstathiou parametrization is used
 
-# filename of tabulated input in Mpc/h
-FileWithInputSpectrum = ../../powerspectrum-wmap9.txt
-
-PrimordialIndex = 0.971      # may be used to tilt the primordial index
-
 Seed = 181170    #  seed for IC-generator
-
-UnitLength_in_cm = 3.085678e21   # defines length unit of output (in cm/h) 
-UnitMass_in_g = 1.989e43      # defines mass unit of output (in g/cm)
-UnitVelocity_in_cm_per_s = 1e5           # defines velocity unit of output (in cm/sec)

sbird:
and let the defaults do the work

sbird:
Also the following line added to paramfile.gadget:
MetalReturnOn = 0
(new model since this code was written)

sbird:
index c767059a..35ca4df4 100644
--- a/examples/es_test/paramfile.gadget
+++ b/examples/es_test/paramfile.gadget
@@ -2,7 +2,7 @@
 
 InitCondFile = output_reion/IC
 OutputDir = output_reion
-MetalCoolFile = ../../cooling_metal_UVB
+MetalCoolFile = ../cooling_metal_UVB
 OutputList = 0,09090,0.09259,0.09434,0.09615,0.09804,0.1000,0.1020,0.1042,0.1064,0.1087,0.1111,0.1136,0.1163,0.1190,0.1220,0.1250,0.1282,0.1316,0.1351,0.1389,0.1429,0.1471,0.1515,0.1563,0.1613
 EnableAggregatedIO = 1
 # CPU time -limit
@@ -115,4 +115,5 @@ EscapeFraction = 1.0
 UVBGTimestep = 10.0 #time between UVBG calculations in Myr
 ReionFilterType = 0 #Filter to use with 21CMFAST, 0 = real-space top-hat, 1 = k-space top-hat, 2 = gaussian
 RtoMFilterType = 0 #Filter to use with 21CMFAST radius to mass calculation, 0 = top-hat, 1 = gaussian
-J21CoeffFile = ../../J21_to_rates_test.txt
+J21CoeffFile = ../J21_to_rates_test.txt
+MetalReturnOn = 0

sbird:
Oh, I'm sorry. My mistake. I see you start at z=25, so in order to form stars you do need to normalise the z=99 power spectra, or generate new z=25 power spectra.

sbird:
So in fact you do need Sigma8 = 0.850 in the genic paramfile.

sbird:
Ok, this passes my smoke test by producing 0 change in output on the hydro example with ExcursionSetFlagOn = 0. I think I'm happy to merge it once you have finished doing your testing.

sbird:
@rainwoodman Do you have time to give a quick review?

sbird:
See https://github.com/daviesje/MP-Gadget/pull/1
To confirm, when I call get_xgrids I want something like:
`python ../../tools/get_xgrids.py output_reion/PART_020 output_reion/PART_020_grid --weight "0/J21" --pos "0/Position"`

Right?

sbird:
Probably not since the output has a lot of NaN :) This should be documented in the script somewhere please.

daviesje:
> See [daviesje#1](https://github.com/daviesje/MP-Gadget/pull/1)
> To confirm, when I call get_xgrids I want something like:
> `python ../../tools/get_xgrids.py output_reion/PART_020 output_reion/PART_020_grid --weight "0/J21" --pos "0/Position"`
> 
> Right?

At the moment you need to input the box size, although I'm making an update that will read in the attribute from the file header and add usage comments. I had the default from when I was working with the bluetides outputs, thanks for catching that.

sbird:
Nice, thanks. From me I am happy to merge this when you are ready.

daviesje:
Almost, I ran count_photons.py on the frontera25 run in order to check it and perhaps decide an escape fraction for the bigger runs. It seemed to suggest that we need an escape fraction of ~1 to get a similar reionisation history, which is too large. I'm checking over the way I count stars in both the python script and the model. I'm also sending an update email to everyone with the plots.

sbird:
Sounds fine - that is the research portion of this and I am willing to trust you to do it right :)

libgadget/fof.c
```diff
@@ -801,6 +801,43 @@ fof_alloc_group(const struct BaseGroup * base, const int NgroupsExt)
     return Group;
 }
 
+/* TODO: It would be a good idea to generalise this to arbitrary fof/particle properties */
+static void fof_set_escapefraction(struct FOFGroups * fof, const int NgroupsExt)
+{
+    int i = 0;
+    for(i = 0; i < PartManager->NumPart; i++){
+        if(P[i].Type == 4){
+            STARP(i).EscapeFraction = 0.;	/* will mark particles that are not in any group */
```
sbird, line 810: Add #pragma parallel for here ?

libgadget/fof.c
```diff
@@ -844,6 +881,10 @@ fof_compile_catalogue(struct FOFGroups * fof, const int NgroupsExt, double BoxSi
 
     fof_finish_group_properties(fof, BoxSize);
 
+    /* feed group property back to each particle. */
+    /* TODO: put a flag here to only run if we need it */
+    fof_set_escapefraction(fof, NgroupsExt);
```
sbird, line 886: Please do add the flag before merging :)

sbird:
Hi @daviesje I'm almost ready to merge: one thing I noticed is that the es_test example doesn't work for me. Here is the output with a threaded run (1 process, 12 threads):

[ 000027.36 ] Treewalk DENSITY iter 0: total part 2097152 max/MPI: 2097152 min/MPI: 2097152 balance: 1.
[ 000037.93 ] iteration 0, max relative change in EgyWtDensity = nan
[ 000038.05 ] Treewalk DENSITY iter 0: total part 2097152 max/MPI: 2097152 min/MPI: 2097152 balance: 1.
[ 000048.63 ] iteration 1, max relative change in EgyWtDensity = nan
[ 000048.75 ] Treewalk DENSITY iter 0: total part 2097152 max/MPI: 2097152 min/MPI: 2097152 balance: 1.

Simeon

daviesje:
> Hi @daviesje I'm almost ready to merge: one thing I noticed is that the es_test example doesn't work for me. Here is the output with a threaded run (1 process, 12 threads):
> 
> [ 000027.36 ] Treewalk DENSITY iter 0: total part 2097152 max/MPI: 2097152 min/MPI: 2097152 balance: 1. [ 000037.93 ] iteration 0, max relative change in EgyWtDensity = nan [ 000038.05 ] Treewalk DENSITY iter 0: total part 2097152 max/MPI: 2097152 min/MPI: 2097152 balance: 1. [ 000048.63 ] iteration 1, max relative change in EgyWtDensity = nan [ 000048.75 ] Treewalk DENSITY iter 0: total part 2097152 max/MPI: 2097152 min/MPI: 2097152 balance: 1.
> 
> Simeon

Sorry about that, there was a typo in the output list in that example. It should be fixed now, there was also a bug in the temperature boost which I somehow missed

libgadget/sfr_eff.c
```diff
@@ -463,9 +463,15 @@ cooling_direct(int i, const double redshift, const double a3inv, const double hu
...
-        const double meanweight = 4 / (8 - 5 * (1 - HYDROGEN_MASSFRAC));
+    if(sfr_params.HIReionTemp > 0 && uvbg.zreion >= redshift && uvbg.zreion < lastred) {
+        /* We assume singly ionised helium at the time of reionisation */
+        /* The 100% correct thing to do is to solve for the equilibrium ne based on the local UVBG
+         * then calculate the mean weight based on this. The current approach will cause
+         * a boost in reionisation temperatures proportional to the residual neutral fraction,
+         * which should be relatively small most of the time */
+        /* TODO: Make sure that not setting SPHP.Ne(i) here doesn't mess up anything between
+         * now and the next cooling call when it gets set properly */
+        const double meanweight = 4 / (8 - 6 * (1 - HYDROGEN_MASSFRAC));
```
sbird, line 474: Did you mean to make that (8 - 6 * (1 - HYDROGEN_MASSFRAC)) ?

sbird: The 5 has changed to 6

sbird:
Can you explain why changing the output list causes a NaN to come out of the initial density estimate?

libgadget/sfr_eff.c
daviesje: The mean weight should have a 6 assuming Helium singly ionises with hydrogen, and the HeII -> HeIII ionisation happens later. It's a 5 if the gas is totally ionised at the same time.

daviesje:
There was a typo in the previous list (, instead of .) which caused a syncpoint to appear at a=0. This messed with the EgyWgtDensity calculation via the timebins and dloga. I could add a check to the syncpoint generation for times before the ICs, or set the no_snapshot_until_time argument in the initialisaition to the IC time.

libgadget/sfr_eff.c
sbird: Good point!

sbird:
Ah right. A check for times before the ICs would be best I think - I think this would happen even with the master branch? It is a fiddly error to find.

daviesje:
Would it be safe to simply exclude any syncpoints provided that are outside the range [TimeIC,TimeMax]? There is an edge case where if we restart a run from some snapshot with the same output list and the timeline is reset (init.c : 77 where the snapshot time is passed as TimeIC) the integer times will be different.

There's a comment in timebinmgr.c: 156 that suggests this might be important, but I'm not sure what could be affected.
` * TimeIC and TimeMax are used to ensure restarting from snapshot obtains exactly identical integer stamps.`

sbird:
re init.c:77
If you restart from a snapshot output that isn't on the integer timeline (eg: because you ran out of time for the job) the exact number of timesteps to the next output is not necessarily the same and so you cannot do precise profile timing. That warning is quite old and not really important.

I think it's fine to just exclude syncpoints before TimeIC in general. We would just have to be careful that we still output snapshots with an increasing snapshot number on restart.

sbird:
That's a good solution, I like it.

sbird:
Ok, I can see nothing more wrong with this. I did write a few cleanups which I will send as a later PR and ask you, James, for review. Merging! Excitement!

PR #479, opened 2020-11-15
Treefix

sbird:
Fix a potential failure of the tree to build if there are multiple black holes at zero distance, by making sure these black holes merge.


sbird:
Thanks @daviesje for finding this bug!

PR #482, opened 2020-12-03
fix a bug in BH_DRAG

yueyingn:
`fac*= a`  should be outside the loop.


sbird:
Oops! Well noticed and just in time!

sbird:
Will merge when it goes green.

PR #483, opened 2020-12-04
Metal return model

sbird:
his is the stellar population metal yield model. It follows the TNG paper https://arxiv.org/abs/1703.02970 but with slightly different SNII yields which follow SIMBA.


sbird:
The results for this in test cases look ok, but could @rainwoodman and @yueyingn give it a quick look over? I may have missed something obvious.

PR #487, opened 2020-12-15
Log Mseed for blackholes and append SnapNum to BlackholeDetail file

yueyingn:
When doing the power law seeding, we might need to log the seed mass for each black hole (e.g. when calculating the faint end population we might only want to account for the BH accretes well beyond the seed mass).  We can also extract this information from BlackholeDetail history, but that's a lot of work when we have tons of BHs.

I also want to append the RestartSnapNum to BlackholeDetail file, otherwise the binary file would be huge (and sometimes be re-written) and would be hard to post process.

### DIFF
--- a/libgadget/petaio.c
+++ b/libgadget/petaio.c
@@ -860,6 +860,7 @@ SIMPLE_PROPERTY_PI(BlackholeSwallowID, SwallowID, uint64_t, 1, struct bh_particl
 SIMPLE_PROPERTY_PI(BlackholeSwallowTime, SwallowTime, float, 1, struct bh_particle_data)
 SIMPLE_PROPERTY_PI(BlackholeJumpToMinPot, JumpToMinPot, int, 1, struct bh_particle_data)
 SIMPLE_PROPERTY_PI(BlackholeMtrack, Mtrack, float, 1, struct bh_particle_data)
+SIMPLE_PROPERTY_PI(BlackholeMseed, Mseed, float, 1, struct bh_particle_data)
 
 SIMPLE_SETTER_PI(STBlackholeMinPotPos , MinPotPos[0], double, 3, struct bh_particle_data)
 static void GTBlackholeMinPotPos(int i, double * out, void * baseptr, void * smanptr) {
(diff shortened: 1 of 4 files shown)


yueyingn:
I can also post process a 5/BlackholeMseed column for the latest snapshot of the large run, therefore we can restart properly.

libgadget/petaio.c
```diff
@@ -1033,6 +1034,7 @@ void register_io_blocks(struct IOTable * IOTable, int WriteGroupID) {
     IO_REG(BlackholeMinPotPos, "f8", 3, 5, IOTable);
     IO_REG(BlackholeJumpToMinPot,   "i4", 1, 5, IOTable);
     IO_REG(BlackholeMtrack,         "f4", 1, 5, IOTable);
+    IO_REG(BlackholeMseed,         "f4", 1, 5, IOTable);
```
sbird, line 1037: Could you make this IO_REG_NONFATAL for backwards compatibility?

libgadget/run.c
```diff
@@ -574,10 +574,16 @@ open_outputfiles(int RestartSnapNum)
...
+    if(RestartSnapNum != -1) {
+        postfix = fastpm_strdup_printf("-R%03d", RestartSnapNum);
+    } else {
+        postfix = fastpm_strdup_printf("%s", "");
+    }
 
     /* all the processors write to separate files*/
     if(All.BlackHoleOn && All.WriteBlackHoleDetails){
-        buf = fastpm_strdup_printf("%s/%s/%06X", All.OutputDir,"BlackholeDetails",ThisTask);
+        buf = fastpm_strdup_printf("%s/%s%s/%06X", All.OutputDir,"BlackholeDetails",postfix,ThisTask);
```
sbird, line 586: This is a good idea.

sbird:
This looks good. I think we don't need to postprocess with the nonfatal array?

yueyingn:
I think we still need to postprocess the current latest snapshot. Otherwise the current 20000 BHs in the simulation would just have 0 Mseed in the later log?

sbird:
Sure - isn't that what we want? It marks clearly that we don't have a known seed. Or are you going to somehow deduce the seed from the BH progenitors?

PR #490, opened 2020-12-21
Compute DHSML for BHs as well as gas

sbird:
This PR uses density gradients to find the smoothing length of the black holes as well as the SPH particles and the stars. This makes density() converge much faster for BHs, at the cost of a small amount of extra memory during density().


libgadget/density.c
```diff
@@ -449,6 +446,8 @@ density_reduce(int place, TreeWalkResultDensity * remote, enum TreeWalkReduceMod
     else if(P[place].Type == 5)
     {
         TREEWALK_REDUCE(BHP(place).Density, remote->Rho);
+        int bhpi = P[place].PI + SlotsManager->info[0].size * DENSITY_GET_PRIV(tw)->DoEgyDensity;
```
rainwoodman, line 449: I feel I have seen this type of access to 'temporal/transient' variables several times. What about introducing a TemporalSlots class to capture these tw->??? variables?

```
struct TemporalSlots {
union {
  double *,
  int *,
};
int NPerType[6];
};

tw->DhsmlDensityFactor = temporal_slots_new_double(int NPerType[6], ...); 
```

and (perhaps a macro ...)
```
double * pvalue = temporal_slots_get_double(tw->DhsmlDensityFactor, type, P[place].PI);
```

The advantage is that the logic for relative indexing no longer need to explicitly do a branching per access (and thus no need to maintain the consistency).

sbird: We are certainly growing new boilerplate around this so something would be good. Could you expand on what your suggestion is?

I don't think branching is really a problem (these branches should be easily predicted), although a fairly expensive thing in the current code is the non-linear memory access inherent in the slots having a different ordering than the main particle table.

PR #492, opened 2020-12-22
Treeopts

sbird:
Optimize the threaded work splitting in the tree, removing the globally locked export queue and ensuring that small numbers of particles are more evenly distributed across threads.


sbird:
Zero change to output, speedup of about 5% on my workstation and a 512^3 simulation on Frontera.

sbird:
bit identical output, 10% faster, merging.

PR #494, opened 2020-12-24
Start the recursive treebuild from the leaf topnodes

sbird:
This should help the parallelization of the tree moment calculation by
ensuring that each task has real particles to compute and is roughly
equal in depth.


sbird:
This reduces tree moment calculation time by a factor of 5 on frontera! It is totally negligible now.

PR #500, opened 2021-01-12
Largescaling

sbird:
At Asterix scale, any loop over all particles becomes a bottleneck. This PR removes some of these loops, the initialisation of the predicted quantities and the BH injected_BH_energy loop. More coming.

### DIFF
--- a/libgadget/blackhole.c
+++ b/libgadget/blackhole.c
@@ -153,12 +153,12 @@ struct BHPriv {
     MyFloat * BH_accreted_Mass;
     MyFloat * BH_accreted_BHMass;
     MyFloat * BH_accreted_Mtrack;
-    MyFloat * Injected_BH_Energy;
 
     /* This is a temporary computed in the accretion treewalk and used
      * in the feedback treewalk*/
     MyFloat * BH_FeedbackWeightSum;
 
+    double a3inv;
     /* Counters*/
     int64_t * N_sph_swallowed;
     int64_t * N_BH_swallowed;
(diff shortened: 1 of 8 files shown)


libgadget/blackhole.c
```diff
@@ -1227,28 +1208,40 @@ blackhole_feedback_ngbiter(TreeWalkQueryBHFeedback * I,
...
+
+        const double injected_BH = I->FeedbackEnergy * mass_j * wk / I->FeedbackWeightSum;
+        /* Set a flag for star-forming particles:
+            * we want these to cool to the EEQOS via
+            * tcool rather than trelax.*/
+        if(sfreff_on_eeqos(&SPHP(other), BH_GET_PRIV(lv->tw)->a3inv)) {
+            /* We cannot atomically set a bitfield.
+             * This flag is never read in this thread loop, nor
+             * are other flags set here. So lack of atomicity is (I think) not a problem*/
+            //#pragma omp atomic write
```
sbird, line 1235: @rainwoodman  Can you confirm this is true?

sbird:
No real change in output: some very slight increase (<0.1%) in star formation from 
9367892, probably due to increased fp roundoff in the BH energy injection.

libgadget/blackhole.c
rainwoodman: To the extend of my knowledge I agree with you if we only use a single bit this way, we do not need atomicity.

rainwoodman:
I thought we already rebuild the force tree very step? That's O[n log(n)] -- it is difficult to imagine a few O[n] loops becoming bottleneck.

The word 'on-demand' scares me a bit (implies more states), does it matter in this context?

libgadget/blackhole.c
sbird: Thanks

sbird:
We do already rebuild the forcetree every step, and we need to deal with that too eventually. All these things are effectively constant time overhead for each timestep, even the short ones, so anything we can do to reduce it decreases time to solution. 

For Asterix scale a short timestep looks like:
Total: 1.9 s
Treebuild: 0.9 s
Drift: 0.3
Exchange/Togo: 0.2 s
Active: 0.2 s
HydroInit: 0.1 s
Density Init: 0.2 s

(where they do not add up exactly because of rounding errors)

Removing the per-timestep treebuild is the big one, but it is hard. We can still get halfway by removing these other steps.

It does imply more state: there is an extra implicit state of "EntVarPred == 0" which means "Predicted quantities not computed". I think this is ok because entropy is bounded to never be zero.

sbird:
(Also O(n log n) is pretty close to O(n) practically! Especially as n is the number of particles per topnode which is pretty constant).

PR #504, opened 2021-01-16
Metalsknn

sbird:
As suggested by @rainwoodman this implements k nearest neighbours for the metal return. This is currently only for seeding the density() loop (which works at reducing the wait tim). I could also simplify the code drastically by returning metals to the k nearest neighbours, but not sure if I want to yet.

### DIFF
--- a/libgadget/metal_return.c
+++ b/libgadget/metal_return.c
@@ -1154,15 +1154,14 @@ stellar_knn_reduce(int place, TreeWalkResultStellarKNN * remote, enum TreeWalkRe
         /* Walk through the new list until we find someone not in the original list.
          * Add them, then continue.*/
         int i;
+        int * nnb = &STELLAR_KNN_GET_PRIV(tw)->nnb[pi];
         for(i = 0; i < remote->nnb; i++) {
-            /* Check whether the list changed: if the furthest particle is still the same we have no merging to do.*/
-            if(remote->nnb == STELLAR_KNN_GET_PRIV(tw)->nnb[pi]
-                    && remote->neighbours[remote->nnb-1].part == neigh[remote->nnb-1].part)
+            /* If all future particles are further away, we have nothing to do.
+             * This is unusual: we start with a local list*/
+            if(*nnb == NUMNB && remote->neighbours[i].dist2 >= neigh[remote->nnb-1].dist2)
                 break;
             const struct Neighbour nn = remote->neighbours[i];
-            if(nn.part == neigh[i].part)
-                continue;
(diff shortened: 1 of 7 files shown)


libgadget/metal_return.c
```diff
@@ -1021,3 +1021,435 @@ stellar_density(const ActiveParticles * act, MyFloat * StellarAges, MyFloat * Ma
...
+
+
+/* Here comes the neighbour finding. We want to return metals to the nearest N neighbours to the star.
+ * As a seed for the Hsml we find the kth nearest neighbour algorithm using a heap.
+ * This is always an underestimate of hsml, but should be enough for a gradient lookup to be efficient.*/
+#define NUMNB 50
+
+struct Neighbour
+{
+    int part;
```
rainwoodman, line 1033: This is undefined if the neighbour is from a remote rank?

libgadget/metal_return.c
```diff
@@ -1021,3 +1021,435 @@ stellar_density(const ActiveParticles * act, MyFloat * StellarAges, MyFloat * Ma
...
+        /* Walk through the new list until we find someone not in the original list.
+         * Add them, then continue.*/
+        int i;
+        for(i = 0; i < remote->nnb; i++) {
+            /* Check whether the list changed: if the furthest particle is still the same we have no merging to do.*/
+            if(remote->nnb == STELLAR_KNN_GET_PRIV(tw)->nnb[pi]
+                    && remote->neighbours[remote->nnb-1].part == neigh[remote->nnb-1].part)
+                break;
+            const struct Neighbour nn = remote->neighbours[i];
+            if(nn.part == neigh[i].part)
```
rainwoodman, line 1163: I think this comparison requires part to be globally unique (e.g. particle ID).

libgadget/metal_return.c
```diff
@@ -1021,3 +1021,435 @@ stellar_density(const ActiveParticles * act, MyFloat * StellarAges, MyFloat * Ma
...
+    /* now test against the minimal sphere enclosing everything */
+    dist += FACT1 * current->len;
+
+    if(r2 > dist * dist) {
+        return 0;
+    }
+    return 1;
+}
+/*****
+ * This is the internal code that looks for particles in the ngb tree from
```
rainwoodman, line 1237: Is this section updated for the knn search?

libgadget/metal_return.c
sbird: I think in practice I don't need this part variable at all - I was keeping hold of it so that I could do metal return by just weighting the k nearest neighbours, but I decided I didn't want to monkey with the big run like that.

libgadget/metal_return.c
sbird: Indeed! Well noticed, Yu!

libgadget/metal_return.c
sbird: No, let me fix it. I copy-pasted more code than I actually needed.

PR #506, opened 2021-01-20
Fixmetals2

sbird:
Another possible solution


sbird:
Will merge these as clear bugfixes when travis goes green.

PR #507, opened 2021-01-25
Memprint

sbird:
This PR fixes a crash when the Frontera nodes are fullly loaded and we use 28 threads. 3525ab4 is the actual bugfix. The others are pre-emptive extension of counter variables to int64_t in the hope of preventing future bugs.


sbird:
Zero change in MPI, within normal change for thread.

PR #508, opened 2021-01-25
allow to softening the effective equation of state

yueyingn:
Hi Simeon, I will have a run testing softening the equation of state. Can you check if the modification like this make sense?


yueyingn:
Sorry this PR messed up with some other commits. I will make a clean PR after tested.

sbird:
Thanks @yueyingn - that looks right to me (once the int->int64 is fixed). I would hope that this makes very little difference to star formation but fingers crossed it speeds up the code!

sbird:
We decided not to use this - shall we merge it anyway? It does no harm and is a not unreasonable feature to have.

yueyingn:
How about just drop it? It does no harm but also no use and makes the code longer. I don't expect we are going to revisit it in the near future. Up to you though!

sbird:
It's your code so your decision and I will drop it :)

PR #511, opened 2021-01-30
Mergedstardens

sbird:
This PR computes the stellar density with the other density estimates. This should improve balancing, as it can overlap with the SPH density computations.


sbird:
Bit-identical up to f3e0361 small changes thereafter (presumably because the exact end of the density iteration changes).

sbird:
Small time improvements on a small 550^3 box, but that may be because the balance is pretty good there anyway.

sbird:
It seems that this doesn't really do anything - it makes the balance better but at the cost of an extra treewalk, and they more or less cancel. It's less code, but it is nice to have the metal code all in one place, and I could reduce duplication in other ways if I wanted to.

PR #512, opened 2021-02-01
Mergedstardens

sbird:
Slimmed down PR about merging stellar density with regular density


sbird:
Dropping this because I decided I prefer to have the stellar code self-contained.

PR #520, opened 2021-02-17
Do multiple density iterations at once

sbird:
This is the latest attempt to improve the balance of the stellar density (metal return) code. If this works we could perhaps use it for the gas density as well. The idea is that to use that computing the neighbour number, NumNgb, for any smoothing length h < H (where H is the current particle Hsml) is essentially free, because the treewalk is dominated by the memory latency involved in fetching the nodes and particles. So we can compute NumNgb(h_i) for multiple h_i values in parallel and use this information to efficiently bound the desired SPH smoothing length. This dramatically reduces the maximum number of density iterations, from about 10 to 3. This should also improve the balance as the final iterations usually have very few particles active.

A side effect is that we can stop a treewalk and reduce H if h_i > N from a partial treewalk, because NumNgb is monotonic during the walk. This should avoid the very slow treewalks we sometimes see when one particle has acquired an overly large h value.

Open questions:
- What is the optimal number of simultaneous evaluations?
- I removed the Dhsmldensity (which finds dNumNgb(h)/dh by kernel differentiation) in cd3f82e as it didn't really seem to do anything in my little hydro example test. Is this conclusion correct in a bigger box?
- Is the current choice of the h_i array optimal? At the moment the first iteration uses the largest h_i > h, and the others <= h. Later iterations linearly space h_i between Left and Right, but it might be better to weight this spacing by the NumNgb value so that, eg, if NumNgb(Left) - DesNumNgb = -5 and NumNgb(Right) - DesNumNgb = 30 we have finer spacing near Left than Right. 
- (Once we merge this) What does this look like if implemented for the gas density code and the SFR_WIND code?

and of course more testing is needed!

Currently I tested on the hydro example only. There was a speedup of about 20% in the stellar density code, but none of the pathological behaviour seen in bigger runs shows up on my single-node workstation, so I would hope improvements are larger in the bigger boxes.

### DIFF
--- a/libgadget/fof.c
+++ b/libgadget/fof.c
@@ -1120,6 +1120,9 @@ fof_secondary_ngbiter(TreeWalkQueryFOF * I,
         O->MinID = HaloLabel[other].MinID;
         O->MinIDTask = HaloLabel[other].MinIDTask;
     }
+    /* No need to search nodes at a greater distance
+     * now that we have a neighbour.*/
+    iter->base.Hsml = iter->base.r;
 }
 
 static void
@@ -1155,7 +1158,7 @@ static void fof_label_secondary(ForceTree * tree)
 
     TreeWalk tw[1] = {{0}};
     tw->ev_label = "FOF_FIND_NEAREST";
-    tw->visit = (TreeWalkVisitFunction) knn_visit;
+    tw->visit = treewalk_visit_nolist_ngbiter;
     tw->ngbiter = (TreeWalkNgbIterFunction) fof_secondary_ngbiter;
     tw->ngbiter_type_elsize = sizeof(TreeWalkNgbIterFOF);
     tw->haswork = fof_secondary_haswork;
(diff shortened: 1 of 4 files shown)


sbird:
I replaced the DhsmlDensity with a numerical extrapolation.

sbird:
Yueying confirms NHSML = 10 is optimal!

sbird:
Changed h_i to be spaced in volume as h_i^3 instead of h_i.

sbird:
Not entirely sure what the minimum Left should be on the first iteration (commit 47cfcc8), but it probably doesn't matter that much to total runtime. Lower is better in that we can exit early from more wildly too large treewalks, but higher means the initial bracket will be tighter. There is some tension between working well for rare bad cases (which suggests low initial Left) and working well for the common case where the initial guess is pretty good. For stars the bad cases are sufficiently non-rare that it makes sense to optimize for them, but I expect that for gas it will be the other way around.

sbird:
With this the time in stellar_density in asterix is 1.5%!

PR #522, opened 2021-02-23
Windmulti

yueyingn:
Applying multi-evaluation of DMradius in wind ngb iter. This PR reduces the number of iterations, but on small test runs it does not significantly save the total time usage.


libgadget/winds.c
```diff
@@ -396,17 +476,28 @@ sfr_wind_weight_ngbiter(TreeWalkQueryWind * I,
...
+                    O->V1sum[i][d] += vel;
+                    O->V2sum[i] += vel * vel;
+                }
+            }
+        }
+    }
+    
+    for(i = 0; i<NWINDHSML; i++){
+        if(O->Ngb[i] > NUMDMNGB){
+            O->maxcmpte = i+1;
```
sbird, line 499: For this to be effective at reducing computation time, you would need to do
            iter->base.Hsml = I->Hsml[i];
and use 
tw->visit = treewalk_visit_nolist_ngbiter;
for the treewalk, as in metal_return.c
The reason is that the standard ngbiter function gathers all neighbours in the Ngblist before evaluating any of them, whereas the modified one evaluates them as it finds them.

libgadget/winds.c
```diff
@@ -283,69 +288,140 @@ winds_evolve(int i, double a3inv, double hubble)
...
 
+static inline double
+effdmradius(int place, int i, TreeWalk * tw)
+{
+    struct winddata * Windd = WIND_GET_PRIV(tw)->Winddata;
+    double left = WINDP(place, Windd).Left;
+    double right = WINDP(place, Windd).Right;
+    /*The asymmetry is because it is free to compute extra densities for h < Hsml, but not for h > Hsml*/
+    if (right > 0.99*tw->tree->BoxSize){
+        right = WINDP(place, Windd).DMRadius * ((1.+NWINDHSML)/NWINDHSML);
```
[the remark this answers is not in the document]
yueyingn: I checked that the difference in initializing left and right does not make noticeable difference. I switched back to 0.1*hsml and 1.0*hsml anyway.

libgadget/winds.c
yueyingn: Thanks, I've fixed that part. The cpu time performance is similar as before. Maybe the new star is too little in my small test run to exhibit the effect.

sbird: Sounds very likely. Thanks though!

libgadget/winds.c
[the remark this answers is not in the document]
sbird: Thanks for confirming - I thought you might have done it for a Really Important Reason and am pleased that it doesn't seem to matter.

libgadget/treewalk.c
```diff
@@ -1342,3 +1342,72 @@ treewalk_do_hsml_loop(TreeWalk * tw, int * queue, int64_t queuesize, int update_
     ta_free(tw->NPRedo);
     ta_free(tw->NPLeft);
 }
+
+
+/* find the closest index from radius and numNgb list, update left and right bound, return new hsml */
+double
+ngb_narrow_down(double *right, double *left, const double *radius, const double *numNgb, int maxcmpt, int desnumngb, int *closeidx, double BoxSize)
```
sbird, line 1349: Nice. Good name.

sbird:
This looks really nice.  How much does it affect the output of a small (hydro) simulation? I'm happy to merge it once you are happy with the testing results.

yueyingn:
I did a 256^3/15Mpc test run down to z=3, commit-0cd3d33 looks good to me (saved 8% time in wind.c and produced the same GSMF as for the sanity check). 
I am currently working on density.c, still debugging. Maybe saving that for a separate PR?

sbird:
Yup, make density a separate PR. Let's merge this guy! One other thing I look at is the TotNumPart in the snapshots: is the total number of stars the same? Usually it changes by a small amount from run to run due to threading (ie, because Hsml converges slightly differently depending on what order the tree was built in), but it should be within 1%.

yueyingn:
Yes, I've checked that the difference of TotNumPart is within 1% indeed.

sbird:
Then let's merge it!

PR #523, opened 2021-02-26
Fdm

yueyingn:
Implementation of FDM dynamic by calculating the quantum pressure (potential) acceleration using SPH scheme following the work of https://arxiv.org/abs/1801.08144 (Eq4 and Eq.23 - Eq.26). 
This can be extended to incorporate sidm in the future.

### DIFF
--- a/libgadget/fdm.c
+++ b/libgadget/fdm.c
@@ -10,6 +10,7 @@
 #include <gsl/gsl_errno.h>
 #include <omp.h>
 
+#include "allvars.h"
 #include "physconst.h"
 #include "walltime.h"
 #include "slotsmanager.h"
@@ -128,6 +129,8 @@ void dm_density_check_neighbours (int i, TreeWalk * tw)
         
         DhsmlDensity[pi] *= P[i].Hsml / (NUMDIMS * (FDMP(i).Density));
         DhsmlDensity[pi] = 1 / (1 + DhsmlDensity[pi]);
+        /* We will also use this in quantum pressure calculation*/
+        FDMP(i).DhsmlDensityFactor = DhsmlDensity[pi];
 
         /* If we need more neighbours, move the lower bound up. If we need fewer, move the upper bound down.*/
         if(NumNgb[pi] < desnumngb) {
@@ -187,7 +190,7 @@ dm_density_ngbiter(
         iter->kernel_volume = density_kernel_volume(&iter->kernel);
 
         iter->base.Hsml = h;
-        iter->base.mask = 1+2; /* gas+dm ? */
+        iter->base.mask = 1+2; /* dm-only ? */
         iter->base.symmetric = NGB_TREEFIND_ASYMMETRIC;
         return;
     }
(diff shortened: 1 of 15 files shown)


libgadget/run.c
```diff
@@ -271,6 +274,13 @@ run(int RestartSnapNum)
             /* Scratch data cannot be used checkpoint because FOF does an exchange.*/
             slots_free_sph_pred_data(&sph_predicted);
         }
+        
+        if(All.FdmOn)
+        {
+            dm_density(&Act,&Tree);
+            force_update_dm_hmax(Act.ActiveParticle, Act.NumActiveParticle, &Tree, ddecomp);
```
yueyingn, line 281: Actually I am not sure whether I need to do this force_update_dm_hmax

libgadget/fdm.c
```diff
@@ -0,0 +1,585 @@
...
+    if(P[other].Type != 1 || r2 <=0 || !(r2 < iter->kernel_i.HH))
+        return;
+    
+    const double u = r * iter->kernel_i.Hinv;
+    const double dwk = density_kernel_dwk(&iter->kernel_i, u);
+    const double ddwk = density_kernel_ddwk(&iter->kernel_i, u);
+    const double mass_j = P[other].Mass;
+    
+    double fac = (FDMP(other).Density - I->Density)/sqrt(I->Density*FDMP(other).Density);
+    /* \Laplacian(W(r,h)) = 2*dwk/r + ddwk */
```
yueyingn, line 407: Check that I did the laplacian correct:
q = r/h; d/dr = (1/h)d/dq
=> d/dx(W(r,h)) = dr/dx * (1/h)dW/dq = x/r * dwk
=> d^2/dx^2(W(r,h)) = d/dx(x/r)*dwk + (x^2/r^2)*(1/h^2)d^2W/dq^2 = (1/r-x^2/r^3)*dwk + (x^2/r^2)*ddwk

libgadget/fdm.c
```diff
@@ -0,0 +1,585 @@
...
+    int d;
+    for(d = 0; d < 3; d++){
+        gradsq += pow(FDMP(other).GradDensity[d],2);
+    }
+    double fac = FDMP(other).LapDensity/(2*rho_j) - gradsq/(4*rho_j*rho_j);
+    fac *= prefac*(mass_j/rho_j)*FDMP(other).DhsmlDensityFactor;
+    /* Convention of QP_acc unit: same as Grav_Accel, that [Acc_code] = [Acc_physical]*a*a 
+     * As QP is the third order derivative of distance, so [Acc_code] = (a3inv*QP)*a*a = QP/a
+     */
+    fac /= (All.cf.a);
```
yueyingn, line 509: Need double check here, did I get the a_scale factor correct?
I am a bit confused why https://arxiv.org/pdf/2101.01828.pdf Eq.13 following https://arxiv.org/pdf/1911.05746.pdf Eq.9 has a^-2 (instead of a^-3) before the QP term.

libgadget/run.c
yueyingn: If not, I will concatenate the three QP-treewalks together into one function

sbird: You only need force_update_dm_hmax for TREEWALK_SYMMETRIC, and all the treewalks as written are TREEWALK_ASYMMETRIC . But are you sure they should be? The hydro pressure is the classic example of TREEWALK_SYMMETRIC. I don't remember from the FDM papers whether these equations need to be symmetric.

sbird: If Nori 2018 doesn't do it perhaps we shouldn't either indeed. But I'm not sure! I haven't read these papers for a while.

libgadget/init.c
```diff
@@ -210,6 +216,12 @@ inttime_t init(int RestartSnapNum, DomainDecomp * ddecomp)
             if(P[i].Hsml == 0)
                 P[i].Hsml = 0.1 * All.MeanSeparation[0];
         }
+        
+        if(All.FdmOn && RestartSnapNum == -1 && P[i].Type==1)
```
sbird, line 220: Why don't you make the FDM type 3? That way we can run simulations with both CDM and FDM, which is a scientifically interesting case.

yueyingn: Hah, I was waiting for you to say that. Will do it.

sbird: I am predictable :)

libgadget/fdm.c
yueyingn: This should be correct. On idealized test of 1D hyperbolic field, the second order derivative matches the analytical result well. 
hyperbolic1D

libgadget/fdm.c
yueyingn: With an extra 1/a factor for QP, the effect on suppression of ps matches Nori's result surprisingly well from z=49 to z=3:
ps-compare

PR #526, opened 2021-03-04
Feedbackopt

sbird:
A few fairly trivial optimizations to the feedback models.

In particular avoid an ugly loop over all particles in the wind code. We could probably do a similar trick for the swallowing in blackhole, but it is more complicated because you need to also add mass to the BH doing the swallowing (maybe on another processor) and I don't feel like doing the work right now.

### DIFF
--- a/libgadget/blackhole.c
+++ b/libgadget/blackhole.c
@@ -512,6 +512,22 @@ blackhole(const ActiveParticles * act, ForceTree * tree, FILE * FdBlackHoles, FI
 
     priv->a3inv = 1./(All.Time * All.Time * All.Time);
 
+    /* Build the queue once, since it is really 'all black holes' and similar for all treewalks*/
+    treewalk_build_queue(tw_dynfric, act->ActiveParticle, act->NumActiveParticle, 0);
+    /* Now we have a BH queue and we can re-use it*/
+    int * ActiveBlackHoles = tw_dynfric->WorkSet;
+    int64_t NumActiveBlackHoles = tw_dynfric->WorkSetSize;
+    /* If this queue is empty, nothing to do.*/
+    MPI_Allreduce(&NumActiveBlackHoles, &totbh, 1, MPI_INT64, MPI_SUM, MPI_COMM_WORLD);
+    if(totbh == 0) {
+        myfree(ActiveBlackHoles);
+        return;
+    }
+
+    /* We can re-use the current queue for these treewalks*/
+    tw_accretion->haswork = NULL;
+    tw_dynfric->haswork = NULL;
+
     /*************************************************************************/
     /*  Dynamical Friction Treewalk */
 
@@ -522,7 +538,7 @@ blackhole(const ActiveParticles * act, ForceTree * tree, FILE * FdBlackHoles, FI
     priv->BH_SurroundingDensity = mymalloc("BH_SurroundingDensity", SlotsManager->info[5].size * sizeof(priv->BH_SurroundingDensity));
     /* guard treewalk */
     if (blackhole_params.BH_DynFrictionMethod > 0)
-        treewalk_run(tw_dynfric, act->ActiveParticle, act->NumActiveParticle);
+        treewalk_run(tw_dynfric, ActiveBlackHoles, NumActiveBlackHoles);
 
     /*************************************************************************/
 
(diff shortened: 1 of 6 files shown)


yueyingn:
I tested this branch with a 256^3/15Mpc run, it fails with the following

[ 005567.11 ] Begin Step 44471, Time: 0.235789 (4bcdc0), Redshift: 3.24108, Nf = 00000000000066, Systemstep: 1.60567e-06, Dloga: 6.8098e-06, status:
[ 005567.11 ] TotNumPart: 0000033914806 SPH 0000016764461 BH 0000000771 STAR 0000000372358
[ 005567.11 ] Occupied:            0            1            2            3            4            5 dt
[ 005567.11 ]    bin=14     13796785     13781702            0            0          291            0 0.00348662
[ 005567.11 ]    bin=13      1186756      1978771            0            0        10559            0 0.00174331
[ 005567.11 ]    bin=12       821934       873662            0            0        98236           72 0.000871654
[ 005567.11 ]    bin=11       612498       133324            0            0       171526          429 0.000435827
[ 005567.11 ]    bin=10       253183         8636            0            0        61474          231 0.000217914
[ 005567.11 ]    bin= 9        73464         1121            0            0        30255            0 0.000108957
[ 005567.11 ]    bin= 8        18240            0            0            0           17           37 5.44784e-05
[ 005567.11 ]    bin= 7         1535            0            0            0            0            2 2.72392e-05
[ 005567.11 ]  X bin= 6           65            0            0            0            0            0 1.36196e-05
[ 005567.11 ]  X bin= 5            1            0            0            0            0            0 6.8098e-06
[ 005567.11 ]                -----------------------------------
[ 005567.11 ] Total:              66            0            0            0            0            0  Sum:            66
[ 005567.11 ] Tree construction.  (presently allocated=732.68 MB)
[ 005567.11 ] Allocated 243.286 MByte for 1894693 tree nodes. firstnode 3145728. (presently allocated 975.977 MB)
[ 005567.13 ] Tree constructed (moments: 1). First node 3145728, number of nodes 1543015, first pseudo 5040421. NTopLeaves 1576
[ 005567.15 ] ERROR: Task 9: Encountered zero mass particle during density; We haven't implemented tracer particles and this shall not happen
[ 005567.15 ] ERROR: Task 9: Encountered zero mass particle during density; We haven't implemented tracer particles and this shall not happen
[ 005567.15 ] ERROR: Task 9: Encountered zero mass particle during density; We haven't implemented tracer particles and this shall not happen
[ 005567.15 ] ERROR: Task 9: Encountered zero mass particle during density; We haven't implemented tracer particles and this shall not happen
[ 005567.15 ] ERROR: Task 9: Encountered zero mass particle during density; We haven't implemented tracer particles and this shall not happen
[ 005567.15 ] ERROR: Task 9: [ 005567.15 ] ERROR: Task 9: Encountered zero mass particle during density; We haven't implemented tracer particles and this shall not happen
Encountered zero mass particle during density; We haven't implemented tracer particles and this shall not happen
[ 005567.15 ] ERROR: Task 9: Encountered zero mass particle during density; We haven't implemented tracer particles and this shall not happen
[ 005567.13 ] Treewalk DENSITY iter 0: total part 66 max/MPI: 53 min/MPI: 0 balance: 12.8483.
OMP: Warning #190: Forking a process while a parallel region is active is potentially unsafe.
Abort(12) on node 9 (rank 9 in comm 0): application called MPI_Abort(MPI_COMM_WORLD, 12) - process 9
TACC:  MPI job exited with code: 12
TACC:  Shutdown complete. Exiting.

sbird:
That's odd - the only way of getting a zero mass particle is by BH swallowing and that should be unchanged here. Could you send the log from the 2 timesteps before?

sbird:
I will also make that error message more informative

yueyingn:
Oddly, the error is not reproducible. Maybe the previous job hit a very random case. (log in /scratch1/06431/yueyingn/testMPGadget3/210301-multiev/hydro-pr526/slurm-2649255.out)

The run with this PR seems to have more timesteps than the runs before. I am redoing the test with master branch for a fair comparison.

sbird:
Non-reproducible probably means it is a threading error or a memory error. The last thing to happen was:
[ 005567.06 ] Accretion done: 1 gas particles swallowed, 0 BH particles swallowed

which suggests that a gas particle is being swallowed and then not properly marked as garbage. That should be impossible?

sbird:
It might be a latent bug that this PR just exposes because it reduces the overhead of the blackhole treewalks and thus makes whatever race condition it is easier to hit.

sbird:
In which case the way forward is probably to make them yet more simplified and efficient in the hope that this will either fix the bug by accident or make the race constant :)

yueyingn:
This PR indeed save the time usage in BH and wind compared with master branch and gives consistent result. 

The current master branch indeed has more timesteps compared with earlier commit 9090cd7. We now have 28k steps from z=4 to z=3, previously it was 20k steps. I don't quite understand why. (The time usage doesn't differ by a lot though.) But it is a separate issue. I am happy to have this PR merged.

sbird:
That kind of thing can happen pretty easily when you mess with the density code, especially the wind code. Because the number of neighbours is not determined exactly but only up to +- 2, small changes in the iteration can affect the final Hsml number. This can then move particles up or down one timestep and you get 28k instead of 20k in one test, or 15k instead of 20k in another.

sbird:
We could probably tighten the bounds for density finding now it is fast :)

sbird:
Ok let's merge it! If there is a race the big run will find it soon enough!

PR #527, opened 2021-03-12
Driftopt

sbird:
Some minor improvements to the drift code. The final commit changes output marginally because the floating point routines called are slightly different.


sbird:
@yueyingn Appreciate your comments!

PR #528, opened 2021-03-17
Driftdensity

sbird:
This PR improves the estimation of h and rho for inactive particles to the drifting formula used in Gadget-4. 

We currently use the formula from Gadget-2, which is pathologically wrong for really deep timesteps because the exp() runs away. For this reason the formula is disabled in most places and we use the Density at the last active time for the particle. This is the same equation implemented, only a more accurate solution to it. It only affects the hydro (although it would affect other symmetric tree walks if we had any) and should improve force estimation moderately.

### DIFF
--- a/libgadget/drift.c
+++ b/libgadget/drift.c
@@ -70,21 +70,12 @@ void real_drift_particle(struct particle_data * pp, struct slots_manager_type * 
         /* Only does something when Pressure-entropy is enabled*/
         SPH[pi].EgyWtDensity *= densdriftfac;
 
-        //      pp->Hsml *= exp(0.333333333333 * SPHP(i).DivVel * ddrift);
-        //---This was added
-        double fac = cbrt(densdriftfac);
-        if(fac < 1./1.25)
-            fac = 1./1.25;
-
-        inttime_t ti_step = dti_from_timebin(pp->TimeBin);
-
-        /* During deep timestep hierarchies the
-         * expansion factor may get out of control,
-         * so don't let it do that.*/
-        if(ti_step <= 8*(dti))
-            pp->Hsml /= fac;
-        /* Cap the Hsml: if DivVel is large for a particle with a long timestep
-         * (most likely a wind particle) Hsml can very rarely run away*/
+        /* DtHsml is 1/3 DivVel * Hsml evaluated at the last active timestep for this particle.
+         * This predicts Hsml during the current timestep in the way used in Gadget-4, more accurate
+         * than the Gadget-2 prediction which could run away in deep timesteps. */
+        pp->Hsml += pp->DtHsml * ddrift;
+        /* Cap the Hsml just in case: if DivVel is large for a particle with a long timestep
+         * at one point Hsml could rarely run away.*/
         const double Maxhsml = BoxSize /2.;
         if(pp->Hsml > Maxhsml)
             pp->Hsml = Maxhsml;
--- a/libgadget/drift.c
+++ b/libgadget/drift.c
@@ -65,6 +65,9 @@ void real_drift_particle(struct particle_data * pp, struct slots_manager_type * 
          * This predicts Hsml during the current timestep in the way used in Gadget-4, more accurate
          * than the Gadget-2 prediction which could run away in deep timesteps. */
         pp->Hsml += pp->DtHsml * ddrift;
+        if(pp->Hsml <= 0)
+            endrun(5, "Part id %ld has bad Hsml %g with DtHsml %g vel %g %g %g\n",
+                   pp->ID, pp->Hsml, pp->DtHsml, pp->Vel[0], pp->Vel[1], pp->Vel[2]);
         /* Cap the Hsml just in case: if DivVel is large for a particle with a long timestep
          * at one point Hsml could rarely run away.*/
         const double Maxhsml = BoxSize /2.;
--- a/libgadget/drift.c
+++ b/libgadget/drift.c
@@ -61,15 +61,6 @@ void real_drift_particle(struct particle_data * pp, struct slots_manager_type * 
     }
     else if(pp->Type == 0)
     {
-        struct sph_particle_data * SPH = (struct sph_particle_data *) sman->info[0].ptr;
-        int pi = pp->PI;
-        /* This accounts for adiabatic density changes,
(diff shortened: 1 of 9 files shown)


sbird:
Note that most uses of Density (eg in the star formation model) are only for active particles and thus are not affected. It will somewhat change the computation of the sound speed in inactive neighbour particles and thus the timestepping.

sbird:
Ok, this seems to make remarkably little difference to my test box.

sbird:
Looks like the restricted timestep for divvel affects only a very few particles and none in the shortest timebin on frontera25. For reference, here are the timebin stats at z=3 from frontera25:
 002038.93 ] PM timebin: 20000 dloga = 0.00209645  Max = (0.1) Number of Timesteps: Accel: 871078 Soundspeed: 3107201 DivVel: 13793 Accretion 0 Neighbour: 1643
[ 002039.26 ] PM timebin: 20000 dloga = 0.00209645  Max = (0.1) Number of Timesteps: Accel: 0 Soundspeed: 397 DivVel: 0 Accretion 0 Neighbour: 1
[ 002039.64 ] PM timebin: 20000 dloga = 0.00209645  Max = (0.1) Number of Timesteps: Accel: 2 Soundspeed: 2042 DivVel: 4 Accretion 0 Neighbour: 3
[ 002039.95 ] PM timebin: 20000 dloga = 0.00209645  Max = (0.1) Number of Timesteps: Accel: 0 Soundspeed: 421 DivVel: 0 Accretion 0 Neighbour: 0
[ 002040.33 ] PM timebin: 20000 dloga = 0.00209645  Max = (0.1) Number of Timesteps: Accel: 17 Soundspeed: 17555 DivVel: 39 Accretion 0 Neighbour: 3
[ 002040.64 ] PM timebin: 20000 dloga = 0.00209645  Max = (0.1) Number of Timesteps: Accel: 0 Soundspeed: 442 DivVel: 1 Accretion 0 Neighbour: 0
[ 002040.98 ] PM timebin: 20000 dloga = 0.00209645  Max = (0.1) Number of Timesteps: Accel: 0 Soundspeed: 2039 DivVel: 5 Accretion 0 Neighbour: 1
[ 002041.29 ] PM timebin: 20000 dloga = 0.00209645  Max = (0.1) Number of Timesteps: Accel: 0 Soundspeed: 415 DivVel: 0 Accretion 0 Neighbour: 0

So it is about 0.5% of particles affected, max (less if including longer timebins than are here). Enough to make the time integration a bit more accurate but not enough to slow things down. I'm going to see if it affects the GSMF at all and if not merge this since it seems like a useful improvement.

sbird:
Almost no difference to the GSMF at all.

libgadget/hydra.c
```diff
@@ -214,9 +217,12 @@ hydro_force(const ActiveParticles * act, int WindOn, const double hubble, const
     {
         priv->gravkicks[i] = get_exact_gravkick_factor(CP, times.Ti_kick[i], times.Ti_Current);
         priv->hydrokicks[i] = get_exact_hydrokick_factor(CP, times.Ti_kick[i], times.Ti_Current);
+        /* For density: last active drift time is Ti_kick - 1/2 timestep as the kick time is half a timestep ahead.
+         * For active particles no density drift is needed.*/
+        if(!is_timebin_active(i, times.Ti_Current))
+            priv->drifts[i] = get_exact_drift_factor(CP, times.Ti_kick[i] - dti_from_timebin(i)/2, times.Ti_Current);
```
rainwoodman, line 223: So the values in these kicks and drifts table are to predict the quantity at Ti_Current for a particle at time bin i?

Should we track times.Ti_drift[i] as times.Ti_kick[i] - dti_from_timebin(i) / 2 ?

libgadget/hydra.c
```diff
@@ -374,10 +392,15 @@ hydro_ngbiter(
     if(Pressure_j == 0) {
         Pressure_j = PressurePred(P[other].PI, EntVarPred);
```
rainwoodman, line 393: Isn't it strange that PressurePred doesn't depend on Ti_Current?

libgadget/timestep.c
```diff
@@ -188,32 +200,45 @@ find_timesteps(const ActiveParticles * act, DriftKickTimes * times)
...
             continue;
 
+        enum TimeStepType titype;
         inttime_t dti;
         if(TimestepParams.ForceEqualTimesteps) {
             dti = dti_min;
         } else {
-            dti = get_timestep_ti(i, dti_max, times->Ti_Current);
+            dti = get_timestep_ti(i, dti_max, times->Ti_Current, &titype);
+            if(titype == TI_ACCEL)
```
rainwoodman, line 232: switch () {

case:

default:
raise(...)
}

libgadget/drift.c
```diff
@@ -61,30 +61,15 @@ void real_drift_particle(struct particle_data * pp, struct slots_manager_type *
...
-        inttime_t ti_step = dti_from_timebin(pp->TimeBin);
-
-        /* During deep timestep hierarchies the
-         * expansion factor may get out of control,
-         * so don't let it do that.*/
-        if(ti_step <= 8*(dti))
-            pp->Hsml /= fac;
-        /* Cap the Hsml: if DivVel is large for a particle with a long timestep
-         * (most likely a wind particle) Hsml can very rarely run away*/
+        /* DtHsml is 1/3 DivVel * Hsml evaluated at the last active timestep for this particle.
```
rainwoodman, line 64: Perhaps move this comment to the struct where DtHsml is defined.

libgadget/hydra.c
sbird: A good idea. Perhaps Ti_lastactivedrift[i]?

libgadget/hydra.c
sbird: EntVarPred depends on Ti_Current...

sbird: But you're right! It should depend on the predicted density.

libgadget/timestep.c
sbird: I could but it is longer code with a switch: easier to leave it as an if chain

sbird:
Thank you for the review!

sbird:
Looks like it still works, merging!

PR #529, opened 2021-03-22
Run treewalk_build in treewalk_do_hsml_loop before treewalk_run

sbird:
This ensures that the ReDoQueue only needs be as large as the workset,
which is for density 1/2 of the total particles.

Memory optimization


sbird:
Zero change, merging when green.

PR #532, opened 2021-04-06
Gradrho

sbird:
Reduce peak memory usage by allocating a smaller |Grad Rho| for h2 and by not using the NgbList for density()

### DIFF
--- a/libgadget/density.c
+++ b/libgadget/density.c
@@ -205,7 +205,8 @@ density(const ActiveParticles * act, int update_hsml, int DoEgyDensity, int Blac
     struct DensityPriv priv[1];
 
     tw->ev_label = "DENSITY";
-    tw->visit = (TreeWalkVisitFunction) treewalk_visit_ngbiter;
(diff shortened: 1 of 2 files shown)


libgadget/density.c
```diff
@@ -539,6 +538,11 @@ density_postprocess(int i, TreeWalk * tw)
 
         SPHP(i).DivVel /= SPHP(i).Density;
         P[i].DtHsml = (1.0 / NUMDIMS) * SPHP(i).DivVel * P[i].Hsml;
+
+        MyFloat * gradrho = DENSITY_GET_PRIV(tw)->GradRho;
+        /* Before this is |grad rho|^2, after it is |grad rho|*/
```
rainwoodman, line 543: Perhaps note this near the declaration of gradrho too.

sbird: Done!

sbird:
The last commit, which actually reduces GradRho, made some changes to output. I want this partial version for the frontera run, so merging while I figure it out.

PR #533, opened 2021-04-09
Fix an edge case with treewalk buffer fillup

sbird:
If the buffer fills up exactly on the end of a chunk, we don't need to
remove the final partially exported particle, because it is actually
fully exported. Also make the BufferFullFlag atomic to avoid a
(technical) data race.


sbird:
This edge case is hard to test. The code looks right so I will merge it. The test failure was a conda http error.

rainwoodman:
Was the removal an error?

sbird:
I think so, because evaluated[k] was already set. So the externals might not have been evaluated.

I also had this error:
[ 000481.16 ] ERROR: Task 3476: Something screwed up in export queue: nexp 73397 (local 2) last 28117500 != index 28117499

which I think is related: looks like NThisParticleExport was too large. But I'm not sure this is the real fix.

PR #535, opened 2021-04-14
Datarace

sbird:
With recent clang versions the thread sanitizer works on openmp! I ran the code under thread sanitizer and found a couple of (probably harmless) bugs, which this fixes.


sbird:
All these races are harmless, I think, but it is a good idea not to have them!

sbird:
Commit 37564becfce3f4db94624b5cb836f47f212c56b4 marginally changes output, perhaps because it is a real race but more likely because it changes the floating point optimizations done by the compiler.

PR #538, opened 2021-04-25
Some attempts to fix the wind model

sbird:
We have noticed that the DelayTime is sometimes infinite. It seems likely that the wind velocity v is either zero or NaN in these cases. These are some attempts to fix this.


sbird:
Tested, works, merging!

PR #539, opened 2021-05-27
for James -- Would something like this work?

rainwoodman:
@daviesje


sbird:
Closing as James merged this into the other PR

PR #540, opened 2021-05-27
Exclude wind from BH density

sbird:
Exclude wind particles from the BH density, to match the accretion treewalk, where they are excluded.


sbird:
@yueyingn : here you go.

PR #542, opened 2021-06-17
Allow FOF exchanges to have multiple iterations

sbird:
Fixes the PIG output when LARGE_EXCHANGE_BROKEN is on


sbird:
There is still a failure case when the exchange can temporarily exceed MaxPart: not sure what to do there. This should normally never happen with FOF exchanges because every rank ends up with the same number of particles as it started with. However for partial exchanges more may be sent than received.

PR #543, opened 2021-06-29
Subgrid winds

sbird:
This modifies the wind model to improve the 'subgrid' winds, which come from the star-forming gas, and allow sigma subgrid winds.


sbird:
Zero change to output for examples/hydro with subgrid winds off, some nice cleanups and good to support winds more like Illustris. Merging.

PR #545, opened 2021-07-02
Replace MPI_LARGE_EXCHANGE_BROKEN with a limit on each alltoall

sbird:
Instead of limiting the total size of the outgoing buffer, limit the size that will be sent via each Alltoallv. So instead of the limit being base part + all slots < 2 GB, it is now max((base part), slots) < 2GB . This is a bit less constraining and corresponds better to the things that actually crash.

This is on by default because I have not yet had an exchange that exceeds these bounds succeed.


sbird:
It works!

PR #557, opened 2021-11-18
Typetree

sbird:
Build trees containing just SPH particles for the hydro code and the metal return code.

This makes the tree walk faster (especially for the metal return code) and makes hmax computation much faster. The cost is an extra tree build, but because there are much fewer particles this is a net win. Some changes to output because edge decisions in the tree walk are now slightly different.


sbird:
I will remove the timing information and rebase to asterix before merging.

sbird:
Not worth it as-is, but keeping in case useful for Gadget-4 style hamiltonian splitting

sbird:
Closing as it is all in the hierarchical gravity branch

PR #558, opened 2021-11-18
Treebuildsched

sbird:
Optimize the treebuild merge with some scheduling.


sbird:
Makes treebuild a lot faster, merging.

PR #561, opened 2021-12-04
Fastertimestep

sbird:
Use openmp 4.5 array reductions to avoid false variable sharing.


sbird:
This turns out to make the code (slightly) slower: I think the gcc implementation of array reduction is perhaps not optimal.

PR #566, opened 2021-12-27
Notimeallvars

sbird:
Long overdue cleanup of the various time variables, which were set every timestep. I was able to disentangle them enough to pass them as variables and remove the All members.


sbird:
This avoids doing Time = exp(log(TimeInit)) at the beginning, which should avoid us auto-writing snapshots on step 0.

sbird:
Some small (fp roundoff) changes from commit https://github.com/MP-Gadget/MP-Gadget/pull/566/commits/ac0c05df63e57817440663fb0ed2f6a4388746af which changes a multiplication to a division and some more from 1546c19 which removes the round-tripping of All.TimeInit into All.Time. Otherwise nothing.

PR #570, opened 2021-12-30
Hierarchical Gravity and timestep splitting

sbird:
Use two timesteps in the code: one for hydro and one for gravity. Allow hydro to subsample gravity. This is what Gadget-4 does and it allows us to sometimes avoid building the full gravity tree.

Question: does BH repositioning and dynamic friction go on the hydro timescale or the gravity timescale? 
If it can be on the longer gravity timescale this would help us avoid more full treebuilds.

### DIFF
--- a/gadget/params.c
+++ b/gadget/params.c
@@ -180,7 +180,7 @@ create_gadget_parameter_set()
     param_declare_int(ps, "TreeGravOn", OPTIONAL, 1, "Enables tree gravity");
     param_declare_int(ps, "RadiationOn", OPTIONAL, 1, "Include radiation density in the background evolution.");
     param_declare_int(ps, "FastParticleType", OPTIONAL, 2, "Particles of this type will not decrease the timestep. Default neutrinos.");
-    param_declare_double(ps, "PairwiseActiveFraction", OPTIONAL, 0, "Pairwise gravity instead of tree gravity is used if N(active particles) / N(particles) is less than this. Usually slower.");
+    param_declare_double(ps, "PairwiseActiveFraction", OPTIONAL, 0, "Pairwise gravity instead of tree gravity is used if N(active particles) / N(particles) is less than this. Currently unimplemented as slower.");
 
     param_declare_double(ps, "GravitySoftening", OPTIONAL, 1./30., "Softening for collisionless particles; units of mean separation of DM. ForceSoftening is 2.8 times this.");
     param_declare_int(ps, "GravitySofteningGas", OPTIONAL, 1, "0 to use adaptive softening, where the gas softening is the smoothing length of the last step.");
--- a/gadget/params.c
+++ b/gadget/params.c
@@ -107,7 +107,7 @@ create_gadget_parameter_set()
     param_declare_double(ps, "BHOpeningAngle", OPTIONAL, 0.175, "Barnes-Hut opening angle. Alternative purely geometric tree opening angle. Lower values are more accurate.");
     param_declare_double(ps, "TreeRcut", OPTIONAL, 6, "Number of mesh cells at which we cease walking.");
     param_declare_int(ps, "TreeUseBH", OPTIONAL, 2, "If 1, use Barnes-Hut opening angle rather than the standard Gadget acceleration based opening angle. If 2, use BH criterion for the first timestep only, before we have relative accelerations.");
-    param_declare_int(ps, "SplitGravityTimestepsOn", OPTIONAL, 0, "This flag enables the momentum conserving hierarchical timestepping, where only active particles gravitate, from Gadget 4, for the short-range gravity, and splits the hydro and gravitational timesteps.");
+    param_declare_int(ps, "SplitGravityTimestepsOn", OPTIONAL, 1, "This flag enables the momentum conserving hierarchical timestepping, where only active particles gravitate, from Gadget 4, for the short-range gravity, and splits the hydro and gravitational timesteps.");
 
     param_declare_double(ps, "Asmth", OPTIONAL, 1.5, "The scale of the short-range/long-range force split in units of FFT-mesh cells."
                                                       "Larger values suppresses grid anisotropy. ShortRangeForceWindowType = erfc supports any value. 'exact' only supports 1.5. ");
--- a/gadget/params.c
+++ b/gadget/params.c
@@ -107,7 +107,7 @@ create_gadget_parameter_set()
     param_declare_double(ps, "BHOpeningAngle", OPTIONAL, 0.175, "Barnes-Hut opening angle. Alternative purely geometric tree opening angle. Lower values are more accurate.");
     param_declare_double(ps, "TreeRcut", OPTIONAL, 6, "Number of mesh cells at which we cease walking.");
     param_declare_int(ps, "TreeUseBH", OPTIONAL, 2, "If 1, use Barnes-Hut opening angle rather than the standard Gadget acceleration based opening angle. If 2, use BH criterion for the first timestep only, before we have relative accelerations.");
-    param_declare_int(ps, "HiSplitGravityOn", OPTIONAL, 0, "This flag enables the momentum conserving hierarchical timestepping, where only active particles gravitate, from Gadget 4, for the short-range gravity, and splits the hydro and gravitational timesteps.");
+    param_declare_int(ps, "SplitGravityTimestepsOn", OPTIONAL, 0, "This flag enables the momentum conserving hierarchical timestepping, where only active particles gravitate, from Gadget 4, for the short-range gravity, and splits the hydro and gravitational timesteps.");
 
     param_declare_double(ps, "Asmth", OPTIONAL, 1.5, "The scale of the short-range/long-range force split in units of FFT-mesh cells."
                                                       "Larger values suppresses grid anisotropy. ShortRangeForceWindowType = erfc supports any value. 'exact' only supports 1.5. ");
--- a/libgadget/blackhole.c
+++ b/libgadget/blackhole.c
@@ -634,7 +634,11 @@ blackhole(const ActiveParticles * act, double atime, Cosmology * CP, ForceTree *
     if(!tree->tree_allocated_flag)
     {
         message(0, "Building tree in blackhole\n");
-        force_tree_rebuild_mask(tree, ddecomp, GASMASK | DMMASK | STARMASK | BHMASK, NULL);
+        int treemask = GASMASK | STARMASK | BHMASK;
+        /* Don't necessarily need dark matter */
+        if(blackhole_params.BH_DynFrictionMethod > 1 || blackhole_params.BlackHoleKineticOn)
+            treemask += DMMASK;
+        force_tree_rebuild_mask(tree, ddecomp, treemask, NULL);
         walltime_measure("/BH/Build");
     }
 
--- a/libgadget/blackhole.c
+++ b/libgadget/blackhole.c
@@ -917,7 +917,9 @@ blackhole_dynfric_ngbiter(TreeWalkQueryBHDynfric * I,
         LocalTreeWalk * lv){
 
    if(iter->base.other == -1) {
-        iter->base.mask = GASMASK + DMMASK + STARMASK + BHMASK;
+        iter->base.mask = GASMASK + STARMASK + BHMASK;
+        if (blackhole_params.BH_DynFrictionMethod > 1 )
+            iter->base.mask += DMMASK;
         iter->base.Hsml = I->Hsml;
         iter->base.symmetric = NGB_TREEFIND_ASYMMETRIC;
         density_kernel_init(&iter->dynfric_kernel, I->Hsml, GetDensityKernelType());
@@ -1134,7 +1136,10 @@ blackhole_accretion_ngbiter(TreeWalkQueryBHAccretion * I,
         for(d = 0; d < 3; d++) {
             O->BH_MinPotPos[d] = I->base.Pos[d];
         }
-        iter->base.mask = GASMASK + DMMASK + STARMASK + BHMASK;
+        iter->base.mask = GASMASK + STARMASK + BHMASK;
+        /* Add DM if needed*/
+        if(blackhole_params.BlackHoleKineticOn == 1)
+            iter->base.mask += DMMASK;
         iter->base.Hsml = I->Hsml;
         iter->base.symmetric = NGB_TREEFIND_ASYMMETRIC;
 
--- a/libgadget/blackhole.c
+++ b/libgadget/blackhole.c
@@ -197,6 +197,9 @@ struct BHPriv {
     /* Counters*/
     int64_t * N_sph_swallowed;
     int64_t * N_BH_swallowed;
+    double FgravkickB;
+    double gravkicks[TIMEBINS+1];
+    double hydrokicks[TIMEBINS+1];
 };
 #define BH_GET_PRIV(tw) ((struct BHPriv *) (tw->priv))
 
@@ -513,7 +516,7 @@ collect_BH_info(int * ActiveBlackHoles, int NumActiveBlackHoles, struct BHPriv *
 
 
 void
-blackhole(const ActiveParticles * act, double atime, Cosmology * CP, ForceTree * tree, DomainDecomp * ddecomp, const struct UnitSystem units, FILE * FdBlackHoles, FILE * FdBlackholeDetails)
+blackhole(const ActiveParticles * act, double atime, Cosmology * CP, ForceTree * tree, DomainDecomp * ddecomp, DriftKickTimes * times, const struct UnitSystem units, FILE * FdBlackHoles, FILE * FdBlackholeDetails)
 {
     /* Do nothing if no black holes*/
     int64_t totbh;
@@ -579,6 +582,19 @@ blackhole(const ActiveParticles * act, double atime, Cosmology * CP, ForceTree *
     priv->a3inv = 1./(atime * atime * atime);
     priv->hubble = hubble_function(CP, atime);
     priv->CP = CP;
+
+    priv->FgravkickB = get_exact_gravkick_factor(CP, times->PM_kick, times->Ti_Current);
+    memset(priv->gravkicks, 0, sizeof(priv->gravkicks[0])*(TIMEBINS+1));
+    memset(priv->hydrokicks, 0, sizeof(priv->hydrokicks[0])*(TIMEBINS+1));
+    /* Compute the factors to move a current kick times velocity to the drift time velocity.
+     * We need to do the computation for all timebins up to the maximum because even inactive
+     * particles may have interactions. */
+    #pragma omp parallel for
+    for(i = times->mintimebin; i <= TIMEBINS; i++)
+    {
+        priv->gravkicks[i] = get_exact_gravkick_factor(CP, times->Ti_kick[i], times->Ti_Current);
+        priv->hydrokicks[i] = get_exact_hydrokick_factor(CP, times->Ti_kick[i], times->Ti_Current);
+    }
 
     /* Build the queue once, since it is really 'all black holes' and similar for all treewalks.*/
     treewalk_build_queue(tw_dynfric, act->ActiveParticle, act->NumActiveParticle, 0);
@@ -912,9 +928,15 @@ blackhole_dynfric_ngbiter(TreeWalkQueryBHDynfric * I,
             int k;
             O->SurroundingParticles += 1;
             O->SurroundingDensity += (mass_j * wk);
+            MyFloat VelPred[3];
+            if(P[other].Type == 0)
+                SPH_VelPred(other, VelPred, BH_GET_PRIV(lv->tw)->FgravkickB, BH_GET_PRIV(lv->tw)->gravkicks, BH_GET_PRIV(lv->tw)->hydrokicks);
+            else {
+                DM_VelPred(other, VelPred, BH_GET_PRIV(lv->tw)->FgravkickB, BH_GET_PRIV(lv->tw)->gravkicks);
+            }
             for (k = 0; k < 3; k++){
-                O->SurroundingVel[k] += (mass_j * wk * P[other].Vel[k]);
-                O->SurroundingRmsVel += (mass_j * wk * pow(P[other].Vel[k], 2));
+                O->SurroundingVel[k] += (mass_j * wk * VelPred[k]);
+                O->SurroundingRmsVel += (mass_j * wk * pow(VelPred[k], 2));
             }
         }
     }
@@ -1160,10 +1182,11 @@ blackhole_accretion_ngbiter(TreeWalkQueryBHAccretion * I,
             double dv[3];
             double da[3];
             int d;
-
+            MyFloat VelPred[3];
+            DM_VelPred(other, VelPred, BH_GET_PRIV(lv->tw)->FgravkickB, BH_GET_PRIV(lv->tw)->gravkicks);
             for(d = 0; d < 3; d++){
                 dx[d] = NEAREST(I->base.Pos[d] - P[other].Pos[d], PartManager->BoxSize);
-                dv[d] = I->Vel[d] - P[other].Vel[d];
+                dv[d] = I->Vel[d] - VelPred[d];
                 /* we include long range PM force, short range force from the last long timestep and DF */
                 da[d] = (I->Accel[d] - P[other].FullTreeGravAccel[d] - P[other].GravPM[d] - BHP(other).DFAccel[d]);
             }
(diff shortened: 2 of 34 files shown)


sbird:
At the moment (on the hydro example, which is too small for to really help) this avoids moment calculation for about 50% of timesteps. Almost all timesteps where we avoid the gravity treebuild still build a tree in blackhole.

sbird:
Rebased and fixed a bug which stopped treegrav running on PM steps. Avoid building a tree in BH if no BHs active.

sbird:
I have made the BH feedback treewalk asymmetric. I think this is fine because flagging the particles to be swallowed is also asymmetric, but someone else should confirm this.

sbird:
For the hydro example this now avoids building a tree (except for the gas tree) on the shortest timebin. The level up usually builds a gravity tree so it is hard to tell the real cost of the blackhole/wind tree. The cost of moment computation is now much smaller. The cost of the nodes is about the same, because we often build a gas tree and a full tree in the same timestep. This could be avoided if it makes sense on bigger problems.

sbird:
There are a few more stars forming in the hydro example with threads than before but there will be some variance because of the slightly different Hsml values that come out of the density iteration.

libgadget/blackhole.c
```diff
@@ -540,6 +541,21 @@ blackhole(const ActiveParticles * act, double atime, Cosmology * CP, ForceTree *
...
+     * accretion uses: all types but ONLY for repositioning potential minimum. Otherwise gas + black holes. gas + stars + BH is probably fine.
+     * gas + BH probably not enough if gas is sparse in halo.
+     * feedback uses: gas + black holes.
+     * The DM in dynamic friction and accretion doesn't really do anything, so could perhaps be removed from the treebuild later.
+     * However, we would still need a tree with gas + DM in the wind code.
+     */
+    if(!tree->tree_allocated_flag)
+    {
+        message(0, "Building tree in blackhole\n");
+        force_tree_rebuild_mask(tree, ddecomp, GASMASK + DMMASK + STARMASK + BHMASK, 0, NULL);
```
rainwoodman, line 555: "|"?

libgadget/gravshort-tree.c
```diff
@@ -118,14 +117,18 @@ grav_short_tree(const ActiveParticles * act, PetaPM * pm, ForceTree * tree, doub
...
         endrun(2, "Gravtree called before tree moments computed!\n");
 
     tw->ev_label = "GRAVTREE";
     tw->visit = (TreeWalkVisitFunction) force_treeev_shortrange;
-    /* gravity applies to all particles. Including Tracer particles to enhance numerical stability. */
-    tw->haswork = NULL;
+    /* gravity applies to all gravitationally active particles.
+     * If we are a PM step then all particles are active. */
+    tw->haswork = gravshort_haswork;
+    if(!act->ActiveParticle)
```
rainwoodman, line 130: This may be easier to read:

if(act->ActiveParticle) {
  tw->haswork = NULL;
} else {
  tw->haswork = gravshort_haswork;
}

Also, can we / shall we move the conditional into gravshort_haswork? (e.g. if !act_activeParticle return True) -- Because haswork function is supposed to be "the" conditional.

libgadget/run.c
```diff
@@ -386,14 +387,21 @@ run(int RestartSnapNum)
...
 
-        if(All.TreeGravOn) {
+        /* Build the force tree for tree gravity.*/
+        ForceTree Tree = {0};
+        int anygravactive = MPIU_Any(Act.NumActiveGravity, MPI_COMM_WORLD);
+
+        if(is_PM || (All.TreeGravOn && anygravactive))
+            force_tree_rebuild(&Tree, ddecomp, HybridNuGrav, !pairwisestep && All.TreeGravOn, All.OutputDir);
+
+        if(All.TreeGravOn && (is_PM || anygravactive)) {
```
rainwoodman, line 397: This looks puzzling -- how is this condition related to the one in line 394? Could you add a comment on when the first is True but the second is False? And looks like we will crash if the first is False but the second is True?

libgadget/blackhole.c
sbird: Sure, I suppose. I was following the style we use for the .mask in ngbiter.

libgadget/gravshort-tree.c
sbird: Yes you are right! I will adjust the condition. We maybe should but cannot move it into gravshort_haswork because we make memory allocation decisions in treewalk_build_queue depending on whether ->haswork is NULL.

libgadget/blackhole.c
```diff
@@ -524,8 +524,9 @@ blackhole(const ActiveParticles * act, double atime, Cosmology * CP, ForceTree *
     priv->hubble = hubble_function(CP, atime);
     priv->GravInternal = CP->GravInternal;
 
-    /* Build the queue once, since it is really 'all black holes' and similar for all treewalks*/
-    treewalk_build_queue(tw_dynfric, act->ActiveParticle, act->NumActiveParticle, 0);
+    /* Build the queue once, since it is really 'all black holes' and similar for all treewalks.
+     * Allocate high so we can keep the tree we build after making the list.*/
+    treewalk_build_queue(tw_dynfric, act->ActiveParticle, act->NumActiveParticle, 0, 1);
```
rainwoodman, line 529: Question 1: Does the force_tree_rebuild_mask alloc-high, such that build_queue must also alloc-high?

Question 2: The alloc_high knob adds a fragile coupling between build_queue and caller. 
This seems to be the only user of the knob. What if  we do an alloc-high here (on the caller side), and memcpy the queue?

libgadget/run.c
sbird: Yes, this is indeed confusing. The first is true and the second false when tree gravity is off. Then we still need a tree to do the PM. The first being false but the second being true should be impossible (I think): crashing is a legit thing to do there. I will add a comment.

libgadget/winds.c
```diff
@@ -252,8 +255,17 @@ int cmp_by_part_id(const void * a, const void * b)
 
 /* Find the 1D DM velocity dispersion of the winds by running a density loop.*/
 static void
-winds_find_weights(TreeWalk * tw, struct WindPriv * priv, int * NewStars, int NumNewStars, const double Time, const double hubble, ForceTree * tree)
+winds_find_weights(TreeWalk * tw, struct WindPriv * priv, int * NewStars, int NumNewStars, const double Time, const double hubble, ForceTree * tree, DomainDecomp * ddecomp)
 {
+    /* Flags that we need to free the tree to preserve memory order*/
+    priv->tree_alloc_in_wind = 0;
```
rainwoodman, line 261: The logic here reads like:

1. If there is a tree, use it
2. If there is no tree, build one, use it, then free it.

The usual way of implementing such logic is checking tree == NULL; So checking tree->tree_allocated_flag looks a bit surprising. I guess it is complicated to change the calling convention everywhere.

libgadget/blackhole.c
sbird: 1) The force_tree_rebuild_mask is alloc-low but I want to let the built force tree persist past the end of blackhole so I can reuse it for the winds. So I then can't free build_queue in time.
2) It isn't a bad idea. I was just trying to avoid a memcpy, but this is going to be a small one.

libgadget/winds.c
sbird: Indeed - it is not the tree that should be == NULL, but the tree->Nodes pointer and that has its own oddities with being offset from Nodes_base.

libgadget/blackhole.c
rainwoodman: I usually weight a small memcpy cheaper than a new function argument. :P

sbird:
A small 256^3 15 Mpc box on Frontera is 30 % faster. Let's see if it does anything to P(k), GSMF, HMF or lyman alpha...

libgadget/run.c
rainwoodman: "still need a tree to do the PM": it sounds like even after the GasTree split, the remaining gravTree can still be refactored into two -- the spatial indexing for PM, and the "moments" for short-range grav; in the future.

libgadget/blackhole.c
sbird: It is fair :)

libgadget/run.c
sbird: Yes such a split would be a good idea. We could save some memory for the gasTree, and reduce peak memory usage during density(). Just needs some messing about because C does not do generics :) It is not worth optimising the case where gravTree is disabled and density is on to avoid an extra rebuild because we almost never run in that mode.

sbird:
Thanks for the review!

sbird:
Looks like there is something wrong: I think the gravitational kick for the hydro-active but gravity-inactive SPH particles is not being computed correctly

[powerspectrum-0.3333.pdf](https://github.com/MP-Gadget/MP-Gadget/files/7825409/powerspectrum-0.3333.pdf)
.

sbird:
Got it: the kick was being applied twice if the particle timestep dropped when the particle was inactive (can happen because atime changes).

sbird:
Still the same: I think I need the Gadget-4 timestepping to make this work.

sbird:
Got it: bin_hydro and bin_gravity were swapped in one place

.

sbird:
(note above was run with this patch so that the double timestepping was enabled): 
diff --git a/libgadget/timestep.c b/libgadget/timestep.c
index 27dd6ac5..cfd19850 100644
--- a/libgadget/timestep.c
+++ b/libgadget/timestep.c
@@ -289,7 +289,7 @@ find_timesteps(const ActiveParticles * act, DriftKickTimes * times, const double
             bin_hydro = bin_gravity;
         /* Enforce that the gravity bin is the hydro bin. This is to ensure the code continues to
          * work while we implement the Hamiltonian timesplitting */
-        bin_gravity = bin_hydro;
+        //bin_gravity = bin_hydro;
         /* Only update if both the old and new timebins are currently active.
          * We know that the shorter hydro timestep is active, but we need to check
          * the gravity timestep.*/

sbird:
Current issues: star formation and cooling code is much slower (by a factor of 2).

sbird:
[bhmf-0.3300.pdf](https://github.com/MP-Gadget/MP-Gadget/files/8682930/bhmf-0.3300.pdf)
[gsmf-0.3300.pdf](https://github.com/MP-Gadget/MP-Gadget/files/8682931/gsmf-0.3300.pdf)
[hmf-0.3300.pdf](https://github.com/MP-Gadget/MP-Gadget/files/8682932/h
[powerspectrum-0.3300.pdf](https://github.com/MP-Gadget/MP-Gadget/files/8682933/powerspectrum-0.3300.pdf)
mf-0.3300.pdf)

sbird:
powerspectrum-0 3300-1
hmf-0 3300-1
gsmf-0 3300-1
bhmf-0 3300-1

sbird:
Mass functions all look good: possibly the 'master branch' had a bug during the test, I will re-run it.

sbird:
Ok - the speed regression in the cooling code is for a fairly straightforward reason: some particles are on shorter timebins now. There is a 'push down' logic which reduces the size of the top timebin so that the hierarchy is less deep. When this happens particles have their cooling evaluated more often, on the shorter timestep (as the hydro timestep is enforced to be shorter than the gravity timestep).
This is done because this code is slower if the timebins do not shrink fast enough (as each bin in the timestep hierarchy is evaluated each timestep it is active). So if the largest bin is bin 20 we can save gravitational evaluations by shrinking the timebin if N(bin 19) > N(bin 20)/3. But the gadget-4 calculation does not account for the fact that we also need to evaluate the hydro and other things twice as often for every hydro particle that shrinks the timebin. So we should be less aggressive at pushing down the bins.

sbird:
Rebased on top of current master

sbird:
Current code completes in 48:54 according to slurm and cpu.txt is:
Step 37818, Time: 0.33, MPIs: 32 Threads: 28 Elapsed: 2900.23
                                2909.65 100.0%        9.42 100.0%        9.42       9.42
 BH                               14.45  0.5%        0.01  0.1%        0.01       0.01
 Cooling                         399.09 13.7%        0.29  3.0%        0.29       0.29
  Build@winds.c:0266               72.58  2.5%        0.00  0.0%        0.00       0.00
  Cooling@sfr_eff.c:0272          179.47  6.2%        0.25  2.7%        0.18       0.29
  StarFormation@sfr_eff.c:0369      134.93  4.6%        0.04  0.4%        0.00       0.11
  Wind@winds.c:0414                12.10  0.4%        0.00  0.0%        0.00       0.00
 Domain                          182.06  6.3%        0.25  2.6%        0.23       0.26
 FOF                              95.91  3.3%        3.64 38.7%        3.64       3.65
PMgrav                          454.69 15.6%        1.24 13.2%        1.22       1.26
 SPH                             594.50 20.4%        0.39  4.1%        0.39       0.39
  Build@run.c:0444                124.82  4.3%        0.00  0.0%        0.00       0.00
  Density                         211.58  7.3%        0.18  1.9%        0.18       0.18
  HmaxUpdate@forcetree.c:1358        4.80  0.2%        0.00  0.0%        0.00       0.00
  Hydro                           253.30  8.7%        0.20  2.2%        0.20       0.20
  Write@checkpoint.c:0037          60.39  2.1%        1.54 16.4%        1.54       1.54
 Timeline                         97.48  3.4%        0.01  0.1%        0.01       0.03
 Tree                            942.01 32.4%        1.99 21.2%        1.99       1.99
  Build                            34.70  1.2%        0.04  0.4%        0.04       0.04
   Moments@forcetree.c:0127          6.89  0.2%        0.01  0.1%        0.01       0.01
   Nodes@forcetree.c:0220           27.81  1.0%        0.03  0.3%        0.03       0.03
  Recv@gravshort-tree.c:0163       39.71  1.4%        0.21  2.2%        0.01       0.26
  Wait1@gravshort-tree.c:0164      473.01 16.3%        0.93  9.9%        0.04       1.37
  Walk1@gravshort-tree.c:0159      345.28 11.9%        0.71  7.5%        0.27       1.60

For comparison the master branch is: 51:27
Step 37889, Time: 0.33, MPIs: 32 Threads: 28 Elapsed: 3050.77
 BH                                2.90  0.1%        0.00  0.0%        0.00       0.00
 Cooling                         334.09 10.9%        0.29  3.4%        0.29       0.29
  Cooling@sfr_eff.c:0270          175.52  5.7%        0.25  3.0%        0.18       0.29
  StarFormation@sfr_eff.c:0367      137.62  4.5%        0.04  0.4%        0.00       0.11
  Wind@winds.c:0398                20.95  0.7%        0.00  0.0%        0.00       0.00
 Domain                          173.35  5.7%        0.23  2.8%        0.22       0.25
 FOF                             109.92  3.6%        3.87 46.3%        3.86       3.87
 PMgrav                          458.19 15.0%        1.24 14.9%        1.23       1.27
 SPH                             566.07 18.5%        0.46  5.5%        0.46       0.46
  Build@forcetree.c:0173            0.68  0.0%        0.00  0.0%        0.00       0.00
  Density                         240.96  7.9%        0.20  2.4%        0.20       0.20
  Hydro                           324.42 10.6%        0.26  3.1%        0.26       0.26
  Timeline                         92.96  3.0%        0.02  0.2%        0.01       0.03
Tree                           1177.07 38.5%        0.89 10.6%        0.88       0.89
  Build                           342.64 11.2%        0.02  0.2%        0.01       0.02
   Moments@forcetree.c:0153         75.43  2.5%        0.00  0.0%        0.00       0.00
   Nodes@forcetree.c:0239          267.21  8.7%        0.02  0.2%        0.01       0.02
  HmaxUpdate@forcetree.c:1373        5.22  0.2%        0.00  0.0%        0.00       0.00
  Recv@gravshort-tree.c:0159      100.86  3.3%        0.12  1.5%        0.00       0.17
  Wait1@gravshort-tree.c:0160      360.90 11.8%        0.24  2.9%        0.00       0.40
  Walk1@gravshort-tree.c:0155      306.75 10.0%        0.42  5.0%        0.25       0.66

Notice that the tree build time is *much* less, although somewhat compensated by the need to build a tree in cooling for the winds. The tree walk imbalance time is a little worse: we could further speed this up at the cost of some memory by re-using the accelerations from the first half-step. The hydro and density modules are faster because they only work on a tree containing gas particles.

sbird:
At this time the code has a few potential optimisations, but no bugs. Let's test it a bit more and then merge it if it passes.

sbird:
Requesting reviews: this is the hierarchical gravity timestepping from Gadget-4.

sbird:
Build is failing because conda has apparently decided to take over the C compiler variables

sbird:
Compiler is broken in github - not sure why.

libgadget/gravshort.h
```diff
@@ -43,13 +43,19 @@ struct GravShortPriv {
...
     /* Matter density in internal units.
      * rho_0 = Omega0 * rho_crit
      * rho_crit = 3 H^2 /(8 pi G).
      * This is (rho_0)^(1/3) ,
      * Note: should account for
      * massive neutrinos, but doesn't. */
     double cbrtrho0;
+    /* Whether to calculate the gravitational potential.
+     * Note this is only useful on steps where all particles are active*/
+    int CalcPotential;
```
rainwoodman, line 56: The comment is a bit confusing -- you meant CalcPotential is True only on steps where all particles are active?

If so, perhaps rename this to AllParticlesAreActive?

libgadget/slotsmanager.h
```diff
@@ -46,6 +46,13 @@ struct bh_particle_data {
     /*******************************************************/
     double DragAccel[3];
     double DFAccel[3];
+    MyFloat FullGravAccel[3]; /* Short-range tree acceleration at the most recent timestep
```
rainwoodman, line 49: Does this include the PM / long range accel?

libgadget/tests/test_forcetree.c
```diff
@@ -226,7 +226,9 @@ static void do_tree_test(const int numpart, ForceTree tb, DomainDecomp * ddecomp
     /*Time creating the nodes*/
     double start, end;
     start = MPI_Wtime();
-    int nodes = force_tree_create_nodes(tb, numpart, ALLMASK, ddecomp, 0);
+    ActiveParticles Act = {0};
+    Act.NumActiveParticle = numpart;
```
rainwoodman, line 230: Is it necesary to set numactiveparticle on Act = {0}? I noticed above the NumActiveParticle line is omitted.

rainwoodman:
I'll take a look at the CI build error. Looks like I need to upgrade my laptop's OS first in order to reproduce the github workflow locally. (with act). Hopefully it is related to how conda changed there cross-platform compiler configurations on Linux.

gadget/params.c
```diff
@@ -106,6 +106,8 @@ create_gadget_parameter_set()
     param_declare_double(ps, "BHOpeningAngle", OPTIONAL, 0.175, "Barnes-Hut opening angle. Alternative purely geometric tree opening angle. Lower values are more accurate.");
     param_declare_double(ps, "TreeRcut", OPTIONAL, 6, "Number of mesh cells at which we cease walking.");
     param_declare_int(ps, "TreeUseBH", OPTIONAL, 2, "If 1, use Barnes-Hut opening angle rather than the standard Gadget acceleration based opening angle. If 2, use BH criterion for the first timestep only, before we have relative accelerations.");
+    param_declare_int(ps, "HiSplitGravityOn", OPTIONAL, 0, "This flag enables the momentum conserving hierarchical timestepping, where only active particles gravitate, from Gadget 4, for the short-range gravity, and splits the hydro and gravitational timesteps.");
```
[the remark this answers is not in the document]
sbird: I started with "HierarchicalGravityOn" but I kept mis-spelling "Hierarchical" and I suspected so would everyone else. Maybe HierarchGravityOn? Or SplitTimesteppingOn?

libgadget/forcetree.c
```diff
@@ -106,74 +106,52 @@ static void force_validate_nextlist(const ForceTree * tree)
...
 
     if(force_tree_allocated(tree)) {
         force_tree_free(tree);
     }
     walltime_measure("/Misc");
 
-    *tree = force_tree_build(PartManager->NumPart, ALLMASK, ddecomp, HybridNuGrav, DoMoments, EmergencyOutputDir);
+    *tree = force_tree_build(ALLMASK, ddecomp, act, HybridNuGrav, DoMoments, EmergencyOutputDir);
 
-    event_listen(&EventSlotsFork, force_tree_eh_slots_fork, tree);
```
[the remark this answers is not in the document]
sbird: It was to add a newly created star particle to the tree. But now we rebuild the tree more often and in smaller chunks so it is no longer needed. Actually it was not needed before, because the only tree walk we did after star formation was the winds, which do not use star particles. Also it didn't really work reliably: it could only add particles to the lowest level of the tree.

libgadget/tests/test_forcetree.c
sbird: That is a good point - it should not be necessary at the moment but I have been trying to do it in case that changes. This should be a proper function so it is consistent everywhere, and now it is.

sbird:
Force-push to pick up the fix to CI.

libgadget/slotsmanager.h
sbird: It does not...I will rename to make that clearer.

libgadget/gravshort.h
sbird: I reworded the comment: it is not quite AllParticlesAreActive because we still need the potential when hierarchical gravity is off.

libgadget/blackhole.c
```diff
@@ -600,10 +600,31 @@ blackhole(const ActiveParticles * act, double atime, Cosmology * CP, ForceTree *
         return;
     }
 
+    /* Move the working set high so we can keep the tree we build after making the list and still free this active set.*/
+    ActiveBlackHoles = mymalloc2("activeBH", NumActiveBlackHoles * sizeof(int));
```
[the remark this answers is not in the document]
sbird: It isn't exactly - it is used as the argument to all the later treewalks. I keep it in a separate array because it is weird to be passing a subarray of tw_dynfric around.

sbird:
Some changes were found at z=1 in the wind model when hierachical gravity was switched on and off: this might be fixed by using the correct predicted velocities for the DM velocity dispersion. Unfortunately bloats the size of struct particle_data at the moment.

sbird:
Current bug: 
[ 000975.26 ] Begin Step 512, Time: 0.0664731 (24c000), Redshift: 14.0437, Nf = 00000000314182, Systemstep: 0.000107811, Dloga: 0.00162187, status:
[ 000975.26 ] TotNumPart: 0000033554433 SPH 0000016777216 BH 0000000000 STAR 0000000000001
[ 000975.26 ] Occupied:      0      1      2      3      4      5 dt
[ 000975.26 ]  bin=15   16467518   16772733      0      0      0      0 0.00648748
[ 000975.26 ] X bin=14    309674     4454      0      0      1      0 0.00324374
[ 000975.26 ] X bin=13      24      29      0      0      0      0 0.00162187
[ 000975.26 ]        -----------------------------------
[ 000975.26 ] Total:     309698     4483      0      0      1      0 Sum:    314182
[ 000975.26 ] Tree construction for types: 1.
[ 000975.27 ] Tree constructed (type mask: 1) with 33576573 particles. First node 3670016, number of nodes 533848, first pseudo 5569500. NTopLeaves 1156
[ 000975.27 ] Treewalk DENSITY iter 0: total part 309698 max/MPI: 29821 min/MPI: 10521 balance: 1.54065.
[ 000975.30 ] Max ngb=117.35, min ngb=107.335
[ 000975.30 ] Treewalk DENSITY iter 1: total part 5256 max/MPI: 623 min/MPI: 146 balance: 1.8965.
[ 000975.30 ] Treewalk HYDRO iter 0: total part 309698 max/MPI: 29821 min/MPI: 10521 balance: 1.54065.
[ 000975.34 ] Tree constructed with 314180 particles (moments: 1). First node 3670016, number of nodes 15448, first pseudo 5569500. NTopLeaves 1156
[ 000975.34 ] Treewalk GRAVTREE iter 0: total part 314180 max/MPI: 30295 min/MPI: 10676 balance: 1.54281.
[ 000975.34 ] Tree constructed with 53 particles (moments: 1). First node 3670016, number of nodes 5656, first pseudo 5569500. NTopLeaves 1156
[ 000975.34 ] Treewalk GRAVTREE iter 0: total part 53 max/MPI: 32 min/MPI: 0 balance: 9.6602.
[ 000975.34 ] ERROR: Task 11: Tried to add 21562944 active particles, more than 21562864 allowed
[11:43](https://tacc-frontera.slack.com/archives/DR2AUBQBZ/p1655232236221199)
see /scratch3/06431/yueyingn/frontera25/L25N256wdpv/slurm-4490573.out

sbird:
Fixed. 
> Current bug: [ 000975.26 ] Begin Step 512, Time: 0.0664731 (24c000), Redshift: 14.0437, Nf = 00000000314182, Systemstep: 0.000107811, Dloga: 0.00162187, status: [ 000975.26 ] TotNumPart: 0000033554433 SPH 0000016777216 BH 0000000000 STAR 0000000000001 [ 000975.26 ] Occupied: 0 1 2 3 4 5 dt [ 000975.26 ] bin=15 16467518 16772733 0 0 0 0 0.00648748 [ 000975.26 ] X bin=14 309674 4454 0 0 1 0 0.00324374

sbird:
It runs astrid! 30% faster! It gives the same results on a small test and a large run.

gsmf-0 6000-1

bhmf-0 6000-1
hier-gsmf-0 2500-1

Merging!

PR #572, opened 2022-02-07
Implement an option to cool star forming gas faster to the equation of state.

sbird:
This lets gas cool to the equation of state using the cooling time instead of the relaxation time. Mirrors the Illustris behaviour.


sbird:
Dropping this since Yueying showed it makes no difference

PR #575, opened 2022-03-06
Cplusplus

sbird:
Some cleanups suggested by trying to compile MP-Gadget as C++ . A couple of real bugs but mostly just adding casts.

### DIFF
--- a/libgadget/utils/memory.h
+++ b/libgadget/utils/memory.h
@@ -45,24 +45,22 @@ struct AllocatorIter {
 };
 
 int
-allocator_init(Allocator * alloc, const char * name, size_t size, int zero, Allocator * parent);
+allocator_init(Allocator * alloc, const char * name, const size_t size, const int zero, Allocator * parent);
 
 int
-allocator_malloc_init(Allocator * alloc,
-        const char * name, size_t size, int zero, Allocator * parent
-        );
+allocator_malloc_init(Allocator * alloc, const char * name, const size_t size, const int zero, Allocator * parent);
 
 int
-allocator_split(Allocator * alloc, Allocator * parent, const char * name, size_t request_size, int zero);
+allocator_split(Allocator * alloc, Allocator * parent, const char * name, const size_t request_size, const int zero);
 
 int
 allocator_destroy(Allocator * alloc);
 
 void *
-allocator_alloc(Allocator * alloc, const char * name, size_t size, int dir, char * fmt, ...);
+allocator_alloc(Allocator * alloc, const char * name, const size_t size, const int dir, const char * fmt, ...);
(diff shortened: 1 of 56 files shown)


libgadget/utils/mpsort.c
```diff
@@ -48,8 +48,8 @@ DEFTYPE(uint64_t)
 static int _compar_radix(const void * r1, const void * r2, size_t rsize, int dir) {
     size_t i;
     /* from most significant */
-    const unsigned char * u1 = r1;
-    const unsigned char * u2 = r2;
+    const unsigned char * u1 = (const unsigned char *) r1;
```
rainwoodman, line 51: We'll need to sync the changes to mpsort.

libgadget/utils/memory.h
```diff
@@ -17,8 +17,8 @@ struct Allocator {
     char name[12];
     Allocator * parent;
 
-    void * rawbase;
-    void * base;
+    char * rawbase;
```
rainwoodman, line 20: I was following the convention of using void* for untyped buffer and temporarily cast to char for offsetting by bytes.

libgadget/utils/mpsort.c
sbird: I can open a PR if you would like them

libgadget/utils/memory.h
sbird: I thought of keeping that but we do some arithmetic on the void pointers (in fact that is the whole point of them). Although they have no types we largely use them as pointers to bytes, I think. I was concerned some compiler might do something weird with it if we miss a cast. Is there a benefit to keeping them as void * ?

rainwoodman: I don't have a strong opinion about enforcing this. No benefits that I can think of other than differentiating when the buffer is actually used for strings (char*) or as a untyped buffer. void * for untyped is a convention followed by most other C libraries (even glibc), so we may run into some frictions, but it's not that bad.

Compilers should trigger an error for direct arithmetic on void * it is undefined behavior offsetting an untyped pointer (we shouldn't be doing that in the first place) But I guess they emit warnings by default because void * was considered char* in very older versions of C.

libgadget/utils/mpsort.c
rainwoodman: Yes. That would help. Thanks!

sbird:
No change to output or time (as expected). Merging.

libgadget/utils/memory.h
sbird: Indeed - and for gcc there is no warning as there is an extension to treat void * as char * for arithmetic.

PR #576, opened 2022-03-31
Fixfof

sbird:
Some attempts to fix bugs in the FOF code, observed only with the large run. Came in between 0679e2d097..18698f5583
 
I suspect a race condition that became bigger once the fof treewalking got faster. Move the FOF test work in here.

### DIFF
--- a/libgadget/Makefile
+++ b/libgadget/Makefile
@@ -69,7 +69,7 @@ TESTED = hci \
 	gravity \
 	exchange
 
-MPI_TESTED = exchange
+MPI_TESTED = exchange fof
 
 TESTBIN :=$(UTILS_TESTED:%=.objs/utils/test_%) $(UTILS_MPI_TESTED:%=.objs/utils/test_%) $(TESTED:%=.objs/test_%) $(MPI_TESTED:%=.objs/test_%)
 SUITE?= $(TESTED:%=test_%) $(UTILS_TESTED:%=utils/test_%)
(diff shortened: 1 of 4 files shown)


sbird:
I can reproduce the problem (LenByType being messed up) restarting from a snapshot. 
It is not to do with Swallowed or Garbage particles. 
I can reproduce it with 1 thread per rank so it is not an openmp race condition.
It seems that LenByType is wrong even for DM particles. However, GroupID seems fine.

sbird:
It appears that small simulations are also not affected

sbird:
If I disable secondary labelling then all the lengths are good.

sbird:
git bisect says:
18698f5583d9645ab2bfea8f459c0901779eeeed is the first bad commit
commit 18698f5583d9645ab2bfea8f459c0901779eeeed
Author: Simeon Bird <sbird@ucr.edu>
Date:   Fri Nov 26 22:39:45 2021 -0800

    Add a FOF option to store GrNr in the particle table
    
    Only turn it on for saving, not seeding.

 libgadget/fof.c      | 43 +++++++++++++++++++++++--------------------
 libgadget/fof.h      |  6 ++++--
 libgadget/run.c      |  4 ++--
 libgadget/runtests.c |  3 +--
 4 files changed, 30 insertions(+), 26 deletions(-)

sbird:
Ok. Commit 827c393 is confirmed to fix the bug. I still do not understand the huge zero-DM halo!

sbird:
I will merge a minimal fix and continue debugging.

PR #577, opened 2022-04-04
Fixbiggas

sbird:
This series fixes the 'single big gas group' issue seen in the FOF tables since a long time. These particles were really 'all gas particles not in a group'. Also fix a potential issue with particles at the same positions, although I don't think we hit this in practice.

Confirmed that 2ba7b76 fixes the issue alone. b408475 makes similar issues impossible in future.


sbird:
I notice when testing this that the ordering of the groups varies between runs. Is it worth sorting the groups by total mass or DM particle length before output?

sbird:
I looked into this and the particles are sorted by GroupID, which is assigned by MinID: almost but not entirely DM length. I think it is better not to sort it by DM length as the code needs the MinID sorting in a couple of places.

PR #578, opened 2022-04-05
Noallvars

sbird:
This branch is an enormous chunk of cleanups which finally removes the huge single global structure and allvars.h
Various bits of the code got moved to smaller modules and split up along the way.


sbird:
Some (small) change from the whole series. No change in particle position up to 8da85354 .

sbird:
Hydro example is unchanged between [e1c6299](https://github.com/MP-Gadget/MP-Gadget/pull/578/commits/e1c6299ad90d2d20bf5b94449f3f9707a2dc52a0) and c58ed47

sbird:
Looks like the change in output rounding is caused by the rounding changes to the random offset in [e1c629](https://github.com/MP-Gadget/MP-Gadget/pull/578/commits/e1c6299ad90d2d20bf5b94449f3f9707a2dc52a0)

sbird:
No real changes to the output of power spectrum, hmf, gsmf or bhmf. Percent level changes due to the tree randomisation being a little different.

PR #579, opened 2022-04-14
Hierarchicalgravity

sbird:
This implements the hierarchical gravity splitting from Gadget-4. No tests yet but it runs!


sbird:
Here is the current change in powerspectrum between this branch (just before fixing the black hole mergers: [aabf183](https://github.com/MP-Gadget/MP-Gadget/pull/579/commits/aabf18303a69cb5f75ed8654b62820316875f288)) and an older copy of master. Pretty good!

powerspectrum-0 3300-1

sbird:
And at z=3: 
powerspectrum-0 2500-1

sbird:
With this version the tree is faster: there is still a 20% speed regression on a small problem which is due to the cooling code taking twice as long! Not sure why.

sbird:
Looks good, merging into the twotimestep branch for further work!

PR #581, opened 2022-05-26
Remove explicitly setting the BH feedback radius.

sbird:
We just noticed it was buggy: the accretion uses the BH Hsml, the feedback uses the FeedbackRadius, but the treewalk is only computed out to FeedbackRadius, so there may not be any particles inside the kernel.


sbird:
@rainwoodman  Do you know why the compile is failing on the github action? Looks like it can't find a compiler in conda anymore.

PR #583, opened 2022-06-06
Two tweaks to the wind model:

sbird:
1. Use the predicted velocity for the wind velocity dispersion. The current velocity may be evaluated at different times as
the other particle is inactive, and so we are not comparing two objects at the same kick time.
2. Make the type of particle whose velocity dispersion is used configurable.

This is on top of master but I need it for the hierarchical gravity mode: in that mode the kick times of inactive particles are somewhat different on short timesteps, thus increasing the difference between particle velocities and so wind velocity launch. Using predicted velocities is the physically correct thing to do (even without hierarchical gravity) and I would like to know how much difference it makes.

I also made the type of particle used configurable: it would be worth checking if WindVDispType == 0 works about the same as WindVDispType == 1. If so the hierarchical model can be made more efficient and avoid building the full tree entirely on most timesteps.


sbird:
WindVDIspType == 0 does not seem to work at all, the winds are not effective. Will drop it.

sbird:
However, the first change, using the predicted velocities, makes a small difference to the model when applied to the master branch. I will port it to hierarchical gravity and merge there.

PR #584, opened 2022-07-18
Fix spurious rare export buffer full crash:

sbird:
The error is:
[ 059851.77 ] ERROR: Task 673: Something screwed up in export queue: nexp 1272278 (local 2) last 79115500 != index 79115499

This happens when lastreal (the last element in the workqueue
which was evaluated) is larger than the last element in the export
buffer. But that is actually fine: there can be particles with no
exports. Weaken the condition to allow for this.


sbird:
Yueying does this fix it?

yueyingn:
Yes, this fixes the issue. The restart run has finished without hitting into trouble.

PR #586, opened 2022-08-09
Fixhmaxupdate

sbird:
This series fixes a bug in the computing of hmax - Node.len was assumed to be radius when it is really diameter. 
This should not be that serious in the master branch, but becomes more serious in the twotimestep branch.

Also some cleanups from the twotimestep branch: some helper scripts and speeding up the DEBUG mode so it is a bit more usable.


sbird:
@yueyingn This should be applied to the astrid branch as well.

PR #589, opened 2022-08-15
Reorder the exchange memory allocations so there is a little more ava…

sbird:
Reorder the exchange memory allocations so there is a little more available when we reserve new slots.


sbird:
@yueyingn Does this fix it?

yueyingn:
This fix is good!

sbird:
Thanks!

PR #591, opened 2022-08-23
Excursionsetcleanup

sbird:
Some small cleanups after the excursion set merge.


sbird:
@daviesje Does this look good to you?

sbird:
And pass your tests :)

PR #593, opened 2022-08-30
Separatevdisp

sbird:
Separate the velocity dispersion calculation from the rest of the winds and only run it on PM steps. This produces about the same result in a small simulation (and is what Illustris did) but it means we do not have to build a tree with DM particles for winds on short timesteps.

### DIFF
--- a/libgadget/blackhole.c
+++ b/libgadget/blackhole.c
@@ -1690,3 +1690,166 @@ void blackhole_make_one(int index, const double atime) {
     /* Initialize KineticFdbkEnergy, keep zero if BlackHoleKineticOn is not turned on */
     BHP(child).KineticFdbkEnergy = 0;
 }
+
+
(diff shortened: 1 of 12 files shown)


sbird:
This also largely simplifies computing predicted quantities, largely removing the overhead from computing predicted pressures.

sbird:
The Kinetic feedback is done at the same time as the wind velocity dispersion. This means that we don't generally need a dark matter tree at all on short timesteps.

sbird:
Tests that have been done:
- Check the GSMF and BHMF. No real change.
- Instrument the code so that it prints VDisp on a kinetic feedback event. It looks like over a PM step the BH VDisp changes by about 0.1 %, which seems unlikely to make a difference, especially since VDisp is only used to threshold the kinetic feedback event, not control the amount of energy returned.

PR #595, opened 2022-09-08
Timebase64

sbird:
64-bit timeline. This is not strictly necessary but it means we can have a lot more syncpoints and it won't crash if the timestep gets a little too short.


sbird:
We don't need this, but I'm keeping it around in case we do at some point.

PR #598, opened 2022-10-11
Bhcleanup2

sbird:
I noticed that blackhole.c was getting a bit long, so I split it up into smaller files. Also more functions are created and some cleanups applied.


sbird:
@nianyic7 This should make your ongoing Dynfric work easier, but can you check?

sbird:
Thank you @nianyic7 !

sbird:
No change to hydro example at all, except from [01f386f](https://github.com/MP-Gadget/MP-Gadget/pull/598/commits/01f386fe672aff8f163537a9e6107b4255df92f1) which changes the defaults. The hydro example should use dynfric anyway so that it gets tested!

sbird:
Merging, confirmed by @nianyic7

PR #599, opened 2022-10-28
Dynfricstep

sbird:
This gives dynamic friction its own timestep, using an acceleration criterion like the gravity timestep. This way we don't need the tree with the stars so often.


sbird:
@nianyic7 What do you think of this? If we can make the dynamic friction timestep longer it would save us a lot of time on the big run. It seems to work ok on a small box. If you get a chance to try it on a chunk of a larger box and make sure the BHs still sink, that would be very helpful!

PR #600, opened 2023-02-08
Fix a timestep bug

sbird:
The code for handling the special case where all the active gas particles had formed stars looks through all stars, rather than all recently formed stars, and was incorrectly not collective.

Fix this. Thanks to Nishikanta Khandai <nkhandai@niser.ac.in> for the bug report.


sbird:
@nianyic7 @yueyingn This should go to the astrid-ii branch

nkhandai:
I still get some (8)  instances of  -->  HYDRO Ngblist: min 35184372088832 max 0 avg -nan average exports: 0 stdout in the hydro run ... 10Mpc, 64^3 . The code does not crash or gracefully exit as before . Before the fix, I had 18 instances of -->  HYDRO Ngblist: min 35184372088832 max 0 avg -nan average exports: 0 and 2 instances of  -- > GRAVTREE Ngblist: min 35184372088832 max 0 avg -nan average exports: 0

sbird:
The residual message is harmless - it just means that no gas particles are active that timestep, or no particles are gravity active that timestep. I could get rid of it but I'd rather keep it for diagnostics.

PR #601, opened 2023-02-17
Remove 64-bit debug variable unless DEBUG is on

sbird:
@yueyingn What do you think of this? It makes every sph_particle_data and every star_particle_data 64-bits smaller, but it removes a check in the exchange code that has caught bugs before (although I think not bugs in production). Living dangerously but likely reduces memory use noticeably.


yueyingn:
It looks good to me! Let's just merge it and see how that goes with the production.
I am stilling waiting for the tests of the star particle slot.

PR #602, opened 2023-02-20
reduce memory of star particle slot

yueyingn:
reduce memory usage of the star particle struct


yueyingn:
gas particle displacement of hydro example

disp-hist

and gsmf of hydro example at z=0
hydro-example-gsmf

sbird:
Merge it!

PR #603, opened 2023-03-17
fix bhinfo collect error when df off

nianyic7:
When bhdyfric_method=0, the dynpriv struct is not allocated and yet used when collecting bhinfo, resulting in an error. This fix allocates dynpriv even when dynfric is off and set all variables in the structure to zero. Should not affect the run when the bhdyfric is turned on.


sbird:
@nianyic7 Thanks for noticing the bug. I don't much like the patch, I'm afraid, because it means the allocation of the object is duplicated. The second allocation also wastes a bit of memory. I'd prefer if the dynpriv structure were allocated to null and then we checked the arrays when we write it in bhinfo and when we free the struct.

PR #604, opened 2023-03-21
Bhfixes

sbird:
Some potential fixes for the UMBH NaN issue


yueyingn:
This seems to work. We did not hit NaN this time in the new run.

PR #607, opened 2023-05-02
Separatetree

sbird:
This lets DF use a separate, star-only, tree. The minimum potential computation uses the same tree.

The idea here is that accretion and feedback now use a gas only tree, and so the treewalks are faster.

### DIFF
--- a/examples/dynfric/paramfile.gadget
+++ b/examples/dynfric/paramfile.gadget
@@ -3,13 +3,9 @@ InitCondFile = ICS/IC
 OutputDir = output
 TreeCoolFile = ../TREECOOL_fg_june11
 OutputList = 0.1,0.1111111, 0.125, 0.142857, 0.16666, 0.181818,0.2,0.222222, 0.25, 0.285714, 0.3333333,0.4
-#OutputList =  0.2, 0.222222, 0.25, 0.285714, 0.333333 
-#OutputList = 0.333333,0.4
-Nmesh = 1100
 # CPU time -limit
-	
-#TimeLimitCPU = 43000 #= 8 hours
-TimeLimitCPU = 230000
+
+TimeLimitCPU = 28000
 
 # Code options
 #  Characteristics of run
@@ -31,45 +27,22 @@ MassiveNuLinRespOn = 0
 WindOn = 1
 MetalCoolFile = ../cooling_metal_UVB
 
-# Accuracy of time integration
-MaxSizeTimestep = 0.1
-MinSizeTimestep = 0.00
-
 SnapshotWithFOF = 1
-FOFHaloLinkingLength = 0.2
-FOFHaloMinLength = 32
 
 #  Further parameters of SPH
 DensityKernelType = quintic
 
-DensityContrastLimit = 100   # max contrast for hydro force calculation
-DensityResolutionEta = 1.0  # for Cubic spline 1.0 = 33
-MaxNumNgbDeviation = 2
-ArtBulkViscConst = 0.75
-InitGasTemp = 580.0        # always ignored if set to 0 
-MinGasTemp = 5.0
-
-
-# Softening lengths
-
-MinGasHsmlFractional = 0.01
-
 #----------------------BH Stuff-------------------------
 BlackHoleFeedbackFactor = 0.05
-BlackHoleFeedbackRadius = 0.
-BlackHoleFeedbackRadiusMaxPhys = 0.
 BlackHoleFeedbackMethod = spline | mass
 SeedBlackHoleMass = 5.0e-5
-BlackHoleAccretionFactor = 100.0
-BlackHoleNgbFactor = 2.0
-BlackHoleEddingtonFactor = 3.0
 MetalReturnOn = 1
 MinFoFMassForNewSeed = 1
 TimeBetweenSeedingSearch = 1.03
 
 WriteBlackHoleDetails = 1
 BH_DynFrictionMethod = 2  # no DM/Star
-BH_DRAG = 1  # No drag
+BH_DRAG = 1  # drag
 BH_DFBoostFactor = 1
 SeedBHDynMass = 2e-3
 BH_DFbmax = 10.
@@ -82,9 +55,6 @@ CritPhysDensity = 0       #  critical physical density for star formation in
 CritOverDensity = 57.7   #  overdensity threshold value
 
 WindModel = ofjt10,decouple
-WindEfficiency = 2.0
-WindEnergyFraction = 1.0
-WindSigma0 = 353.0 #km/s
 WindSpeedFactor = 3.7
 
 WindFreeTravelLength = 20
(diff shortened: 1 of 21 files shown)


sbird:
This also fixes some bugs in the BH merger.

sbird:
This looks good, but for some reason the hydro compute becomes very slow on the test case. I thought this was because the hmax can be larger for the BHs, but it happens even before the first BH is formed! Mysteriously, [5b15f68](https://github.com/MP-Gadget/MP-Gadget/pull/607/commits/5b15f686a55d42f9f6d042277950dd8d5ea83843) does make things better, again even before BHs form. I will investigate more.

rainwoodman:
Do we have a dual tree walk iterator? I donot recall we did and it is
unclear to me,  after a quick glance that 5b15f68 adds that. Given that I
am somewhat worried how the interaction between bh and gas is iterated if
bh and gas are now on two different trees. Not sure if the second part is
exactly true however.

Yu

sbird:
https://github.com/MP-Gadget/MP-Gadget/commit/5b15f686a55d42f9f6d042277950dd8d5ea83843 was an attempt to fix the hydro getting slow and will probably be dropped. It does actually work, partially, but I don't understand why, so the true fix will probably replace it. 

 I tried to fix it another way by moving the BH mergers into a different treewalk with a tree containing only BHs, see here: https://github.com/MP-Gadget/MP-Gadget/pull/615 which I think is what you are talking about. Here the gas accretion and feedback will be handled first, then the BH mergers will be done after.

What do you mean by a dual tree walk iterator? I don't think we do have it.

rainwoodman:
On Mon, Aug 7, 2023 at 8:21 AM Simeon Bird ***@***.***> wrote:

> 5b15f68
> <https://github.com/MP-Gadget/MP-Gadget/commit/5b15f686a55d42f9f6d042277950dd8d5ea83843>
> was an attempt to fix the hydro getting slow and will probably be dropped.
> It does actually work, partially, but I don't understand why, so the true
> fix will probably replace it.
>

The essence there is splitting out bh hmax? which tends to be larger than
hydro (by how much) the hydro speed is negatively affected by hmax, as that
increases the overlap between tree cells making comm and tree pruning both
inefficient.

> I tried to fix it another way by moving the BH mergers into a different
> treewalk with a tree containing only BHs, see here: #615
> <https://github.com/MP-Gadget/MP-Gadget/pull/615> which I think is what
> you are talking about. Here the gas accretion and feedback will be handled
> first, then the BH mergers will be done after.
>
> What do you mean by a dual tree walk iterator? I don't think we do have it.
>
No we don’t. I thought about this a bit more. Bh and hydro are active
particles (a list) vs tree particles forces so we never used a dual tree
algorithm for them. I just forgot about that.

A dual tree algorithm means building another tree for the active particles,
and walk two trees together. I think the speed gain is at best marginal as
most of the steps the active list is very short.

sbird:
> On Mon, Aug 7, 2023 at 8:21 AM Simeon Bird ***@***.***> wrote: [5b15f68](https://github.com/MP-Gadget/MP-Gadget/commit/5b15f686a55d42f9f6d042277950dd8d5ea83843) <[5b15f68](https://github.com/MP-Gadget/MP-Gadget/commit/5b15f686a55d42f9f6d042277950dd8d5ea83843)> was an attempt to fix the hydro getting slow and will probably be dropped. It does actually work, partially, but I don't understand why, so the true fix will probably replace it.
> The essence there is splitting out bh hmax? which tends to be larger than hydro (by how much) the hydro speed is negatively affected by hmax, as that increases the overlap between tree cells making comm and tree pruning both inefficient.

That's right. The difference between BH hmax and hydro hmax is by default 2, because we set BlackHoleNgbFactor = 2.0
by default.

> I tried to fix it another way by moving the BH mergers into a different treewalk with a tree containing only BHs, see here: #615 <#615> which I think is what you are talking about. Here the gas accretion and feedback will be handled first, then the BH mergers will be done after. What do you mean by a dual tree walk iterator? I don't think we do have it.
> No we don’t. I thought about this a bit more. Bh and hydro are active particles (a list) vs tree particles forces so we never used a dual tree algorithm for them. I just forgot about that. A dual tree algorithm means building another tree for the active particles, and walk two trees together. I think the speed gain is at best marginal as most of the steps the active list is very short.

I see!  That could indeed be fast for the longer timesteps: a lot of the time is still spent walking there.

sbird:
Here is the bisect:
git bisect start
# good: [71a18333b22f5cfca333fd9977795fae5ae25625] Write black hole details from hydro example
git bisect good 71a18333b22f5cfca333fd9977795fae5ae25625
# bad: [a8b2c10ea3ed8609dd7b7256c0e337991aa2c323] Promote a temporary variable
git bisect bad a8b2c10ea3ed8609dd7b7256c0e337991aa2c323
# good: [0febefb79ff69b1df28525f6623fb448dd23d6c4] Move the check for FeedbackWeightSum > 0 into the thermal feedback clause
git bisect good 0febefb79ff69b1df28525f6623fb448dd23d6c4
# bad: [948f6b312ad2baeac237567af2d66d6e64bf8b79] Add BHs to the hydro tree and re-use it for BH and star.
git bisect bad 948f6b312ad2baeac237567af2d66d6e64bf8b79
# bad: [87d7287e0c58ef4ccc1b33c0ed6804250b270777] Move BH minimum potential finding into Dynamic friction
git bisect bad 87d7287e0c58ef4ccc1b33c0ed6804250b270777
# good: [52702835c03d4ccf5a6f07ebb34de6fbbf883f3e] Adjust tree building so DF builds its own tree
git bisect good 52702835c03d4ccf5a6f07ebb34de6fbbf883f3e
# first bad commit: [87d7287e0c58ef4ccc1b33c0ed6804250b270777] Move BH minimum potential finding into Dynamic friction

sbird:
Note that https://github.com/MP-Gadget/MP-Gadget/commit/87d7287e0c58ef4ccc1b33c0ed6804250b270777 is nowhere near the treewalk or the hydro code. Also the times are not quite the same for either 'bad' or 'good.

sbird:
Annotated with the times for hydro:

good: [71a18333b22f5cfca333fd9977795fae5ae25625] Write black hole details from hydro example: 0.96
bad: [a8b2c10ea3ed8609dd7b7256c0e337991aa2c323] Promote a temporary variable : 2.84
good: [0febefb79ff69b1df28525f6623fb448dd23d6c4] Move the check for FeedbackWeightSum > 0 into the thermal feedback clause: 0.96
bad: [948f6b312ad2baeac237567af2d66d6e64bf8b79] Add BHs to the hydro tree and re-use it for BH and star. 7.0
bad: [87d7287e0c58ef4ccc1b33c0ed6804250b270777] Move BH minimum potential finding into Dynamic friction: 4.0
good: [52702835c03d4ccf5a6f07ebb34de6fbbf883f3e] Adjust tree building so DF builds its own tree: 1.25

# first bad commit: [87d7287e0c58ef4ccc1b33c0ed6804250b270777] Move BH minimum potential finding into Dynamic friction

Commit [5b15f686a5] is 1.63

sbird:
I wonder if this is just something in the communication playing up...these numbers look pretty random.

sbird:
I think what was happening was that the initial hsml check had an inverted condition, fixed by https://github.com/MP-Gadget/MP-Gadget/commit/c6ce0bc5c9f722f275be4bcb6471bbfb6830cfc0
which allowed for the reading of non-initialised memory (although it was quickly over-written) and that was leading the compiler to do crazy optimisation things. It looks fixed for now, but will run more tests.

PR #608, opened 2023-05-15
Fixentropylimit

sbird:
Fix the edge case where entropy becomes negative.


yueyingn:
I think this is good, let's merge it.

It saves the negative Entropy cases like:
EntVarPred<=0, ID=272419106789, SphP[PI].Entropy=2.05846e+10, SphP[PI].DtEntropy=2.70533e+16, dloga=-9.08339e-07

sbird:
Ok! Can you confirm that relaxing the entropy limit doesn't mess up the BHMF ?

sbird:
Entropy limit is very rarely hit, according to Yueying.

PR #611, opened 2023-06-01
Bhbugfixes

sbird:
This fixes some clear (but small effect) bugs in the BH module.

To Black hole mergers:
- Make the BH treewalk symmetric again, so that we check if either BH is close to the other one. Only matters if the BH Hsml is < 2 * Force softening.
- Ensure that the SwallowID does not persist over timesteps. This ensures that black holes which are marked but not swallowed (because the swallowing BH is swallowed first) are not prevented from merging in future timesteps. This should be rare (as the comment said), but is worth fixing.
 - We avoid two BHs trying to swallow each other (ie, break the symmetry) by checking that a black hole only merges another with a larger ID. But only active black holes can swallow, so this can prevent an inactive black hole from being merged. This again is a rare case, as BHs take their timesteps from the shortest gas timestep surrounding them and close BHs likely have the same shortest gas timestep.

To gas accretion:
- Conserve momentum on gas accretion. Small effect because most gas accretion happens for massive black holes.
- Fix an edge case where mass conservation could be broken when a small BH swallows a gas particle and the mass is just enough to push it over the SeedBHDynMass. This is rare because most BHs grow by merger in this regime.

### DIFF
--- a/examples/dynfric/paramfile.gadget
+++ b/examples/dynfric/paramfile.gadget
@@ -1,0 +1,91 @@
+#  Relevant files
+InitCondFile = ICS/IC
+OutputDir = output
+TreeCoolFile = ../TREECOOL_fg_june11
+OutputList = 0.1,0.1111111, 0.125, 0.142857, 0.16666, 0.181818,0.2,0.222222, 0.25, 0.285714, 0.3333333,0.4
+#OutputList =  0.2, 0.222222, 0.25, 0.285714, 0.333333 
+#OutputList = 0.333333,0.4
+Nmesh = 1100
+# CPU time -limit
+	
+#TimeLimitCPU = 43000 #= 8 hours
+TimeLimitCPU = 230000
+
+# Code options
+#  Characteristics of run
+TimeMax = 0.5
+
+Omega0 = 0.2865      # Total matter density  (at z=0)
+OmegaLambda = 0.713413      # Cosmological constant (at z=0)
+OmegaBaryon = 0.04628     # Baryon density        (at z=0)
+HubbleParam = 0.6932      # Hubble paramater (may be used for power spec parameterization)
+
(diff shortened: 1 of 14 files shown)


sbird:
Note to self: I have tested the effect of [6c9b6eb](https://github.com/MP-Gadget/MP-Gadget/pull/611/commits/6c9b6eb89e1be10940362e14f528c09040186ad1) in the hydro example and confirmed it makes *absolutely no difference* (the BHs merge at exactly the same time). Probably this just means that Hsml > FORCE_SOFTENING for all mergers. Let's leave it on anyway though, to save my confusion in future.

sbird:
@nianyic7 Can you test this please?

nianyic7:
Will do it this week!

sbird:
Here are all the merger redshifts in the small hydro box: it is a bit noisy

bhbugfix:
2.33196
1.93243
1.36667
0.944391
0.432518
0.280975
0.27347
0.238145

master:
2.49349
1.93165
1.39756
1.09913
0.910415
0.722506
0.502889
0.231111
0.152211
0.136876

Even 984e85b1a03437733a72d4e4c55f63ad61bbeb97 became a little different near the end:
bhbugfix not symmetric:
2.33196
1.93243
1.74705
1.51373
0.892852
0.739862
0.639497

sbird:
And here is the current separatetree branch in the hydro example: 
2.18318
1.43123
0.989074
0.837575
0.825227
0.640649
0.611867
(hard to test).

sbird:
Test results from Nianyi:
bhmf-test_master_c69f0bc-test_bhbugfixes_0febefb
mergers-test_master_c69f0bc-test_bhbugfixes_0febefb
offset-test_master_c69f0bc-test_bhbugfixes_0febefb

PR #615, opened 2023-08-07
Sepbhmerger

sbird:
This splits out the BH mergers into a separate set of tree walks.


sbird:
This does not seem to be faster.

PR #623, opened 2023-11-13
Treetweaks

sbird:
Tweaks to the tree to improve our communication and load balancing.

### DIFF
--- a/libgadget/blackhole.c
+++ b/libgadget/blackhole.c
@@ -1010,7 +1010,6 @@ blackhole_feedback(int * ActiveBlackHoles, int64_t NumActiveBlackHoles, ForceTre
     tw_feedback->result_type_elsize = sizeof(TreeWalkResultBHFeedback);
     tw_feedback->tree = tree;
     tw_feedback->priv = priv;
-    tw_feedback->repeatdisallowed = 1;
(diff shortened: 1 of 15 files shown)


sbird:
We ran a small hydro simulation, Yanhui checked it was the same. Mild 2% increase in the HMF for the smallest halos. Single threaded single MPI is bit by bit identical.

sbird:
Confirm that the FOF output is bit by bit identical when run on the same DM halos.

sbird:
Ok, the FOF mode itself is exactly the same before and after the changes. I took a harder look at the halos where the mass function changed by ~3% and they are all < 10 particles. They don't seem to cluster near domain boundaries in any obvious way.

sbird:
This is consistent with unresolved objects being moved around a little when the order of floating point calculations changes.  Merging!

sbird:
For posterity, here is the single most different halo from the dm-only run at z=1. It has 97 particles in the master branch, 104 in the treetweaks branch. Green shows particles in the treetweaks branch, blue in the master branch. 
test2
This particular halo wraps around the box boundary at x=250000 so some particles are on the other side of the box. These particles are all in the same positions.

PR #624, opened 2023-12-01
Use the toptree to do exchanges

sbird:
This means we don't need to compute the Peano keys every timestep anymore!

### DIFF
--- a/libgadget/slotsmanager.c
+++ b/libgadget/slotsmanager.c
@@ -374,28 +374,6 @@ slots_gc_slots(int * compact_slots, struct part_manager_type * pman, struct slot
     return 0;
 }
 
-static int
-order_by_type_and_key(const void *a, const void *b)
-{
-    const struct particle_data * pa  = (const struct particle_data *) a;
-    const struct particle_data * pb  = (const struct particle_data *) b;
-
-    if(pa->IsGarbage && !pb->IsGarbage)
-        return +1;
-    if(!pa->IsGarbage && pb->IsGarbage)
-        return -1;
-    if(pa->Type < pb->Type)
-        return -1;
-    if(pa->Type > pb->Type)
-        return +1;
-    if(pa->Key < pb->Key)
-        return -1;
-    if(pa->Key > pb->Key)
-        return +1;
-
-    return 0;
-}
-
 /*Returns the number of non-Garbage particles in an array with garbage sorted to the end.
  * The index returned always points to a garbage particle.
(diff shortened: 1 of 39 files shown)


libgadget/slotsmanager.c
```diff
@@ -424,21 +402,84 @@ slots_get_last_garbage(int nfirst, int nlast, int ptype, const struct part_manag
...
+    int64_t garbage=0;
+    struct PeanoOrder * peanokeys = mymalloc("Keydata", pman->NumPart * sizeof(struct PeanoOrder));
+    #pragma omp parallel for reduction(+: garbage)
+    for(i = 0; i < pman->NumPart; i++) {
+        peanokeys[i].Key = PEANO(pman->Base[i].Pos, pman->BoxSize);
+        peanokeys[i].Type = pman->Base[i].Type;
+        if(pman->Base[i].IsGarbage) {
+            garbage++;
+            /* Set it to a type far beyond the existing types, so it sorts to the end.*/
+            peanokeys[i].Type = 255;
```
rainwoodman, line 449: Since Type in peanokeys differs from Type (a value of 255 for garbage), perhaps rename the field to TypeKey?

libgadget/tests/test_forcetree.c
```diff
@@ -36,9 +36,9 @@ order_by_type_and_key(const void *a, const void *b)
         return -1;
     if(pa->Type > pb->Type)
         return +1;
-    if(pa->Key < pb->Key)
+    if(pa->Pos[0] < pb->Pos[0])
```
rainwoodman, line 39: Is this ordering important? (It was ordered by key and now just by the first pos component, suggesting the order may be only for readability?)

sbird: As I remember the sort was so that the test was more realistic: the default build order was so close to the natural tree ordering that it was unreasonably fast all the time! Maybe I can put this back as it was.

libgadget/slotsmanager.c
sbird: Done!

sbird:
[3d2a05a](https://github.com/MP-Gadget/MP-Gadget/pull/624/commits/3d2a05af873c5dac2d790b1d249a6398a9209ed5) makes a small run ~10% faster: this is not by reducing the communication times, which are negligible, but by improving the balance in the hydro code! I think it might be because the end of the particle table is now all gas, rather than a mixture of gas and DM.

sbird:
Merging, thank you @astro-YYH for the testing.

PR #632, opened 2024-02-27
Parallelcompact

sbird:
A cleanup and make the array compaction parallel.


sbird:
Tested to make no difference to output, does actually speed things up a bit, perhaps through balance, perhaps through better compiler optimisation, perhaps through parallelization.

PR #633, opened 2024-03-06
Fasterpeano

sbird:
Add a test of the Peano curve calculator and do some micro-optimisations with godbolt. Speeds up the timing test from 50 ms to 36 ms on my desktop. Commit 0a353b4c6c7 goes from 37.6 ms to 36, which is small but not nothing.


sbird:
Tested no change to output

PR #634, opened 2024-04-01
Domaincleanup

sbird:
Some cleanups to the domain code


sbird:
This just makes the failure path for domain build, when we could not find a good balance initially, more robust. Should do almost nothing for a regular run.

sbird:
Cleanups: fix the error handling path in domain when the initial balance is poor.

No real changes to the non-failure path.

PR #636, opened 2024-05-15
Fixasynctreefullbuffergravity

sbird:
This should fix the asynchronous treewalk gravity problem when the exchange buffer fills up. As a bonus fix the problem where the potential was not updated every PM step and a bug in the density test.


sbird:
Zero changes to  hydro example up to z=2, small changes therafter probably due to timing differences

PR #637, opened 2024-05-16
potential plane generation

astro-YYH:
added 5 parameters related to potential planes:
PlaneOutputList = "0.2463687286034,0.25331915596808" (example)
PlaneCutPoints = "25000, 75000"
PlaneNormals = "0, 1, 2"
PlaneResolution = 128
PlaneThickness = 50000

output files (planes) would be named snap{}_potentialPlane{}_normal{}.fits (CFITSIO will be required to be installed and -lcfitio should be added when compiling MP-Gadget)

An info file named info.txt summarizing plane comoving distances and redshifts will also be output


gadget/params.c
```diff
@@ -73,6 +74,13 @@ create_gadget_parameter_set()
     param_declare_string(ps, "CpuFile", OPTIONAL, "cpu.txt", "File to output cpu usage information");
     param_declare_string(ps, "OutputList", REQUIRED, NULL, "List of output scale factors.");
 
+    /*Potential plane parameters*/
+    param_declare_string(ps, "PlaneOutputList", OPTIONAL, NULL, "List of potential plane output scale factors.");
+    param_declare_int(ps, "PlaneResolution", OPTIONAL, 0, "Number of pixels per dimension in the potential plane (should be an even number).");
+    param_declare_double(ps, "PlaneThickness", OPTIONAL, 0, "Thickness of the potential plane in the normal direction in kpc/h by default.");
```
sbird, line 80: "kpc/h by default" -> "internal gadget units (kpc/h by default)"

.DS_Store
sbird, line 1: Can you remove this and add to .gitignore?

gadget/params.c
```diff
@@ -373,6 +381,9 @@ create_gadget_parameter_set()
     param_set_action(ps, "BlackHoleFeedbackMethod", BlackHoleFeedbackMethodAction, NULL);
     param_set_action(ps, "StarformationCriterion", StarformationCriterionAction, NULL);
     param_set_action(ps, "OutputList", OutputListAction, NULL);
+    param_set_action(ps, "PlaneOutputList", PlaneOutputListAction, NULL);
+    // param_set_action(ps, "PlaneCutPoints", PlaneCutPointsAction, NULL);
+    // param_set_action(ps, "PlaneNormals", PlaneNormalsAction, NULL);
```
sbird, line 386: Please remove these commented lines

libgadget/plane.h
```diff
@@ -0,0 +1,9 @@
+#ifndef PLANE_H
+#define PLANE_H
+
+#include "cosmology.h"
+
+void set_plane_params(ParameterSet * ps);
+void write_plane(int snapnum, const double atime, const Cosmology * CP, const char * OutputDir, const double UnitVelocity_in_cm_per_s, const double UnitLength_in_cm);
```
sbird, line 7: Please some comments here explaining what the functions do.

libgadget/lenstools.c
```diff
@@ -0,0 +1,412 @@
...
+} GridDimensions;
+
+// Function to determine the bin index for a given value
+int find_bin(double value, double *bins, int num_bins) {
+    for (int i = 0; i < num_bins - 1; i++) {
+        if (value >= bins[i] && value < bins[i + 1]) {
+            return i;
+        }
+    }
+    return -1;  // Return the last bin if the value is on the boundary or beyond
```
sbird, line 98: In C the correct return value is num_bins-1

libgadget/lenstools.c
```diff
@@ -0,0 +1,412 @@
...
+}
```
sbird, line 99: Binary search is more efficient (slightly edited chatgpt):
// Function to determine the bin index for a given value
int find_bin(double value, double *bins, int num_bins) {
    int low = 0, high = num_bins - 1;
    if (value > bins[num_bins-1])
         return num_bins - 1;
if (value < bins[0])
return 0;
    while (low < high) { //chatgpt actually had while(low <= high) but I think that is not right!
        int mid = (low + high) / 2;
        if (value >= bins[mid] && value < bins[mid + 1]) {
            return mid;
        } else if (value < bins[mid]) {
            high = mid - 1;
        } else {
            low = mid + 1;
        }
    }
   
    return low; 
}

libgadget/lenstools.c
```diff
@@ -0,0 +1,412 @@
...
+void linspace(double start, double stop, int num, double *result) {
+    double step = (stop - start) / (num - 1);
+    for (int i = 0; i < num; i++) {
+        result[i] = start + i * step;
+    }
+}
+
+double ***allocate_3d_array(int nx, int ny, int nz) {
+    // Allocate memory for the 3D array
+    double ***array = malloc(nx * sizeof(double **));
```
sbird, line 26: In MP-Gadget we use mymalloc/myfree which has a code-managed heap to avoid memory fragmentation. The restriction is that memory has to be de-allocated in the reverse order that it is allocated. So you write something like:

array = mymalloc("3ds", nx*sizeof(double **));
array[i] = mymalloc("2ds", ny * sizeof(double *));
array[i][j] = ...
(stuff)
myfree(array[i][j]);
myfree(array[i]);
myfree(array);

libgadget/lenstools.c
```diff
@@ -0,0 +1,412 @@
...
+    if (array == NULL) {
+        fprintf(stderr, "Memory allocation failed\n");
+        exit(EXIT_FAILURE);
+    }
```
sbird, line 30: If you are using mymalloc the code auto-exits when malloc fails and you don't need the checks.

libgadget/lenstools.c
```diff
@@ -0,0 +1,412 @@
...
+        for (int j = 0; j < ny; j++) {
+            array[i][j] = malloc(nz * sizeof(double));
+            if (array[i][j] == NULL) {
+                fprintf(stderr, "Memory allocation failed for layer %d, row %d\n", i, j);
+                exit(EXIT_FAILURE);
+            }
+            // Initialize elements (optional)
+            for (int k = 0; k < nz; k++) {
+                array[i][j][k] = 0.0; // Example initialization
+            }
```
sbird, line 47: Better to use memset()

libgadget/lenstools.c
```diff
@@ -0,0 +1,412 @@
...
+int find_bin(double value, double *bins, int num_bins) {
+    for (int i = 0; i < num_bins - 1; i++) {
+        if (value >= bins[i] && value < bins[i + 1]) {
+            return i;
+        }
+    }
+    return -1;  // Return the last bin if the value is on the boundary or beyond
+}
+
+void grid3d_nfw(const struct particle_data * Parts, int num_particles, double **binning, GridDimensions dims, double ***density) {
```
sbird, line 101: Curious why grid3d_nfw? Maybe grid3d_cic would be better?

libgadget/lenstools.c
sbird: Ah but I see you check for -1 below! This is really an error value.

libgadget/lenstools.c
```diff
@@ -0,0 +1,412 @@
...
+                    density_projected[i][j] += density[i][k][j];
+                } else {
+                    density_projected[i][j] += density[i][j][k];
+                }
+            }
+        }
+    }
+}
+
+void calculate_lensing_potential(double **density_projected, int plane_resolution, double bin_resolution_0, double bin_resolution_1, double chi,double smooth, double **lensing_potential) {
```
sbird, line 151: I think this might be a function we already have in the PM code. Let me look...

libgadget/lenstools.c
```diff
@@ -0,0 +1,412 @@
...
+        }
+    }
+    return -1;  // Return the last bin if the value is on the boundary or beyond
+}
+
+void grid3d_nfw(const struct particle_data * Parts, int num_particles, double **binning, GridDimensions dims, double ***density) {
+    
+    double position[3];
+    // Process each particle
+    for (int p = 0; p < num_particles; p++) {
```
sbird, line 105: We want this to be #pragma omp parallel for but then density[] will be written to from different threads and will race. In petapm the similar code guards density with #pragma omp atomic, but it might be more effective to allocate separate arrays for each  processor if density is small.

libgadget/lenstools.c
```diff
@@ -0,0 +1,412 @@
...
+        }
+        int ix = find_bin(position[0], binning[0], dims.nx + 1);
+        int iy = find_bin(position[1], binning[1], dims.ny + 1);
+        int iz = find_bin(position[2], binning[2], dims.nz + 1);
+
+        // continue if the particle is outside the grid
+        if (ix == -1 || iy == -1 || iz == -1) {
+            continue;
+        }
+        // Increment the density in the appropriate bin
```
sbird, line 121: eg: #pragma omp atomic update

libgadget/lenstools.c
```diff
@@ -0,0 +1,412 @@
...
+    dims.nx = (normal == 0) ? thickness_resolution : plane_resolution;
+    dims.ny = (normal == 1) ? thickness_resolution : plane_resolution;
+    dims.nz = (normal == 2) ? thickness_resolution : plane_resolution;
+    // printf("nx: %d, ny: %d, nz: %d\n", dims.nx, dims.ny, dims.nz);
+
+    double ***density = allocate_3d_array(dims.nx, dims.ny, dims.nz);
+
+    grid3d_nfw(P, num_particles_rank, binning, dims, density);
+
+    projectDensity(density, density_projected, dims, normal);
```
sbird, line 300: Am I reading this correctly that it is only density_projected that is used below? Can we skip computing density completely and save the memory?

libgadget/lenstools.c
sbird: If not we should rewrite it to be mpi-parallel along the lines of petapm. I can have a go at that.

libgadget/plane.c
```diff
@@ -0,0 +1,279 @@
...
+            // print input parameters
+
+            
+
+            /*computing lensing potential planes*/
+            num_particles_plane = cutPlaneGaussianGrid(num_particles_tot,  comoving_distance, BoxSize, CP, atime, PlaneParams.Normals[j], PlaneParams.CutPoints[i], thickness, left_corner, plane_resolution, plane_result);
+            
+            /*sum up planes from all tasks*/
+            for (int k = 0; k < plane_resolution; k++) {
+            MPI_Reduce(plane_result[k], (ThisTask == 0 ? summed_plane_result[k] : NULL), plane_resolution, MPI_DOUBLE, MPI_SUM, 0, MPI_COMM_WORLD);
```
sbird, line 253: Should be :
MPI_Reduce(plane_result, summed_plane_result, plane_resolution, MPI_DOUBLE, MPI_SUM, 0, MPI_COMM_WORLD);
and no need for the loop, it reduces the whole array.

sbird: Oh, I see, you are trying to reduce a 2D array! Maybe easier to use a flat array of memory so you can have a single reduce. Multiple reduces in a loop sometimes cause MPI to break.

sbird:
Looking good...can you address the comments and then we can look at the MPI parallelisation. Remind me many numbers a single plane usually is? Was it 128^2?

libgadget/lenstools.c
astro-YYH: Sorry, actually that comment is not correct, and I have changed it. (nothing wrong with the code, yes, that is an 'error' value)

libgadget/lenstools.c
astro-YYH: Yes, binary search is better, and I have switched the function to this method (with some modifications: eg, 'high = mid -1' --> 'high = mid')!

libgadget/lenstools.c
astro-YYH: I rewrite the function as:

double ***allocate_3d_array_new(int nx, int ny, int nz) {
    // Allocate memory for the 3D array
    double ***array = mymalloc("3ds", nx * sizeof(double **));
    if (array == NULL) {
        fprintf(stderr, "Memory allocation failed\n");
        exit(EXIT_FAILURE);
    }

    for (int i = 0; i < nx; i++) {
        array[i] = mymalloc("2ds", ny * sizeof(double *));
        if (array[i] == NULL) {
            fprintf(stderr, "Memory allocation failed for layer %d\n", i);
            exit(EXIT_FAILURE);
        }
        for (int j = 0; j < ny; j++) {
            array[i][j] = mymalloc("1ds",nz * sizeof(double));
            if (array[i][j] == NULL) {
                fprintf(stderr, "Memory allocation failed for layer %d, row %d\n", i, j);
                exit(EXIT_FAILURE);
            }
            // Initialize elements (optional)
            for (int k = 0; k < nz; k++) {
                array[i][j][k] = 0.0; // Example initialization
            }
        }
    }

    return array;
}

But got error messages like:

1ds                  | B | 000000000000 000000000004 | lenstools.c:70
[ 000003.86 ] Task 1:  1ds                  | B | 000000000000 000000000004 | lenstools.c:70
[ 000003.86 ] Task 0:  1ds                  | B | 000000000000 000000000004 | lenstools.c:70
[ 000003.86 ] Task 1:  1ds                  | B | 000000000000 000000000004 | lenstools.c:70
[ 000003.86 ] Task 1:  1ds                  | B | 000000000000 000000000004 | lenstools.c:70
[ 000003.86 ] Task 0:  1ds                  | B | 000000000000 000000000004 | lenstools.c:70
[ 000003.86 ] Task 0:  1ds                  | B | 000000000000 000000000004 | lenstools.c:70
[ 000003.86 ] Task 0: [ 000003.86 ] Task 1:  1ds                  | B | 000000000000 000000000004 | lenstools.c:70
[ 000003.86 ] Task 1:  1ds                  | B | 000000000000 000000000004 | lenstools.c:70
[ 000003.86 ] Task 0:  1ds                  | B | 000000000000 000000000004 | lenstools.c:70
 1ds                  | B | 000000000000 000000000004 | lenstools.c:70
[ 000003.86 ] Task 1:  TopLeaves            | T | 000000000000 000000000004 | domain.c:198
[ 000003.86 ] Task 0:  1ds                  | B | 000000000000 000000000004 | lenstools.c:70
[ 000003.86 ] Task 1:  TopNodes             | T | 000000000000 000000000004 | domain.c:196
[ 000003.86 ] Task 1: [ 000003.86 ] Task 0:  1ds                  | B | 000000000000 000000000004 | lenstools.c:70
[ 000003.86 ] Task 0:  Tasks                | T | 000000000000 000000000004 | domain.c:374
[ 000003.86 ] Task 1:  Mesh2Task            | T | 000000000001 000000000004 | petapm.c:119
 1ds                  | B | 000000000000 000000000004 | lenstools.c:70
[ 000003.86 ] Task 0:  1ds                  | B | 000000000000 000000000004 | lenstools.c:70
[ 000003.86 ] Task 0: [ 000003.86 ] ERROR: Task 1: Not enough memory for 1ds 4096 bytes
 1ds                  | B | 000000000000 000000000004 | lenstools.c:70
[ 000003.86 ] Task 0:  1ds                  | B | 000000000000 000000000004 | lenstools.c:70

The way I call mymalloc might not be right?

libgadget/lenstools.c
astro-YYH: I just followed the function name in lenstools, but actually this function here should only be part of the original one (all we need in our case) which doesn't look like related to NFW. So yes, we may rename it. What does cic mean?

libgadget/lenstools.c
astro-YYH: Yes, currently only density_projected is used, but I recommend reserving the 3D density since it could be useful in future (3D density planes for visualization for example), and the density should not occupy much memory since we now have fixed thickness_resolution to 1 (in future if needed, we can add ThicknessResolution and PlaneKind to generate 3D density planes):

`int thickness_resolution = 1;  // Number of bins along the thickness direction, fixed to 1 for now`

density and density_projected should be equal in size and they are not very large, typically e.g., 2048^2 double (64 MB), what do you think?

libgadget/lenstools.c
sbird: CIC is short for cloud-in-cell, which is a way to assign density to a regular grid. Actually this should be NGB (nearest grid point). The three methods are NGB, CIC and TSC (triangular shaped cloud).

libgadget/lenstools.c
```diff
@@ -0,0 +1,412 @@
...
+    int nx, ny, nz;
+} GridDimensions;
+
+// Function to determine the bin index for a given value
+int find_bin(double value, double *bins, int num_bins) {
+    for (int i = 0; i < num_bins - 1; i++) {
+        if (value >= bins[i] && value < bins[i + 1]) {
+            return i;
+        }
+    }
```
sbird, line 97: Actually now I think about it, for these regularly spaced arrays you don't need a for loop anyway, you can just calculate directly the bin from the position. (pos[0] / box * nx)

libgadget/lenstools.c
sbird: We can already compute the 3D density in petapm.c very efficiently. That code is quite involved and complicated because of the MPI parallelisation: each processor has a different part of physical space. The code is set up so that it only allocates memory for that specific part of physical space. For these arrays if they are really that small it makes sense to allocate them on each processor and reduce them to make the code simpler.

If I understand right though, density_projected doesn't do anything when thickness_resolution is 1? So we don't need two arrays, I think. Even if you wanted to keep the 3D array it would be more efficient to reduce it to 2D in place.

libgadget/lenstools.c
sbird: Maybe try doing a single allocation and setting each element to an offset within that allocation?
            array[0][0] = mymalloc("1ds",nx * ny * nz * sizeof(double));
for(i, j)
array[i][j] = array + i * nx *ny + ny * j;
(or something like that: I didn't work out the actual offsets here).

astro-YYH: I have rewritten those allocation functions and use mymalloc and myfree! (and access 2D and 3D arrays in a 1D style)

sbird: Nice! I will take a look.

libgadget/lenstools.c
sbird: Ignore this: the arrays are small enough that we should not worry about it.

libgadget/lenstools.c
astro-YYH: Yes, changed!

libgadget/lenstools.c
```diff
@@ -0,0 +1,336 @@
...
+}
+
+typedef struct {
+    int nx, ny, nz;
+} GridDimensions;
+
+// Function to determine the bin index for a given value
+int find_bin(double value, double *bins, int resolution, const double L) { // L is the box size
+    // float index
+    float iflt = (value - bins[0]) / (bins[resolution] - bins[0]) * resolution;
```
sbird, line 49: double iflt and (int) floor(iflt);

libgadget/lenstools.c
```diff
@@ -0,0 +1,336 @@
...
+    int index = (int)floorf(iflt);
+
+    // check if the value is within the range
+    if (index >= 0 && index < resolution) {
+        return index;
+    }
+    else {
+        return -1;
+    }
+    endrun(0, "Error in find_bin\n");
```
sbird, line 61: Don't need the endrun because it either returns -1 or index so endrun will never be reached.

libgadget/lenstools.c
```diff
@@ -0,0 +1,336 @@
...
+        if (ix == -1 || iy == -1 || iz == -1) {
+            continue;
+        }
+        // Increment the density in the appropriate bin
+        ACCESS_3D(density, ix, iy, iz, dims.ny, dims.nz)++;
+    }
+
+}
+
+void projectDensity(double *density, double *density_projected, GridDimensions dims, int normal) {
```
sbird, line 90: I think the best way to keep the code general but still fast is to keep grid3d_nfw as is, but make this function do a reduce in place along the z directrion. Thus it would be a no-op for thickness = 1. 

So something like:
for (k = 1 ; k < DimNorm; k++) //Note that this does nothing if there is only one z-bin!
for(i)
for (j)
 ACCESS_3D(density_projected, i, j, 0, Dim1) += ACCESS_3D(density, i, k, j, dims.ny, dims.nz);

If you arrange the memory right you can use realloc() afterwards to shrink the memory (although in this case there is no need for that)

libgadget/lenstools.c
```diff
@@ -0,0 +1,336 @@
...
+    // Initialize density_ft to zero
+    for (int i = 0; i < plane_resolution * (plane_resolution / 2 + 1); i++) {
+        density_ft[i][0] = 0.0;  // Real part
+        density_ft[i][1] = 0.0;  // Imaginary part
+    }
+
+    double *l_squared = allocate_2d_array_as_1d(plane_resolution, plane_resolution / 2 + 1);
+
+    // Create FFTW plans
+    fftw_plan forward_plan = fftw_plan_dft_r2c_2d(plane_resolution, plane_resolution, density_projected, density_ft, FFTW_ESTIMATE);
```
sbird, line 126: I think that this is openmp parallel automatically because we call pfft_plan_with_nthreads before we do this

libgadget/lenstools.c
```diff
@@ -0,0 +1,336 @@
...
+}
+
+int64_t cutPlaneGaussianGrid(int num_particles_tot, double comoving_distance, double Lbox, const Cosmology * CP, const double atime, const int normal, const double center, const double thickness, const double *left_corner, const int plane_resolution, double *lensing_potential) {
+    // Get the rank of the current process
+    int ThisTask;
+    MPI_Comm_rank(MPI_COMM_WORLD, &ThisTask);
+    // smooth
+    float smooth = 1.0; // fixed in our case
+
+    int num_particles_rank = PartManager->NumPart;  // dark matter-only simulation: NumPart = number of dark matter particles
```
sbird, line 182: this is int64_t

libgadget/lenstools.c
```diff
@@ -0,0 +1,336 @@
...
+    // }
+    myfree(l_squared);
+}
+
+int64_t cutPlaneGaussianGrid(int num_particles_tot, double comoving_distance, double Lbox, const Cosmology * CP, const double atime, const int normal, const double center, const double thickness, const double *left_corner, const int plane_resolution, double *lensing_potential) {
+    // Get the rank of the current process
+    int ThisTask;
+    MPI_Comm_rank(MPI_COMM_WORLD, &ThisTask);
+    // smooth
+    float smooth = 1.0; // fixed in our case
```
sbird, line 180: make it a double

libgadget/lenstools.c
astro-YYH: Got it!

libgadget/lenstools.c
astro-YYH: Right, let me remove it.

libgadget/lenstools.c
astro-YYH: Yes

libgadget/run.c
```diff
@@ -710,6 +716,14 @@ run(const int RestartSnapNum, const inttime_t ti_init, const struct header_data
             }
         }
 
+        /* Write the potential planes*/
+        if(WritePlane)
+#ifdef USE_CFITSIO
+            write_plane(PlaneSnapNum, atime, &All.CP, All.OutputDir, units.UnitVelocity_in_cm_per_s, units.UnitLength_in_cm);
```
sbird, line 722: walltime_measure("/Lensing");

libgadget/lenstools.c
astro-YYH: If I remember, we do not want to make it omp parallel anymore?

sbird: No want to make it omp parallel but not MPI parallel

libgadget/lenstools.c
astro-YYH: yes, done!

libgadget/run.c
astro-YYH: Done

libgadget/lenstools.c
astro-YYH: Yes, I think it runs in omp parallel automatically already. I have compared with explicitly using omp here, the times spent are similar.

astro-YYH:
Tests done:
1. potential planes generated by MP-Gadget are exactly the same as those from lenstools;
2. potential planes generated by MP-Gadget can be used by lenstools to produce weak lensing maps.

sbird:
It is good. Are you ready to merge or you are still checking?

astro-YYH:
> It is good. Are you ready to merge or you are still checking?

Yes, I am ready, can you merge them? (or let me do that?)

sbird:
Let me do the merge so I can check it still works ok on my system

sbird:
Can you also set the default values for PlaneResolution and PlaneThickness to something sensible? Whatever you have been using is fine.

astro-YYH:
> Can you also set the default values for PlaneResolution and PlaneThickness to something sensible? Whatever you have been using is fine.

done!

sbird:
Ok! Looks good. Let's merge this. Can you merge my PR into this if it looks good and then I will merge the whole thing to master?

astro-YYH:
> Ok! Looks good. Let's merge this. Can you merge my PR into this if it looks good and then I will merge the whole thing to master?

Done!

PR #650, opened 2024-08-17
Revert "Fix a rare edge case where kicks could be mis-applied."

sbird:
This reverts commit 5dfc97198cbc538bfbd03a4bf082e8ebe2fa8e93.

Trying to work out why the dm-small fof test started failing.


sbird:
Fix the test by updating the fof values

PR #656, opened 2024-09-27
fix the memory issue of export/import 

YihaoZhou27:
1. change the ImportBufferBoost to be a float parameter 
2. make DomainOverDecompositionFactor half of the available threads, so there are fewer export particles


sbird:
Thank you @YihaoZhou27

PR #661, opened 2024-10-17
Save a snapshot for the code used for Astrid production run.

YihaoZhou27:
This MP-Gadget is being used to run Astrid. It is in general efficient and bug-free.


sbird:
Something odd happened with the PR here: let me make a branch that tracks your commit.

sbird:
Also tracked in MP-Gadget/main-run

PR #668, opened 2024-11-28
Make some petapm variables 64-bit

sbird:
In case they overflow (or the compiler does something odd).


sbird:
Not clear to me why this doesn't work!

sbird:
It was not working before because the address space partition for the nodes has to be the same on all ranks.

PR #670, opened 2024-12-20
Make shared omp variables const and move to firstprivate

sbird:
In the omp task in forcetree_recursive, we had some shared variables. However, these variables could all be made const. Once this happens they can be moved to firstprivate because the compiler can optimise them away and avoid using new stack space. The shared clause can go away.

This also means that we no longer fail to compile on old gcc that has default shared openmp variables!

Thanks to Asher Sang for the suggestion.


sbird:
Looks slightly faster.

PR #674, opened 2025-05-02
Fix for bug in plot-structure.py extent handling

dangallego:
This PR corrects a data type issue in plot-structure.py where the extent argument passed to plt.imshow() could include NumPy arrays instead of floats, causing rendering errors. Now the box size is explicitly cast to a float when needed. Comments and cleanup have also been added for clarity.


sbird:
Thank you! Looks good, will merge.

PR #675, opened 2025-05-07
Fixbalance

sbird:
Fix assignment of particles to topleaves, which was being done before the domain was finalised for full domain decompositions. This was usually fine because most of the time the final balance didn't change the topleaf order! There are some edge cases where the balance should be much better, however.


sbird:
This produces bit-identical output on the hydro run. The error cases fixed indeed do not appear to occur with a standard run.

PR #677, opened 2025-06-03
Fixsfrtxt

sbird:
Fixes to printing sfr.txt


sbird:
Yes, I agree! Thanks for the review. It does not explain what we are seeing.

PR #686, opened 2026-01-16
Fix the issue of gas SFR in the last snapshot.

YihaoZhou27:
The gas StarFormationRate has a lot of NaN in the last snapshot (the one with TimeMax), which was caused by zero dtime at the last sync point.


sbird:
Thanks!

PR #690, opened 2026-08-30
Fix the issue that AutoCheckPointTime is never triggered

YihaoZhou27:
hci.c initialized TimeLastCheckPoint with timer_begin, which is an absolute MPI_Wtime() value (>1e9), but hci_query_auto_checkpoint compare it againthci_get_elapsed_time(), which returns the time since the run started. As a result, AutoCheckPointTime is never triggered.


sbird:
Thanks Yihao - this also explains why it sometimes worked: I think some MPIs start MPI_Wtime at zero, or at last boot.

PR #691, opened 2026-09-12
Fix the issue of missing h in time unit conversion

YihaoZhou27:
Correct two time-unit conversions involving the Hubble parameter `h`. In the metal return module, we need to convert the internal time unit to the physical time scale for the star age. A hubble parameter was missed in this conversion.

### DIFF
--- a/libgadget/metal_return.c
+++ b/libgadget/metal_return.c
@@ -168,7 +168,7 @@ static double atime_to_myr(Cosmology *CP, double atime1, double atime2, gsl_inte
     gsl_function ff = {atime_integ, CP};
     double tmyr, abserr;
     gsl_integration_qag(&ff, atime1, atime2, 1e-4, 0, GSL_WORKSPACE, GSL_INTEG_GAUSS61, gsl_work, &tmyr, &abserr);
-    return tmyr * CP->UnitTime_in_s / SEC_PER_MEGAYEAR;
+    return tmyr * CP->UnitTime_in_s / (CP->HubbleParam * SEC_PER_MEGAYEAR);
 }
 
 /* Functions for the root finder*/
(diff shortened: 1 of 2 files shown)


libgadget/sfr_eff.c
```diff
@@ -994,7 +994,7 @@ void init_cooling_and_star_formation(int CoolingOn, int StarformationOn, Cosmolo
     }
 
     if(sfr_params.WindOn) {
-        init_winds(sfr_params.FactorSN, sfr_params.EgySpecSN, sfr_params.PhysDensThresh, units.UnitTime_in_s);
```
sbird, line 997: I think this is actually wrong: it affects only MaxWindFreeTravelTime which is a free parameter and has always been specified in Myr/h. The right fix is to change the string in the parameter to say "Maximum time in Myrs/h for the wind to be decoupled" rather than "Maximum time in Myrs for the wind to be decoupled".

libgadget/metal_return.c
```diff
@@ -168,7 +168,7 @@ static double atime_to_myr(Cosmology *CP, double atime1, double atime2, gsl_inte
     gsl_function ff = {atime_integ, CP};
     double tmyr, abserr;
     gsl_integration_qag(&ff, atime1, atime2, 1e-4, 0, GSL_WORKSPACE, GSL_INTEG_GAUSS61, gsl_work, &tmyr, &abserr);
-    return tmyr * CP->UnitTime_in_s / SEC_PER_MEGAYEAR;
```
sbird, line 171: This is correct, but we need to think carefully. The output of the function is stored in LastEnrichmentMyr, and saved in the snapshots. As it stands when this is read from an existing snapshot we are going to get LastEnrichmentMyr  < StellarAge for some stars. Perhaps we should save LastEnrichmentMyr in Myr/h and do the conversion on reload in the petaio function?

YihaoZhou27:
Thanks for the clarification! I have updated the code to revert the changes in `sfr_eff.c` and correct the parameter explanation for `MaxWindFreeTravelTime`. I also added the unit conversion in the IO so now the `LastEnrichmentMyr` unit in the snapshots is preserved.

sbird:
Thank Yihao

